PR #65, opened 2017-10-29
add support for gcp auth in brig command

spheromak:
The `client-go` package doesn't directly import the GCP auth backend. This PR adds an import to the brig command to use all the client-go supported providers.

Without these changes you would get an message saying  `Error: No Auth Provider found for name "gcp"`

### DIFF
--- a/brigade-client/cmd/brig/commands/app.go
+++ b/brigade-client/cmd/brig/commands/app.go
@@ -4,6 +4,10 @@ import (
 	"os"
 
 	"github.com/spf13/cobra"
+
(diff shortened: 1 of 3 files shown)


adamreese:
Can you add the package that includes all auth providers?

`import _ k8s.io/client-go/plugin/pkg/client/auth`

spheromak:
@adamreese  that sounds good to me I will rework and update.

spheromak:
@adamreese  Updated to use all auth providers, and also signed the CLA

adamreese:
Thanks!

PR #67, opened 2017-10-30
fix(worker): disable shared storage by default

technosophos:
Leaving this enabled adds a pretty serious performance hit on at least
some systems. And it is not typically used by containers out of the box.
That is, most of the time it is used, usage is already manually
configured by the script author. So it makes sense to go with
performance as the default metric.


bacongobbler:
Is there documentation we need to update, or are all the examples good to go? I'm pretty sure there must be at least one example that demonstrates shared storage so we should go through and add this in.

technosophos:
@bacongobbler good catch. I just found and fixed the related docs, too.

PR #71, opened 2017-10-30
feat(controller,worker): acid.js not required

technosophos:
This makes the acid.js an optional parameter of the event secret.

If it is not specified, the worker will attempt to get the file out of
an attached VCS repo.

(Migrated from old repo)

### DIFF
--- a/brigade-client/cmd/brig/commands/run.go
+++ b/brigade-client/cmd/brig/commands/run.go
@@ -32,12 +32,17 @@ const runUsage = `Send a Brigade JS file to the server.
 This sends a file into the cluster and waits for it to complete. It accepts
 a project name or project ID.
 
-	$ brigade run deis/empty-testbed
+	$ brig run deis/empty-testbed
 
 When no JS file is supplied, the project will be checked for a brigade.js file
 in the associated repository.
 
-Be careful when setting an event, as many events expect a particular payload.
+To send a local JS file to the server, use the '-f' flag:
+
+	$ brig run -f my.js deis/empty-testbed
+
+While specifying an event is possible, use caution. Mny events expect a
+particular payload.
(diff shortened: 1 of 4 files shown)


brigade-client/cmd/brig/commands/run.go
```diff
@@ -44,6 +45,8 @@ func init() {
 	run.Flags().StringVarP(&runFile, "file", "f", "./brigade.js", "The JavaScript file to execute")
 	run.Flags().StringVarP(&runEvent, "event", "e", "exec", "The name of the event to fire")
 	run.Flags().StringVarP(&runPayload, "payload", "p", "", "The path to a payload file")
+
+	run.Flags().BoolVar(&runNoScript, "no-script", false, "Do not send a script at all")
```
adamreese: rather than a `no-script` flag, can we just send no script?

brigade-controller/cmd/brigade-controller/controller/controller_test.go
```diff
@@ -75,4 +93,173 @@ func TestController(t *testing.T) {
...
+		t.Errorf("expected sidecar %q, got %q", sidecarImage, ic.Image)
+	}
+
+	if ic.VolumeMounts[0].Name != sidecarVolumeName {
+		t.Errorf("expected sidecar volume %q, got %q", sidecarVolumeName, ic.VolumeMounts[0].Name)
+	}
+}
+
+func TestController_WithScript(t *testing.T) {
+	// t.Skip("make better")
```
adamreese: is this made better already?

technosophos: Isn't it better just by commenting it out?

PR #86, opened 2017-11-01
feat(git-sidecar): improve the SSH handling

technosophos:
This increases the security of the SSH configuration so that host key
verification actually works.

Re-opening this post-release.


git-sidecar/rootfs/clone.sh
```diff
@@ -1,5 +1,21 @@
...
+
+# From the SSH key, generate the known hosts file
+if [ "" != "${BRIGADE_REPO_KEY}" ]; then
+  mkdir -p $HOME/.ssh
+  KEY="$HOME/.ssh/id_rsa"
+  echo ${BRIGADE_REPO_KEY} | tr '$' '\n' > $KEY
+  chmod 600 $KEY
+  ssh-keygen -y -f $HOME/.ssh/id_rsa > $HOME/.ssh/id_rsa.pub
+  echo "github.com ssh-rsa $(cat $HOME/.ssh/id_rsa.pub)" >> $HOME/.ssh/known_hosts
+  echo "bitbucket.com ssh-rsa $(cat $HOME/.ssh/id_rsa.pub)" >> $HOME/.ssh/known_hosts
```
j2sol: Maybe I'm confused, but it appears you're placing the ssh user's public key into the known_hosts file as if it were the target hosts's public hostkey. I don't think that will work, as the keys will not match. The hostkey of GitHub.com can be discovered thusly:

```
> ssh-keyscan github.com
# github.com:22 SSH-2.0-libssh_0.7.0
github.com ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ==
```

But you probably don't want to do this programmatically at run time, if the fear is a compromised DNS leading to a nefarious host. If that is the concern, then the host keys for these known hosts should go into the rootfs as `/etc/ssh/ssh_known_hosts` (although it could exist in the user's dir, it just feels cleaner to toss it in /etc/.

mumoshu: Agreed.

Probably you would better either:

- hard-code [GH's ssh key fingerprints](https://help.github.com/articles/github-s-ssh-key-fingerprints/) so that any compromised DNS is unable to trap you
   - not sure if bitbucket has the well-known set of fingerprints like GH though?
- and/or allow one to add more fingerprints via project's values.yaml

FYI, I've seen [this similar issue in an another CI system.](https://github.com/concourse/git-resource/issues/115)

technosophos:
I am not sure whether this PR is worth resurrecting at this point. Any thoughts?

radu-matei:
Ping @technosophos - would this still be mergeable in the current state?

krancour:
Closing this since it's quite stale. @technosophos please do re-open if this is anything you want to revive.

PR #87, opened 2017-11-01
chore(charts): update charts to 0.3.0

technosophos:


chart/brigade/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
-description: A Helm chart for Kubernetes
+description: Brigade provides event-drive scripting of Kubernetes pipelines.
```
bacongobbler: event-driven*

bacongobbler: should probably re-call `make build-chart` after fixing this typo

PR #92, opened 2017-11-02
change node:8 -> node:8-alpine

u2takey:
Why this pr:
This is proposed for better developing and ci/cd experience. If there is no special concern on node image version.


technosophos:
I believe @adamreese is testing this one. If not, I will test.

PR #94, opened 2017-11-02
fix(tests): bump empty-testbed sha

adamreese:


bacongobbler:
nice catch on the rename.

PR #99, opened 2017-11-03
feat(cr-gateway): break container registry gw into separate binary

technosophos:
This takes the CR gateway out of the github gateway.

It's WIP because we need to fix the `:latest` thing "the right way" before this will work.


technosophos:
Depends on #103

technosophos:
Now that #103 is merged, I think this one is ready.

brigade-cr-gateway/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM alpine:3.6
+
+RUN apk update && apk add --no-cache \
+    ca-certificates \
+    git \
+    && update-ca-certificates
+
+COPY rootfs /
+
+ENV GIT_SSH=/gitssh.sh
```
adamreese: gitssh.sh is missing

brigade-cr-gateway/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+CMD /usr/bin/brigade-cr-gateway
```
adamreese: binary was copied to `/` https://github.com/Azure/brigade/blob/master/brigade-gateway/Dockerfile

tests/tests_test.go
```diff
@@ -14,11 +14,6 @@ func TestFunctional(t *testing.T) {
 		t.Fatal(err)
 	}
 	defer githubPushFile.Close()
-	dockerhubFile, err := os.Open("testdata/dockerhub-push.json")
```
adamreese: are these moving somewhere?

brigade-cr-gateway/Dockerfile
technosophos: Weird... I see it in /usr/bin/brigade-cr-gateway

technosophos: Ah... rebasing mistake maybe?

tests/tests_test.go
technosophos: Moved (well... refactored into unit tests)

technosophos:
This is ready for a re-review.

Since we don't have a pattern for multi-binary functional tests yet, I rolled some of the test back into a unit test, and am relying on other unit tests for the rest of the API test on DockerHub. I am also switching the event listener to do `image_push`, which is a breaking change.

bacongobbler:
should brigade-gateway be renamed to brigade-github-gateway with this change? It looks like if this is the move going forward for the project, we should probably rename it to be more clear that "gateways" are merely an API that registers with the controller to handle incoming events.

Also what does this mean for the docs? If the user wants to set up both gateways, they will have to set up two separate load balancers because they're under two separate services in the chart. Is this the right approach for our users, or would this be deemed "too complex"?

technosophos:
Good questions.

1. I would defer the question of renaming the other gateway to @adamreese 

2. I think we actually want different inbound ingress/LB for each gateway, for security reasons. And we need to also make it possible to add SSL-LEGO in front of these. That seems to indicate that we might be heading toward a more sophisticated set of installation requirements. I don't know how I feel about that.

technosophos:
Updated documentation with configuration information

PR #103, opened 2017-11-05
ref(chart): make chart smart about versions

technosophos:
This is part of a bigger effort to make version handling in Brigade
better.

This changes the charts to use the `appVersion` field in `Chart.yaml` to
set the image tag. The `tag` field still exists for overrides, but it
simplifies the chart to assume that each component of the chart should
use the master version number to pull its tag.

Along the way, I improved documentation, and made it trivially possible
to disable API and Gateway.


technosophos:
This directly relates to @adamreese 's work on versioning

PR #106, opened 2017-11-06
Add missing highlights to object names

j2sol:
Event and Job objects did not have the same highlighting as other headings, which caused me some visual confusion.


j2sol:
Actually the job object section seems to be a repeat of some info from the earlier Job class section. Am I missing something or are they describing the same thing?

PR #110, opened 2017-11-06
Update location of brigade-project helm chart

j2sol:
The directions weren't clear that the chart was inside the chart/ subdirectories, like the brigade chart itself.


j2sol:
Hah! Weird timing with you moving it and me getting to that point in the set up.

PR #115, opened 2017-11-09
fix(makefile): add CGO_ENABLED=0  for build-docker-bin

u2takey:
fix https://github.com/Azure/brigade/issues/113


mazzus:
I think "go" should be after "CGO_ENABLED=0" on line 35. Otherwise i think this fix looks good :)

technosophos:
Thanks!

PR #117, opened 2017-11-09
fix(worker): mount docker socket when privileged=true

technosophos:
This fixes the privileged mode to mount the Docker socket.

### DIFF
--- a/brigade-worker/src/k8s.ts
+++ b/brigade-worker/src/k8s.ts
@@ -245,9 +245,22 @@ export class JobRunner implements jobs.JobRunner {
       this.secret.data["main.sh"] = b64enc(newCmd)
     }
 
+    // If job is privileged, add docker socket and mark pod privileged.
     if (job.privileged) {
+      const sockPath = "/var/run/docker.sock"
+      const sockName = "docker-socket"
+
+      var dockerVol = new kubernetes.V1Volume()
+      dockerVol.name = sockName
+      dockerVol.hostPath = { path: sockPath }
+      this.runner.spec.volumes.push(dockerVol)
+
       for (let i = 0; i < this.runner.spec.containers.length; i++) {
+        var dockerMnt = new kubernetes.V1VolumeMount()
+        dockerMnt.mountPath = sockPath
+        dockerMnt.name = sockName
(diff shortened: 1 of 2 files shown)


technosophos:
To be clear, this should allow you to use the `docker` image from inside of Kubernetes. This is useful for running in-cluster `docker build` operations and the like.

u2takey:
It maybe better to add a new flag for mount the Docker socket? (or move all volumn aciton out for user? ). 

User use **docker in docker ** may not need docker socket in host (and it is not safe). For example people can use https://github.com/jpetazzo/dind for docker build and save docker build cache in brigadeCache for next builds.

bacongobbler:
You’ll want to expose the docker engine flags. Some clusters need to use specific storage backend as like overlayfs as the backing file system driver. The default fs backend for docker fails to work on Azure/minikube for example.

bacongobbler:
You can see the current list of issues with using dind on kubernetes for Draft here: https://github.com/Azure/draft/issues?q=is%3Aissue+overlay

technosophos:
I have been testing this out:

```javascript
const { events, Job } = require("brigadier")

events.on("exec", () => {
  const dind = new Job("dind", "docker:edge-dind")
  dind.privileged = true
  dind.tasks = [
    "dockerd-entrypoint.sh &",
    "echo waiting && sleep 20",
    "ps -ef",
    "docker version",
    "killall dockerd"
  ]
  dind.run().then( () => {
    console.log("==== DONE ====")
  })
})
```

So it's a start.

technosophos:
I'm hearing that the DinD method is "working, but slowish" for people. Does it make sense for me to go ahead and try another PR that will expose the host docker socket, or should we just call it good with DinD?

bacongobbler:
I think it'd be better to make a PR to mount the host docker socket instead. It's easier to maintain since the docker engine has been optimized for the cloud provider and we won't need to constantly maintain updating/upgrading the docker daemon container image (and add new support for engine flags). It's overall a significantly easier maintenance burden.

As long as users have a way to run a dind container themselves, that should be more than OK.

See https://github.com/Azure/draft/pull/434

technosophos:
Closing in favor of #154

PR #122, opened 2017-11-11
fix(api,controller,worker): use ULID for ID

### LINKED ISSUE #121: Store.GetBuildJobs returns jobs which are not part of the build
When calling Store.GetBuildJobs with a build executed by brig it will return all jobs executed by brig for the same project.

The reason for this is that Store.GetBuildJobs uses the combination of project and commit as a unique identifier, but brig sets the commit to "master" for every builds.

The code lines responsible:
https://github.com/Azure/brigade/blob/32599784d6b4c591e44225d02c1fc17d65e5f32f/pkg/storage/kube/job.go#L36

and:
https://github.com/Azure/brigade/blob/32599784d6b4c591e44225d02c1fc17d65e5f32f/brigade-client/cmd/brig/commands/run.go#L112-L133

technosophos:
There were a few places where instead of using the ULID as an ID, the
system was using project + commit as an ID. That, however, is not
unique. This was resulting in incorrect data being returned by the API
server.

One of the reasons for the above was that job pods were not being
labeled with the ULID (or with the worker ID). So there was no way to
correlated a job with its worker parent or build. So this adds two new
labels on a job pod: worker and build. The build ID is also added to the
PVC for build storage.

Finally, the API server has been fixed to use the build ID instead of
the commit for finding associated jobs.

Closes #121

### DIFF
--- a/brigade-controller/cmd/brigade-controller/controller/controller_test.go
+++ b/brigade-controller/cmd/brigade-controller/controller/controller_test.go
@@ -23,7 +23,7 @@ func TestController(t *testing.T) {
 	})
 	config := &Config{
 		Namespace:        v1.NamespaceDefault,
-		WorkerImage:      "deis/brgiade-worker:latest",
+		WorkerImage:      "deis/brigade-worker:latest",
 		WorkerPullPolicy: string(v1.PullIfNotPresent),
 	}
 	controller := NewController(client, config)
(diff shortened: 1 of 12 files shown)


technosophos:
I updated this PR with a notable change: I modified the Controller to reject any Secret that does _not_ have a `build` label. @adamreese I am not sure if this breaks anything in the larger architecture. But since `storage` automatically assigns a ULID when it creates a new build, this should work fine with both Brig and Gateway. External gateways, though, might need to be modified.

PR #135, opened 2017-11-14
fix(chart): fix project chart's missing sidecar value

technosophos:
No default value was placed in the sidecar, which cases the base64 enc
operation to fail. Added blank default value.

Closes #131

PR #137, opened 2017-11-14
fix(chart): fix if-checks and add if wrappers to all optional values

lukepatrick:
Improved fix for #135


technosophos:
I'm cool with this version once it's rebased.

Looks like we all had approximately the same idea at approximately the same time.

lukepatrick:
rebased

bacongobbler:
oof, looks like the commit history got munged up in the rebase. Can you try the following?

```
$ git remote add upstream git@github.com:Azure/brigade
$ git checkout upstream/master
$ git checkout -b fix-135
$ git cherry-pick 25af0af79eb1d379ee91df721fc5fc7c478502f0
$ git push origin fix-135
```

And open a new PR? That would be the easiest way to fix this. :)

PR #141, opened 2017-11-15
fix reference to brigade-server-brigade-gw

bacongobbler:
See https://github.com/Azure/brigade/blob/db6d0f4e51182d6865320fd523bfc1e24ff64d5a/chart/brigade/templates/_helpers.tpl#L19


bacongobbler:
har har

PR #158, opened 2017-11-21
Use alpine edge to get git@2.15.0

soyuka:
Hi there,

At ApiPlatform we want to use Brigade to run some of our tests that require more power then a limit travis instance. 
We have an issue with `git@2.13.5` while executing:

```
+ git fetch origin VCS_REVISION
error: Server does not allow request for unadvertised object VCS_REVISION
```

where VCS_REVISION is a given `sha1` from github servers.

Link to the brigade PR on our repository: https://github.com/api-platform/core/pull/1507

This command does work with `git@2.15.0` which is only available with `alpine:edge` version.

### DIFF
--- a/git-sidecar/Dockerfile
+++ b/git-sidecar/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.6
+FROM alpine:edge
 
 RUN apk update && apk add --no-cache \
     ca-certificates \


technosophos:
While we're normally pretty reluctant to do `edge` versions, I think that given how early we are in the dev process it will be okay for now.

PR #164, opened 2017-11-22
feat(worker): allow disabling privileged mode at project level

technosophos:
This makes it possible to disable the ability of a job to enter
privileged mode. A project can now disable the feature for all jobs.


brigade-worker/test/mock.ts
```diff
@@ -17,7 +17,9 @@ export function mockProject(): Project {
     kubernetes: {
       namespace: "default",
       vcsSidecar: "deis/git-sidecar:latest"
-    }
+    },
+    allowPrivilegedJobs: true,
+    allowHostMounts: false,
```
bacongobbler: nice fix. 👍

PR #166, opened 2017-11-22
docs(design): explain why we use secrets instead of CRDs

technosophos:


technosophos:
@bacongobbler That was my first _EVER_ PR from a Windows machine!

bacongobbler:
😱

PR #171, opened 2017-11-23
fix: delete the dead link in the Github status payload

### LINKED ISSUE #126: GitHub status payload has invalid url
When a status is set on the commit, a URL is provided with it. However this URL appears to be broken:

```
{
  "id": 4312453694,
  "sha": "11426f2c689a6a7ccb78a6c6e366ce040e5f797d",
  "name": "j2sol/demobrigade",
  "target_url": "http://169.48.217.138:32586/log/j2sol/demobrigade/id/11426f2c689a6a7ccb78a6c6e366ce040e5f797d",
  "context": "brigade",
  "description": "Brigade build passed",
  "state": "success",
```

That IP and port is indeed the exposed service for the Brigade gateway service, but that URL results in a 404.

meyerbaptiste:
This PR fixes #126.

### DIFF
--- a/pkg/webhook/github.go
+++ b/pkg/webhook/github.go
@@ -6,9 +6,7 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
-	"net/url"
 	"os"
-	"path"
 
 	"github.com/google/go-github/github"
 	"gopkg.in/gin-gonic/gin.v1"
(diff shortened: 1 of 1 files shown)


dunglas:
It was convenient to have a webpage with the logs. Is it intended to remove it?

technosophos:
The new version will be out by the end of this week (just in time for KubeCon)

technosophos:
Gonna confirm with @adamreese that we can merge this. Then I will.

adamreese:
I am fine deleting this until we can fix properly

technosophos:
Okay, we can now set status from inside of a brigade.js like this: https://github.com/technosophos/github-notify

PR #178, opened 2017-11-29
feat(gateway): make forked pull-request builds optional

adamreese:
Makes building pull requests from forked repos optional.  Disabled by
default.

### DIFF
--- a/pkg/webhook/github.go
+++ b/pkg/webhook/github.go
@@ -23,9 +23,10 @@ const (
 )
 
 type githubHook struct {
-	store        storage.Store
-	getFile      fileGetter
-	createStatus statusCreator
+	store                   storage.Store
+	getFile                 fileGetter
+	createStatus            statusCreator
+	buildForkedPullRequests bool
 }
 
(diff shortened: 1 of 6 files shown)


adamreese:
Good point, let me look into that quick

technosophos:
Re-LGTM with the env var change

adamreese:
Merging this PR.  I will start a WIP for disabling at the project level

pkg/webhook/github.go
```diff
@@ -23,21 +23,23 @@ const (
...
 
 // NewGithubHook creates a GitHub webhook handler.
-func NewGithubHook(s storage.Store) *githubHook {
+func NewGithubHook(s storage.Store, buildForkedPullRequests bool) *githubHook {
 	return &githubHook{
-		store:        s,
-		getFile:      getFileFromGithub,
-		createStatus: setRepoStatus,
+		store: s,
+		buildForkedPullRequests: buildForkedPullRequests,
```
cordoval: is not the second argument, is the fourth right?

bacongobbler: This is for initializing a struct, not a function. Fields are not positional.

PR #179, opened 2017-11-29
fix(api): improve performance of fetching builds

technosophos:
This massively improves the performance of requests that fetch
lists of build data. It pre-fetches both secrets and pods for the
build data.

### DIFF
--- a/pkg/storage/kube/build.go
+++ b/pkg/storage/kube/build.go
@@ -82,13 +82,18 @@ func (s *store) GetBuilds() ([]*brigade.Build, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	podList, err := s.client.CoreV1().Pods(s.namespace).List(lo)
+	if err != nil {
+		return nil, err
+	}
+
 	buildList := make([]*brigade.Build, len(secretList.Items))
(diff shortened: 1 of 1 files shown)


pkg/storage/kube/build.go
```diff
@@ -82,13 +83,18 @@ func (s *store) GetBuilds() ([]*brigade.Build, error) {
...
 	buildList := make([]*brigade.Build, len(secretList.Items))
 	for i := range secretList.Items {
 		b := NewBuildFromSecret(secretList.Items[i])
-		b.Worker, err = s.GetWorker(b.ID)
-		if err != nil {
-			return buildList, err
-		}
+		// The error is ErrWorkerNotFound, and in that case, we just ignore
+		// it and assign nil to the worker.
+		b.Worker, _ = findWorker(b.ID, podList)
```
adamreese: should this check if the error is `ErrWorkerNotFound`?

technosophos: I _could_, though it is currently the only possible error.

technosophos:
Switched from error to bool on findWorker.

PR #190, opened 2017-12-01
fix(charts): add roles to services, deployments, pods

### LINKED ISSUE #167: Chart: Add `role:` labels to Brigade API, GW, and Controller deployments and services
We need to add:

```yaml
labels:
  role: <ROLE>
```

to the API server, GW, and controller in the chart so that they can be retrieved easily via kubectl commands.

technosophos:
This adds the role label to several chart members, and cleans up some of
the other labels.

Closes #167


technosophos:
To test:

```
$ kubectl get deployment -l role=api
$ kubectl get po -l role=controller
$ ...
```

Issue #161, opened 2017-11-21
Switch to Alpine:3.7 when it is available


technosophos:
We switched the vcs sidecar to use `alpine:edge` because of a Git bug. When 3.7 comes out, we should switch back to `alpine:3.7`.

See #158

PR #192, opened 2017-12-01
fix(*): use alpine:3.7

adamreese:
Closes: #161


adamreese:
holding until image is published

PR #198, opened 2017-12-04
differentiate between github and dockerhub gateways

bacongobbler:
without this, both the dockerhub and github gateway services will be load balancing all gateways.


bacongobbler:
note that this will require a full re-deploy of brigade; `helm upgrade` cannot patch a deployment's `spec.selector` fields due to an upstream limitation: https://github.com/kubernetes/helm/issues/2493

PR #201, opened 2017-12-05
feat(gateway): add several new webhook events

### LINKED ISSUE #127: GitHub gateway should silently ignore events which are not handled
Currently when an event comes into the web hook that the gateway does not handle (anything other than `ping`, `push`, or `pull_request`), an `http.StatusBadRequest` is [returned](https://github.com/Azure/brigade/blob/master/pkg/webhook/github.go#L59). This in turn makes the GitHub project web hook settings show a littering of what looks like errors, specifically from things like `status` events, which wind up being generated by Brigade itself.

It would be better to silently accept the set of events from GitHub that COULD be sent, and only act on ones that Brigade wants to. The rest should just result in a quiet 200, like the `ping` event.

It's not an error that Brigade doesn't do anything with the events that can come at it.

technosophos:
This adds `commit_comment`, `create`, `release`, and `status`

Closes #127


pkg/webhook/github.go
```diff
@@ -94,6 +95,20 @@ func (s *githubHook) handleEvent(c *gin.Context, eventType string) {
 		}
 		repo = e.Repo.GetFullName()
 		commit = e.PullRequest.Head.GetSHA()
+	case *github.CommitCommentEvent:
+		repo = e.Repo.GetFullName()
+		commit = *e.Comment.CommitID
```
adamreese: We may need to validate what CommitID actually is.  For example you can comment on a single commit in a pull request.

technosophos: Yeah, I'm not actually sure how we want to handle this one. In most cases, it sorta seems like a useless hook to have. I almost removed it... then I figured I might as well leave it, since somebody somewhere is gonna have a use for it.

PR #211, opened 2017-12-11
docs(readme) Add link to Kashti.

colemickens:
I guess actually hitting "Create pull request" would help.

cc: @squillace

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,9 @@ for Kubernetes.
 - Easy construction of pipelines
 - Check out the [docs](/docs/) to get started.
 
+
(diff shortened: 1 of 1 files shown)


squillace:
LGTM. +1. @technosophos or @flynnduism or someone can merge. Cole's larger point is right: we should have a bit more describing how and where these work together. I'll follow up.

seanknox:
Thanks Cole!

technosophos:
Thanks! Man, it's like we don't know how to advertise for our own stuff. ;-)

PR #214, opened 2017-12-11
docs(scripting): Actually introduce `brig`

mgalgs:
Currently we reference having introduced the brig tool:

    Using the `brig` tool introduced above, we can see how this works.

but no such introduction currently exists.  Add an introduction to reduce
confusion.


mgalgs:
Actually I guess there is a brief mention of `brig` above, but it doesn't help the user figure out where to obtain `brig`, so I think the added intro would still be good.

PR #215, opened 2017-12-12
docs(intro): require `brigadier` in examples.

jdhuntington:
Noticed the intro snippets don't work against the current build as-is; the `require` statement is necessary.

### DIFF
--- a/docs/intro/tutorial04.md
+++ b/docs/intro/tutorial04.md
@@ -89,7 +89,7 @@ Brigade uses simple JavaScript files to run tasks. When it comes to task running
 Given this, the role of the `brigade.js` file is to declare event handlers. And it's easy. Open `brigade.js` and write this JavaScript code into it:
 
 ```javascript
-const { events, Job } = require("brigadier");
+const { events } = require("brigadier");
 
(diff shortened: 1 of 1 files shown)


technosophos:
@bacongobbler you wanna just verify my LGTM on this one?

docs/intro/tutorial04.md
```diff
@@ -89,6 +89,8 @@ Brigade uses simple JavaScript files to run tasks. When it comes to task running
 Given this, the role of the `brigade.js` file is to declare event handlers. And it's easy. Open `brigade.js` and write this JavaScript code into it:
 
 ```javascript
+const { events, Job } = require("brigadier");
```
bacongobbler: I think this can be simplified to

```
const { events } = require("brigadier");
```

Because the job object isn't actually added or used until the following step.

bacongobbler:
minor nitpick but still semantically correct, so I'm inclined to just merge this. Merge at your discretion, @technosophos :)

docs/intro/tutorial04.md
jdhuntington: Good call; updated.

Issue #228, opened 2017-12-19
vcs-sidecar fails when passed a branch name and commit hash


blimmer:
I encountered these failures when trying to build a branch with brigade.

## Branch Name
I noticed that the vcs-sidecar had a familiar exception (128) and that the whole job failed because of this. Here's the output from that container:

```console
kubectl logs do-nothing-1513714578597-test-ecr --namespace operations vcs-sidecar
+ : /src
+ : git@github.com:blimmer/k8s_deploy_demo.git
+ : test-ecr-pull
+ git clone --depth=50 git@github.com:blimmer/k8s_deploy_demo.git /src
Cloning into '/src'...
Warning: Permanently added 'github.com,192.30.253.113' (RSA) to the list of known hosts.
+ cd /src
+ git fetch origin test-ecr-pull
Warning: Permanently added 'github.com,192.30.253.113' (RSA) to the list of known hosts.
From github.com:blimmer/k8s_deploy_demo
 * branch            test-ecr-pull -> FETCH_HEAD
+ git checkout -qf FETCH_HEAD
+ git reset --hard -q test-ecr-pull
fatal: ambiguous argument 'test-ecr-pull': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

This problem occurs locally for me, too.

The solution is that, in the case of branches, you need to pass the `--branch=<branch-name>` parameter along with the `--depth` parameter. This is because the `--depth` parameter in `git` implies `--single-branch`, as well ([docs](https://git-scm.com/docs/git-clone#git-clone---depthltdepthgt))

The tricky part is that the `--branch` parameter can only accept named tags and branches, and cannot take a commit hash, so switching logic will need to be added for this case.

## Commit Hash
This one exited with exit code (1).

```console
+ : /src
+ : git@github.com:blimmer/k8s_deploy_demo.git
+ : 1353a9d889ae7ec6c53304c2abe893aa0aaee5d1
+ git clone --depth=50 git@github.com:blimmer/k8s_deploy_demo.git /src
Cloning into '/src'...
Warning: Permanently added 'github.com,192.30.253.112' (RSA) to the list of known hosts.
+ cd /src
+ git fetch origin 1353a9d889ae7ec6c53304c2abe893aa0aaee5d1
Warning: Permanently added 'github.com,192.30.253.112' (RSA) to the list of known hosts.
[...]

PR #230, opened 2017-12-19
fix(git-sidecar): don't shallow clone repository

blimmer:
See issue #228 .

Doing a shallow clone either requires us to know if the `VCS_REVISION` parameter is a branch or named tag. If it is a branch or named tag, the script should include the `--single-branch=VCS_REVISION` along with the depth parameter. If it's a commit hash, I'm pretty sure you have to do an entire clone to support commit hashes that are not at the tip of a given branch.

### DIFF
--- a/git-sidecar/rootfs/clone.sh
+++ b/git-sidecar/rootfs/clone.sh
@@ -6,7 +6,7 @@ set -x
 : "${VCS_REPO:?}"
 : "${VCS_REVISION:?}"
 
-git clone --depth=50 "${VCS_REPO}" "${VCS_LOCAL_PATH}"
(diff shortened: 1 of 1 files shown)


technosophos:
I know we're sorta going back and forth on this one, so I will leave the issue open long enough to get some comments. But I am tending to agree here unless there is some other flag that will get us a shallow clone and then fetch all branches.

blimmer:
Yeah, to the extent that I know the git commands I think you have to know if you're trying to shallow clone a branch or a direct commit. If we knew that, we could easily write a switch in this script.

blimmer:
@adamreese is going to open a PR to resolve the shallow clone issue. I'm going to close this up in favor of that PR.

PR #231, opened 2017-12-19
docs(gateways): add doc on creating Brigade gateways

### LINKED ISSUE #227: How to work with cron jobs?
The docs or the announcement blog post mention that brigade is a good fit to run scheduled jobs. Is there an example of integration with Kubernetes cron jobs?

technosophos:
Closes #227

### DIFF
--- a/docs/topics/examples/simple-event/simple-event.sh
+++ b/docs/topics/examples/simple-event/simple-event.sh
@@ -1,0 +1,67 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# The Kubernetes namespace in which Brigade is running.
+namespace="default"
+
+event_provider="simple-event"
+event_type="my_event"
+
+# This is github.com/deis/empty-testbed
+project_id="brigade-830c16d4aaf6f5490937ad719afd8490a5bcbef064d397411043ac"
+commit="master"
+
+b64flags=""
+uuidflags=""
+system=$(uname)
+if [[ $system != "Darwin" ]]; then
+  b64flags="-w 0" # Turn off line wrapping
+  uuidflags="-t"  # generate UUID v1 for sortability
+fi
+
+
+# This is the brigade script to execute
+script=$(cat <<EOF
+const { events } = require("brigadier");
+events.on("my_event", (e) => {
+  console.log("The system time is " + e.payload);
+});
+EOF
+)
+
+# Now we will generate a new event evrey 60 seconds.
+while : ; do
+  # We'll use a UUID instead of a ULID. But if you want a ULID generator, you
+  # can grab one here: https://github.com/technosophos/ulid
+  uuid="$(uuidgen $uuidflags | tr '[:upper:]' '[:lower:]')"
+
+  # We can use the UUID to make sure we get a unique name
+  name="simple-event-$uuid"
+
+  # This will just print the system time for the system running the script.
+  payload=$(date)
+
+  cat <<EOF | kubectl --namespace ${namespace} create -f -
+  apiVersion: v1
+  kind: Secret
+  metadata:
+    name: ${name}
+    labels:
+      heritage: brigade
+      project: ${project_id}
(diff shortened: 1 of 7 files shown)


docs/topics/gateways.md
```diff
@@ -0,0 +1,322 @@
...
+  labels:
+    # 'heritage: brigade' is mandatory, and signals that this is a Brigade event.
+    heritage: brigade
+
+    # This should point to the Brigade project ID in which this event is to be
+    # executed
+    project: brigade-1234567890
+
+    # This MUST be a unique ID. Where possible, it SHOULD be a ULID
+    # Substituting a UUID is fine, though some sorting functions won't be as expected
```
mbarison: what is the purpose of sorting in this context?

docs/topics/gateways.md
```diff
@@ -0,0 +1,322 @@
...
+    build: 01C1R2SYTYAR2WQ2DKNTW8SH08
+
+    # Commit is the commitish/reference for any associated VCS repository. By
+    # default, this should be `master` for Git.
+    commit: master
+
+    # 'component: build' is REQUIRED and tells brigade to start a new build.
```
mbarison: What if I don't want to build stuff? Maybe I just want to schedule tests on an already deployed release.

docs/topics/examples/simple-event/simple-event.sh
```diff
@@ -0,0 +1,59 @@
...
+    name: ${name}
+    labels:
+      heritage: brigade
+      project: ${project_id}
+      build: ${uuid}
+      commit: ${commit}
+      component: build
+  type: Opaque
+  data:
+    commit: $(echo -n "${commit}" | base64)
```
mbarison: you must add -w 0

docs/topics/gateways.md
bacongobbler: Projects like [kashti](https://github.com/Azure/kashti) display the jobs in a UI. By using a ULID we can display them in a logical order (such as by job creation time).

docs/topics/gateways.md
bacongobbler: We're using the term "build" to reference a "job", not necessarily to "build things".

docs/topics/examples/simple-event/simple-event.sh
```diff
@@ -0,0 +1,59 @@
...
+  console.log("The system time is " + e.payload);
+});
+EOF
+)
+
+# Now we will generate a new event evrey 60 seconds.
+while : ; do
+  # We'll use a UUID instead of a ULID. But if you want a ULID generator, you
+  # can grab one here: https://github.com/technosophos/ulid
+  uuid="$(uuidgen | tr '[:upper:]' '[:lower:]')"
```
mbarison: if you use `uuidgen -t` then your UUID is time based, thus sortable.

docs/topics/examples/simple-event/simple-event.sh
technosophos: Yeah, you are right. Alpine and Debian require -w 0 b/c their built-in wrap is 80.

docs/topics/examples/simple-event/simple-event.sh
technosophos: My version of uuidgen does not have a `-t`. But you are right -- a UUIDv1 generator will make it at least sorta sortable.

docs/topics/examples/simple-event/simple-event.sh
technosophos: Ugh... and macOS/BSD's base64 doesn't have a `-w`. It uses `-b` instead, and defaults to unwrapped. I'll do a little primitive OS checking to handle this.

technosophos:
Okay, I think the new commit deals with all of the comments above. Notably, I do some OS detection now, and fixed some of the wording to be clearer about what certain things are doing.

PR #233, opened 2017-12-19
fix(cr-gateway): support branch names with slashes

### LINKED ISSUE #229: cr-gateway cannot accept branch names with slashes
The Git Flow method requires branch names to be prefixed with `<type>/<name>`, e.g. `feature/foobar`. However, the cr-gateway interprets these slashes as part of the restful path and 404s.

```console
 ~  curl -H "Content-Type: application/json" -X POST localhost:8000/events/webhook/blimmer/k8s_deploy_demo/test-ecr-pull
{"status":"Success"}%                                                                                                                                        ~  curl -H "Content-Type: application/json" -X POST localhost:8000/events/webhook/blimmer/k8s_deploy_demo/feature/foobar
404 page not found%                                                                                                                                                                                                                                                                           ~  curl -H "Content-Type: application/json" -X POST localhost:8000/events/webhook/blimmer/k8s_deploy_demo/feature%2Ffoobar
404 page not found%
```

It would be nice if we could URL encode or otherwise pass these kinds of branch names.

technosophos:
This adds support for a commit=? query param. It also adds support for
specifying project by its long (hashed) name instead of org/project.

Closes #229

### DIFF
--- a/brigade-cr-gateway/cmd/brigade-cr-gateway/server.go
+++ b/brigade-cr-gateway/cmd/brigade-cr-gateway/server.go
@@ -48,11 +48,26 @@ func newRouter(store storage.Store) *gin.Engine {
 	router := gin.New()
 	router.Use(gin.Recovery())
 
+	handler := webhook.NewDockerPushHook(store)
+
 	events := router.Group("/events")
 	{
 		events.Use(gin.Logger())
 
-		events.POST("/webhook/:org/:project/:commit", webhook.NewDockerPushHook(store).Handle)
+		// We need to handle the full project name (brigade-00000), the org/project
+		// format of the name (for backward compatibility), and variants where the
+		// commitish has to be supplied as a param.
+
+		// Of the form /webhook/brigade-123456789?commit=master
+		// Here, :org is actually a full project name, but due to Gin's naming rules
+		// we have to keep it named :org.
+		// This is the recommended form.
+		events.POST("/webhook/:org", handler.Handle)
+
+		// Of the form /webhook/deis/empty-testbed?commit=master
+		events.POST("/webhook/:org/:repo", handler.Handle)
+		// Of the form /webhook/deis/empty-testbed/master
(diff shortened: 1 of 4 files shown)


brigade-cr-gateway/cmd/brigade-cr-gateway/server.go
```diff
@@ -48,11 +48,18 @@ func newRouter(store storage.Store) *gin.Engine {
...
 
+	handler := webhook.NewDockerPushHook(store)
+
 	events := router.Group("/events")
 	{
 		events.Use(gin.Logger())
 
-		events.POST("/webhook/:org/:project/:commit", webhook.NewDockerPushHook(store).Handle)
+		// Of the form /webhook/brigade-123456789?commit=master
+		events.POST("/webhook/:org", handler.Handle)
```
blimmer: Should this `:org` be `:repo`? In this example it looks like you're skipping the org and just passing the repo / commit as a query param.

Or should this example be passing the org in the path, with the repo and commit as query params?

technosophos: Gin has a quirk where the positional params must have the same name. So even though it _is_ a repo ID, I have to name it `:org` to satisfy Gin's naming rules.

blimmer: Interesting - you might note that in a comment in the code. That was not obvious to me (but I don't know Gin very well 😄 ).

bacongobbler: it could be changed if we used a different handler function, but yeah, @technosophos is spot on.

This also threw me off for a second before I realized this was to work around that quirk. +1 on adding a comment to note that in this situation, org == repo.

technosophos:
Okay, per @blimmer's suggestion, I added a detailed comment about the URL forms.

PR #243, opened 2018-01-02
feat(api): add /v1/build/{id}/log to get logs.

technosophos:
This exposes build logs on the API server, parallel to job logs.

Also, some testing code was backfilled to provide a storage mock and a
way to test at least some of the API server operations.

### DIFF
--- a/pkg/api/build.go
+++ b/pkg/api/build.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"io"
 	"net/http"
 
 	"gopkg.in/gin-gonic/gin.v1"
@@ -39,4 +40,33 @@ func (api Build) Jobs(c *gin.Context) {
 		return
 	}
 	c.JSON(http.StatusOK, jobs)
+}
+
+// Logs creates a new gin handler for the GET /job/:id/logs endpoint
+func (api Build) Logs(c *gin.Context) {
+	id := c.Params.ByName("id")
+	build, err := api.store.GetBuild(id)
+	if err != nil {
+		c.JSON(http.StatusNotFound, struct{}{})
+		return
+	}
+	if c.Query("stream") == "true" {
+		logReader, err := api.store.GetWorkerLogStream(build.Worker)
+		if err != nil {
+			c.JSON(http.StatusNotFound, struct{}{})
+			return
+		}
+		defer logReader.Close()
+		io.Copy(c.Writer, logReader)
+	} else {
+		logs, err := api.store.GetWorkerLog(build.Worker)
+		if err != nil {
+			c.JSON(http.StatusNotFound, struct{}{})
+			return
+		}
+		if len(logs) == 0 {
+			c.JSON(http.StatusNoContent, nil)
+		}
+		c.JSON(http.StatusOK, logs)
+	}
 }
--- a/pkg/storage/mock/storage.go
+++ b/pkg/storage/mock/storage.go
@@ -1,0 +1,152 @@
+package mock
+
+import (
+	"bytes"
+	"io"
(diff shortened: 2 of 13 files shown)


pkg/api/build.go
```diff
@@ -40,3 +41,29 @@ func (api Build) Jobs(c *gin.Context) {
...
+
+// Logs creates a new gin handler for the GET /job/:id/logs endpoint
+func (api Build) Logs(c *gin.Context) {
+	id := c.Params.ByName("id")
+	build, err := api.store.GetBuild(id)
+	if err != nil {
+		c.JSON(http.StatusNotFound, struct{}{})
+		return
+	}
+	if c.Query("stream") == "true" {
```
bacongobbler: nice touch with adding the `stream` query. 👍

pkg/api/build.go
```diff
@@ -40,3 +41,29 @@ func (api Build) Jobs(c *gin.Context) {
...
+		}
+		defer logReader.Close()
+		io.Copy(c.Writer, logReader)
+	} else {
+		logs, err := api.store.GetWorkerLog(build.Worker)
+		if err != nil {
+			c.JSON(http.StatusNotFound, struct{}{})
+			return
+		}
+		c.JSON(http.StatusOK, logs)
```
bacongobbler: Should we check for a length of zero and return a 204 No Content when there's no logs to ship back? Might make it a bit more user-friendly.

technosophos: /cc @seanknox Would a 204 return code help or hurt? In this implementation, it will actually send back a `""\n`, which is a JSON-encoded empty string.

pkg/api/build.go
```diff
@@ -40,3 +41,29 @@ func (api Build) Jobs(c *gin.Context) {
...
+	id := c.Params.ByName("id")
+	build, err := api.store.GetBuild(id)
+	if err != nil {
+		c.JSON(http.StatusNotFound, struct{}{})
+		return
+	}
+	if c.Query("stream") == "true" {
+		logReader, err := api.store.GetWorkerLogStream(build.Worker)
+		if err != nil {
+			c.JSON(http.StatusNotFound, struct{}{})
```
seanknox: Does an error here always indicate that `GetWorkerLogStream()` couldn't find the log stream?

pkg/api/build.go
seanknox: 204 return code would be good to do as it'll make the client parsing code in kashti clearer (e.g. `if responseCode == 204` vs `if response.body == null`).

pkg/storage/mock/storage.go
```diff
@@ -0,0 +1,152 @@
...
+	"io"
+	"io/ioutil"
+	"time"
+
+	"github.com/Azure/brigade/pkg/brigade"
+)
+
+var (
+	// Now is the date used in all stub date fields.
+	Now = time.Now()
```
seanknox: Curious, are time-based race conditions a problem with go tests? In Ruby/JS specs it's common to bind "now" to a known value to deal with freezing/jumping time (e.g. https://github.com/travisjeffery/timecop). Just a wonder, not blocking this PR whatsoever.

technosophos: Not to the extent it was in rails.

pkg/api/build.go
technosophos: It means it could not fetch the log stream. We don't know all the errors that could be returned from k8s, though. All we have seen are missing logs due to clean ups

technosophos:
Okay, now it returns 204 when no log data is returned, and I added tests for this condition.

PR #245, opened 2018-01-05
docs(scripting): add advanced events section

technosophos:
This adds a section that covers advanced event handling, like nesting
events, firing events, and declaring multiple event handlers.

It also adds information on using DinD instead of the Docker socket.


adamreese:
This is so cool

PR #248, opened 2018-01-08
feat(brig): add -c/--commit flag to brig run

technosophos:
This allows you to specify a commit/branch/tag from the Brig
commandline.


technosophos:
I did one quick fix to remove a fmt.Printf that was unnecessary.

PR #249, opened 2018-01-09
docs(readme): Feedbacks on README for the brig command

mumoshu:
As far as I have tested with 0.7.0:

- There seems like `brig run my-org/my-project` but not `brig my-org/my-project`
- There seems like `--namespace` option but not `--ns` for the `brig run` command


adamreese:
Thanks!

Issue #225, opened 2017-12-19
GitHub access not possible without an oAuth token


blimmer:
## Background

While testing out a proof of concept with a private GitHub repo, I noticed that my calls to the cr-gateway were failing with this message:

```
aborting DockerImagePush event due to error: GET https://api.github.com/repos/blimmer/k8s_deploy_demo/contents/?ref=master: 401 Bad credentials []
```

The culprit causing this message is here:

https://github.com/Azure/brigade/blob/51ebbe1b177e9c8d67a71c0eb951556c3a3471fd/pkg/webhook/client.go#L90-L92

The webhook tries to confirm that the `brigade.js` file is present before responding to the webhook.

https://github.com/Azure/brigade/blob/7b2f3836620ac13862e5cee4793a86cfac6b8a2e/pkg/webhook/dockerhub.go#L57-L59

Note that the comment there isn't exactly right - it does not clone the repo (which would work for my use-case because I provide a [GitHub deploy key](https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys)) that allows read-only ssh access to only this repo), it uses the GitHub API to check for the presence of that file.

## Security Concerns

I definitely understand why it's not desirable to hold the webhook while you clone the entire repo to check for the presence of the `brigade.js` file. However, the OAuth scope required for the current API call requires `repo` scope ([docs](https://developer.github.com/apps/building-oauth-apps/scopes-for-oauth-apps/#available-scopes)) which, in my opinion, provides **way too much** access to all of my private repos, and additional permissions far beyond the read access intended.

> Grants read/write access to code, commit statuses, invitations, collaborators, adding team memberships, and deployment statuses for public and private repositories and organizations.

Leaking that single key (which is visible to others in my organization by viewing the helm chart for the `brigade-project`) is undesirable.

The other option is a [GitHub deploy key](https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys), which seems much more suited to this task. It provides per-repo, read-only (by default) access to just the one repo in question. It also feels like brigade fits really nicely in with the rest of the kinds of apps configured in this section.

![screen_shot_2017-12-18_at_16_25_51](https://user-images.githubusercontent.com/630449/34133805-124832f0-e414-11e7-8eb1-6d027fd92d76.png)

## Options

These are a few options I've thought about for this:

#### Skip the `brigade.js` file check
[...]

PR #255, opened 2018-01-09
fix(cr-gateway): Remove GitHub API requirement from CR server.

technosophos:
Relates to #225

### DIFF
--- a/charts/brigade-project/values.yaml
+++ b/charts/brigade-project/values.yaml
@@ -12,8 +12,7 @@ cloneURL: "https://github.com/deis/empty-testbed.git"
 sharedSecret: "IBrakeForSeaBeasts"
 
 # OPTIONAL: Use this to have Brigade update your project about the build.
-# You probably want this if you want pull requests or commits to show
-# the build status.
+# This is REQUIRED for the GitHub gateway, but optional otherwise.
 github: {}
 #   token: "github oauth token"
 
--- a/pkg/webhook/dockerhub.go
+++ b/pkg/webhook/dockerhub.go
@@ -63,34 +63,24 @@ func (s *dockerPushHook) Handle(c *gin.Context) {
 		return
 	}
 
-	// This will clone the repo before responding to the webhook. We need
-	// to make sure that this doesn't cause the hook to hang up.
-	brigadeJS, err := s.getFile(commit, "./brigade.js", proj)
-	if err != nil {
-		log.Printf("aborting DockerImagePush event due to error: %s", err)
-		c.JSON(http.StatusBadRequest, gin.H{"status": "brigadejs not found"})
-		return
-	}
-
-	go s.notifyDockerImagePush(proj, commit, body, brigadeJS)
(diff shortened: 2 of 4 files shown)


pkg/webhook/dockerhub.go
```diff
@@ -63,34 +63,24 @@ func (s *dockerPushHook) Handle(c *gin.Context) {
...
 
-func (s *dockerPushHook) doDockerImagePush(proj *brigade.Project, commit string, payload, brigadeJS []byte) error {
+func (s *dockerPushHook) doDockerImagePush(proj *brigade.Project, commit string, payload []byte) error {
 	b := &brigade.Build{
 		ProjectID: proj.ID,
 		Type:      "image_push",
 		Provider:  "dockerhub",
 		Commit:    commit,
 		Payload:   payload,
-		Script:    brigadeJS,
```
mumoshu: Can we really just defer determining the script to be executed for a build like this?

As far as I understand, brigade-controller as of today doesn't have an ability to populate missing brigade.js(=`script` in a secret representing an brigade event) - it looks like we have to populate it on gateway-side. 

I'm writing this referring the followings:

- https://github.com/Azure/brigade/search?utf8=%E2%9C%93&q=syncSecret&type=
- https://github.com/Azure/brigade/search?utf8=%E2%9C%93&q=Script&type=

Please correct me if I'm missing something!

mumoshu: Just noticed that brigade does handle this case [here](https://github.com/Azure/brigade/blob/2d6281b12ed97d70ddb675dc79a5e70dc8722583/brigade-controller/cmd/brigade-controller/controller/handler.go#L137-L162).
Sorry for bothering!

For anyone interested, in nutshell brigade seems to create a vcs-sidecar(as an init-container) to populate `/vcs/brigade.js` only when `script` isn't included in an event. Then, the brigade-worker runs either `/etc/brigade/script`(corresponds to the `script` included in a `brigade.Build`=secret) or `/vcs/brigade.js` [according to this code](https://github.com/Azure/brigade/blob/2d6281b12ed97d70ddb675dc79a5e70dc8722583/brigade-worker/prestart.js). So, if we omit a `script` in a brigade.Build coming from the cr-gateway, the worker just tries to git-clone the necessary script=`brigade.js` from the vcs repo. No problem.

charts/brigade-project/values.yaml
```diff
@@ -12,8 +12,7 @@ cloneURL: "https://github.com/deis/empty-testbed.git"
 sharedSecret: "IBrakeForSeaBeasts"
 
 # OPTIONAL: Use this to have Brigade update your project about the build.
-# You probably want this if you want pull requests or commits to show
-# the build status.
+# This is REQUIRED for the GitHub gateway, but optional otherwise.
```
blimmer: 👍

pkg/webhook/dockerhub.go
technosophos: Thanks for taking the time to look through all of that.

PR #257, opened 2018-01-10
fix(git-sidecar): support branch names and tags

### LINKED ISSUE #228: vcs-sidecar fails when passed a branch name and commit hash

adamreese:
Lookup revision in remote repository to get SHA

fixes #228

### DIFF
--- a/git-sidecar/rootfs/clone.sh
+++ b/git-sidecar/rootfs/clone.sh
@@ -6,6 +6,10 @@ set -x
 : "${VCS_REPO:?}"
 : "${VCS_REVISION:?}"
 
+if sha=$(git ls-remote --exit-code "${VCS_REPO}" "${VCS_REVISION}" | cut -f1); then
+  VCS_REVISION="${sha}"
+fi
+
 git clone --depth=50 "${VCS_REPO}" "${VCS_LOCAL_PATH}"
 cd "${VCS_LOCAL_PATH}"
 
--- a/git-sidecar/.dockerignore
+++ b/git-sidecar/.dockerignore
@@ -1,3 +1,4 @@
 cmd
+test
 README.md
 Dockerfile
--- a/git-sidecar/test/test.sh
+++ b/git-sidecar/test/test.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+root_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+cd "$root_dir"
+
+tempdir=$(mktemp -d -t "repos")
+
+export VCS_LOCAL_PATH="${tempdir}/repo"
+
+check_equal() {
+  [[ "$1" == "$2" ]] || {
+    echo >&2 "Check failed: '$1' == '$2' ${3:+ ($3)}"
+    exit 1
+  }
+}
+
+setup_git_server() {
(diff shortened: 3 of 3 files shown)


git-sidecar/rootfs/clone.sh
```diff
@@ -6,6 +6,10 @@ set -x
 : "${VCS_REPO:?}"
 : "${VCS_REVISION:?}"
 
+if sha=$(git ls-remote --exit-code "${VCS_REPO}" "${VCS_REVISION}" | cut -f1); then
```
mumoshu: Nit but shouldn't VCS_REVISION be renamed to e.g. `VCS_REF` as it doesn't stand only for REVISION now?

adamreese: I have another PR that fixing the naming

blimmer:
Thank you @adamreese !

PR #259, opened 2018-01-10
Populate some GitHub specific env var, use them for cloning

dunglas:
Fixes #230.

This PR adds some GitHub specific environment variables (when available, as most other CI systems do), then use them to fix the shallow cloning problem.

(/cc @blimmer, I saw you're working on related topics)


git-sidecar/rootfs/clone.sh
```diff
@@ -5,10 +5,27 @@ set -x
...
-git clone --depth=50 "${VCS_REPO}" "${VCS_LOCAL_PATH}"
-cd "${VCS_LOCAL_PATH}"
-
-git fetch origin "${VCS_REVISION}"
-git checkout -qf FETCH_HEAD
-git reset --hard -q "${VCS_REVISION}"
+if [ "$GITHUB_PULL_REQUEST" ]; then
+    git clone --depth=50 "${VCS_REPO}" "${VCS_LOCAL_PATH}"
+    cd "${VCS_LOCAL_PATH}"
+    git fetch origin "+refs/pull/${GITHUB_PULL_REQUEST}/merge:"
```
mumoshu: Hi! Thanks for the enhancement!

Would this change still allows us to build not only "the commit w/ the pull-request merged into the base branch" but also "the commit which is HEAD of the pull-request"?

I'm wondering if this breaks the latter - which I believe is the existing behavior.

brigade-worker/src/brigadier.ts
```diff
@@ -52,6 +53,11 @@ export class Job extends jobImpl.Job {
   run(): Promise<jobImpl.Result> {
     let jr = new JobRunner(this, currentEvent, currentProject);
     this._podName = jr.name;
+
+    if ("github" === currentEvent.provider) {
```
blimmer: I could imagine this block becoming a giant switch statement as more gateways are developed.

```javascript
switch(currentEvent.provider) {
  case "github":
    something();
  case "trello":
    somethingElse();
}
```

I wonder if there's a way that this logic could live within the gateway instead of within the worker? I kind of like the brigade-worker not needing to know about individual gateways. That would also allow authors of custom gateways to add this information to their respective payloads.

mumoshu: Good point!

I agree that we would be better avoiding tight-coupling between worker<->gateway.

What I'm not clear about though is that how we could implement it cleanly.
What we could extend here would be only `currentEvent` and `currentProject`.
AFAIK, the latter former is passed as-is from gateways and the latter is passed as-is from the project config.
So - there would be no straight-forward way to provide envs to be used by workers from the gateway-side?

mumoshu: Update: In my understanding, brigade works like this:

1. The GitHub gateway receives a webhook and [creates a corresponding `brigade.Build`](https://github.com/Azure/brigade/blob/f6c8b7aa3166d390852f0d34638424ca90e8ed6b/pkg/webhook/github.go#L213-L222)
2. BrigadeController receives the `brigade.Build` and [creates a worker to run it](https://github.com/Azure/brigade/blob/2d6281b12ed97d70ddb675dc79a5e70dc8722583/brigade-controller/cmd/brigade-controller/controller/handler.go#L46).

The step 2. is especially relevant. For example, we could add `Env` to the `brigade.Build` to be populated with [these](https://github.com/Azure/brigade/pull/259/files#diff-ca57cc779014810cd541e48951772a56R10) in the gateway, which is then passed to a worker container's `env` as-is. Do you think it is feasible?

mumoshu: @technosophos Would you be ok with adding `Env` to `brigade.Build` as above?

blimmer: This isn't a blocker right now from my perspective, just something to consider.

mumoshu: @dunglas Thanks for the sharing the context 👍 Yes, this seems like no blocker for me, too.

Perhaps once we get some time and/or another use-cases requiring `brigade.Build` to have  custom Env vars populated in a gateway, we can submit an another PR to do so?

mumoshu:
Hi @dunglas, would you mind rebasing this?

dunglas:
@mumoshu done

mumoshu:
Hi @dunglas, sorry for the slow action on my side.

Would #257 fixes the original issue #228 which this PR intended to?

dunglas:
Partially yes. This PR goes 1 step further by mimicking the Travis's behavior with Pull Request.

mumoshu:
@dunglas Oh, the part that you build against the merge commit rather than the head of a pull request, right?

I remember that travis tend to build against two commits per pull request, one for the merge commit and one for the head of the pr. If I'm correct, would we ideally be better allow the user to choose which commit to be built, merge and/or head?

Sorry for the long roundtrip on this. I'm still interested in how we could support your use-case(which seems very reasonable to me).

mumoshu:
Hi @dunglas, in case my sentence wasn't clear to you - I guess you see after reading [the travis doc about "double builds"](https://docs.travis-ci.com/user/pull-requests/#%E2%80%98Double-builds%E2%80%99-on-pull-requests).

In nutshell, we'd need to trigger 2 brigade builds per a pull-request's [`opened` and `synchronized`](https://developer.github.com/v3/activity/events/types/#pullrequestevent) event. One for the HEAD of the branch, one for the merge commit.

Brigade builds against the HEAD of branch as of today. I understand that your PR changes this to the merge commit.
Both builds have its own benefit so my best bet is to allow both, hence my suggestion to trigger 2 brigade builds per the PR events.

WDYT?

P.S. I can implement this on my own once confirmed your intention and thoughts on my suggestion.

krancour:
This is quite stale, so I am closing. Please do re-open if it's worth reviving.

Issue #256, opened 2018-01-10
GitHub Integration: Support for more event types


mumoshu:
According to the [code](
https://github.com/Azure/brigade/blob/08c49800425b5fd3babd2e98540fcd9d6fde6de9/pkg/webhook/github.go#L53), it misses support for types like:

- label
- deployment
- deployment_status

Especially, deployment* events are useful for my brigade pipeline to react on a gh's deployment
event(can be done with a simple curl) by triggering an actual deployment, instead of running a brig command or kubectl to trigger a brigade event directly.
It also provides nice notifications on the pull-request containing the git commit the deployment event refers, so that it looks more like a gh-native feature imho.

Would you mind if I contribute it?
Would there be more event types which should be supperted?
Any input would be welcomed. Thanks!

technosophos:
We'll gladly accept PRs for any of the GitHub hooks. I am implementing them based on my guesses about what people want.

technosophos:
Should we leave this open?

mumoshu:
@technosophos Thanks for checking-in. No, I believe this can be closed now. Perhaps someone could open a separate issues for adding another event types with associated use-cases 😃

mumoshu:
For anyone interested - When I had open this, I had somehow misunderstood that brigade doesn't support "pull request labeled/unlabeled" events. No, it isn't true.

It does emit `pull_request:labeled` or `pull_request:unlabeled` for the events. See `pkg/webhook/github.go` for more details!

PR #260, opened 2018-01-10
feat(gateway): Support for two event types: deployment and deployment_status

mumoshu:
Ref #256

Just a minimum viable change for now. Probably I need to write tests?


technosophos:
This looks great. Adding a unit test for a new GitHub hook should be really easy. You'll just need to store an example webhook JSON payload in `pkg/webhook/testdata` and then add an entry here:
https://github.com/Azure/brigade/blob/master/pkg/webhook/github_test.go#L109

I got the example payloads from either the GitHub API documentation or by just running the hook and using the GitHub dashboard to get the payload it sent.

mumoshu:
@technosophos Thanks for your support! I've addressed examples in testdata and tests.
Please let me know if there's anything I can do to get this merged.

technosophos:
Merged. Thanks again!

Issue #254, opened 2018-01-09
Quickstart example fails on `brig run`


shimmerjs:
Working through the Quickstart docs, everything works fine up until `brig run -f brigade.js deis/empty-testbed`.

- Brigade server is successfully installed from helm chart
- Deployed `brigade-project` out of the `charts/` directory with default `values.yaml`
- copy/pasted `brigade.js` contents from Quickstart guide
- when running `brig run -f brigade.js`, 

```
brig --verbose run -f ./brigade.js deis/empty-testbed
panic: runtime error: slice bounds out of range

goroutine 1 [running]:
github.com/Azure/brigade/brig/cmd/brig/commands.(*scriptRunner).send(0xc4202fd140, 0x7ffeefbff683, 0x12, 0xc420158a00, 0x7a, 0x27a, 0x0, 0x1ce36a2)
	/Users/alex/go/src/github.com/Azure/brigade/brig/cmd/brig/commands/run.go:132 +0x446
github.com/Azure/brigade/brig/cmd/brig/commands.glob..func5(0x23fc700, 0xc4203dbc80, 0x1, 0x4, 0x0, 0x0)
	/Users/alex/go/src/github.com/Azure/brigade/brig/cmd/brig/commands/run.go:84 +0xe5
github.com/Azure/brigade/vendor/github.com/spf13/cobra.(*Command).execute(0x23fc700, 0xc4203dbb80, 0x4, 0x4, 0x23fc700, 0xc4203dbb80)
	/Users/alex/go/src/github.com/Azure/brigade/vendor/github.com/spf13/cobra/command.go:599 +0x3e9
github.com/Azure/brigade/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0x23fb820, 0xc420072058, 0x0, 0x1)
	/Users/alex/go/src/github.com/Azure/brigade/vendor/github.com/spf13/cobra/command.go:689 +0x339
github.com/Azure/brigade/vendor/github.com/spf13/cobra.(*Command).Execute(0x23fb820, 0x23fb820, 0xc4204dff50)
	/Users/alex/go/src/github.com/Azure/brigade/vendor/github.com/spf13/cobra/command.go:648 +0x2b
main.main()
	/Users/alex/go/src/github.com/Azure/brigade/brig/cmd/brig/main.go:11 +0x31
```

mumoshu:
[...]

PR #262, opened 2018-01-10
docs(brig): note to checkout a stable tag

lukepatrick:
Additional commentary to help with #254

### DIFF
--- a/docs/topics/developers.md
+++ b/docs/topics/developers.md
@@ -33,9 +33,9 @@ Follow these steps when cloning the brigade repository to use an existing `GOPAT
 - `$ cd $(go env GOPATH)/src/github.com/Azure/brigade`
 
 **Note**: this leaves you at the tip of **master** in the repository where active development
-is happening. Likely you want to checkout the most recent stable tag:
+is happening. You might prefer to checkout the most recent stable tag:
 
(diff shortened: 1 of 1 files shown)


lukepatrick:
@bacongobbler Would make sense to have this more as a user-guide, though don't want to be over doing the duplication of similar instructions. All of the user guide docs tell you to come to this doc to get the brig binary.

docs/topics/developers.md
```diff
@@ -32,6 +32,11 @@ Follow these steps when cloning the brigade repository to use an existing `GOPAT
 - `$ git clone https://github.com/Azure/brigade $(go env GOPATH)/src/github.com/Azure/brigade`
 - `$ cd $(go env GOPATH)/src/github.com/Azure/brigade`
 
+**Note**: this leaves you at the tip of **master** in the repository where active development
+is happening. Likely you want to checkout the most recent stable tag:
+
+- `$ git checkout v0.7.0`
+
```
technosophos: Can we change that last phrase to "you might prefer to checkout the most recent stable tag"? I feel like that covers @bacongobbler 's concern (and mine), but still points brig-builders to the right path.

lukepatrick: No problem, I've added the update.

PR #271, opened 2018-01-18
feat(worker): allow overriding of the job's service account.

### LINKED ISSUE #251: Service account per Job
I'd wanted to associate a service account with enough permissions to deploy things to k8s, only to the job runs a k8s deployment.

However, as far as I can see from code, [brigade seems to hard-code `brigade-worker` serviceaccount to be used by all the job runner pod](https://github.com/Azure/brigade/blob/8346887653a41ca7879801e64d1f3f20561273b6/brigade-worker/src/k8s.ts#L585). It would be great if it could be overridden like `job.serviceaccount = "deployer"` in brigade.js for this use-case.

technosophos:
This allows you to set a `job.serviceAccount` that Kubernetes will use
to run the job.

Closes #251

### DIFF
--- a/brigade-worker/src/job.ts
+++ b/brigade-worker/src/job.ts
@@ -163,6 +163,17 @@ export abstract class Job {
    */
   public privileged: boolean = false;
 
+  /** The account identity to be used when running this job.
+   * This is an optional way to override the build-wide service account. If it is
+   * not specified, the main worker service account will be used.
+   *
+   * Different Brigade worker implementations may choose to honor or ignore this
+   * for security or configurability reasons.
+   *
+   * See https://github.com/Azure/brigade/issues/251
+   */
+  public serviceAccount: string;
+
   /**
    * host expresses expectations about the host the job will run on.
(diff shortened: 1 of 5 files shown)


brigade-worker/src/job.ts
```diff
@@ -163,6 +163,16 @@ export abstract class Job {
    */
   public privileged: boolean = false;
 
+  /** The account identity to be used when running this job.
+   * This is an optional way to override the build-wide service account. If it is
+   * not specified, the main worker service account will be used.
+   *
+   * Implementors may choose to ignore this value for security reasons.
```
mumoshu: This wasn't clear for me at glance. Probably behind the word "Implementators", there is an implication that there could be multiple implementations of brigade-worker which runs the brigade.js?
If thats the case, perhaps "Different brigade-worker implementations may choose to ..." sounds like a better note for me.

technosophos: Good point. I will change that.

PR #275, opened 2018-01-23
chore(gateway): Rename brigade-gateway to brigade-github-gateway

### LINKED ISSUE #264: Rename of GitHub gateway (brigade-gateway -> brigade-github-gateway)
Now that gateways are more "microserice"-y (see #99), it might make sense to rename the github gateways.

I think it used to make sense when the gateway did both github and container registry, but I think a rename would be appropriate to make it clear what that gateway handles. It wasn't immediately obvious to me that the `brigade-gateway` was the github gateway until I dug into the code.

This ticket also relates to #263 as they might make sense to do together.

mumoshu:
Resolves #264


charts/brigade/values.yaml
```diff
@@ -73,7 +73,7 @@ worker:
 gw:
```
mumoshu: I'd rather want to refrain from a ad-hoc rename of `gw` to e.g. `github` here, I'll probably address it in an another PR once we get some consensus on it.

Would it be nice to reorganize this and `cr` altogether under a new `gateway` key like:

```yaml
gateway:
  github:
    enabled: true
    registry: deis
    name: brigade-github-gateway
    #tag
    #pullPolicy: IfNotpresent
    #buildForkedPullRequests: true
  cr:
    enabled: false
     # ...
```

mumoshu:
Verified it works after renaming by running:

```
$ make docker-build

$ helm upgrade --install --set gw.tag=v0.9.0-1-g027fcbe mybrig1 ./charts/brigade
```

charts/brigade/templates/_helpers.tpl
```diff
@@ -16,7 +16,7 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 {{- end -}}
 
 {{- define "brigade.gw.fullname" -}}
-{{ include "brigade.fullname" . | printf "%s-gw" }}
+{{ include "brigade.fullname" . | printf "%s-github-gw" }}
```
mumoshu: I had considered to name it `gh-gw` after the existing `cr-gw` for consistency but refrained from doing so.
Perhaps in the future, we are likely to add `bitbucket-gw`, `gitlab-gw` and so on which doesn't have a well-known abbrev form? Then, a name like `gh-gw` will lose consistency very soon.

technosophos: I agree with your decision.

Issue #240, opened 2017-12-22
Thoughts on Reusability in Javascript Files?


blimmer:
We plan to use brigade for a portion of our Continuous Delivery pipeline. That means that many of our projects will have very similar base `brigade.js` files defining the common pipeline.

Some of this re-usability will come from our common containers we use, however, if there are common tasks that we want to pipeline, it will require duplicated code in the `brigade.js` file.

For example, you could imagine a bunch of projects having this kind of brigade file:

```javascript
const { events, Job, Group } = require("brigadier")

events.on("exec", () => {
  var securityCheck = new Job("security-check", "our-k8s-deploy-container:latest", ["security_check"])
  var validateImage = new Job("validate-image", "our-k8s-deploy-container:latest", ["validate_image"])
  var somethingElse = new Job("something-else", "our-k8s-deploy-container:latest", ["something_else"])

  Group.runAll([securityCheck, validateImage, somethingElse]).then(() => {
     var deploy = new Job("deploy", "our-k8s-deploy-container:latest", ["deploy"])
     deploy.run();
  });
});
```

Because we want to parallelize all of the work done initially, we can't abstract this logic away into the container 100%. It'd be great if we could instead require a custom library that could keep all projects up to date with our current deployment pipeline. For instance:

```javascript
const { ourDeploy } = require("our-deploy-node-package")

events.on("exec", () => {
  ourDeploy.doEverything(); // this executes the Brigadier commands, but taking advantage of upstream improvements to the common node package (maybe adding new checks, documentation publishing, etc.)
});
```

As a platform infrastructure team, having the ability to continuously improve all consumers of our deployment without touching their repos is a big benefit in long-term maintainability.

What are the Brigade team's thoughts on something like this?

technosophos:
I am very much open to ideas on this. One thing I have been pondering is allowing `require` to look up things in the Git repository that is associated with the project. The syntax would probably be:

```
const mylib = require("./mylib")
```
The question is then what we resolve `./mylib` to. Assuming `$GIT` is the path to the Git checkout of the Brigade project's associated VCS repo, I can thing of a few options:

- `$GIT/mylib`: Treat all libraries as "local" JS files
- `$GIT/node_modules/mylib`: Treat all libraries as Node libraries
- `$GIT/brigade_modules/mylib`: Invent a class of Brigade libraries

[...]

PR #277, opened 2018-01-23
feat(worker): Allow using a specific brigade-worker image per project

mumoshu:
`worker.(registry|name|tag|pullPolicy)` are added to brigade-project's values.yaml so that
you can differentiate which docker image to be used for running brigade.js per-project basis.

Ref #240

This is a minimum change seemingly required in order to support reusability of node modules, possibly hosted in private git repositories, plus arbitrary binaries usable from within brigade.js'.


technosophos:
This can definitely help with the JavaScript library problem. And it's useful in general, since we do want to make it easy to write custom workers. So I'm definitely in favor of adding this. I'm going to give @adamreese a chance to review before I merge it though.

To get all of the JS support we want for #240, we might also need to remove the whitelist code for `require` in `index.ts`.

mumoshu:
> To get all of the JS support we want for #240, we might also need to remove the whitelist code for require in index.ts.

Thanks for the info!
I've addressed it in the last commit.

brigade-worker/src/require.ts
```diff
@@ -1,24 +1,6 @@
...
-  // library.
-  // {name: "crypto"},
-  // buffer.allocUnsafe could potentially allow other parts of the program to be
-  // accessed. This could presumably include something like a Kubernetes token.
-  // The present security model does not consider this to be a critical issue,
-  // since this information is obtainable via other pods as well. So we leave
-  // the buffer module. In the future, it might be good to mask that function.
-  { name: "buffer" },
-  { name: "assert" }
+  { name: "brigadier", override: "./brigadier" }
```
technosophos: Excellent. /cc @blimmer

technosophos:
One of my recent merges introduced a merge conflict. Would you mind rebasing?

mumoshu:
@technosophos Sure! Just finished rebasing. Would you mind taking a look again?

mumoshu:
Hi @adamreese, this is just a friendly reminder that we're looking forward to your review. Thanks!

pkg/brigade/project_test.go
```diff
@@ -43,3 +43,37 @@ func TestProjectSecrets(t *testing.T) {
 		t.Error("Project.Repo.SSHKey should not be exported")
 	}
 }
+
+func TestProjectWorkerConfig(t *testing.T) {
```
adamreese: What is this testing?

mumoshu: This is testing that `WorkerConfig` can be serialized to/deserialized from JSON.
But I'm open to suggestion like e.g. merging this test into [`TestProjectSecrets`](https://github.com/Azure/brigade/pull/277/files/c6cfb753a2ed9eb6a9e2e47dec549e14d5a742eb#diff-df2235e76b97ddcdf1c446d032d54964R19)!

charts/brigade-project/values.yaml
```diff
@@ -102,3 +102,11 @@ kubernetes:
...
 # Uncomment the line below to increase logging verbosity.
 # workerCommand: "yarn start"
+
+# OPTIONAL: Project-specific worker settings which takes precedence over brigade-wide settings
+# Useful when you want a specific brigade-worker image for running brigade.js of this project.
+worker:
+  #registry: deis
+  #name: brigade-worker
+  #tag:
+  #pullPolicy: IfNotPresent
```
mumoshu: I believe this `worker.*` keys should be consistent vs `workerCommand`.
I'll open a dedicated github issue for discussion if/once this gets merged.

mumoshu: Also, we should have a general rule like "If two or more keys are prefixed by the same word, we should make them nested under the word".

PR #281, opened 2018-01-24
Configurable Build Storage ; Initialize Submodules

brandon-bethke-neudesic:
The Build Shared Storage Space is now user configurable through the brigade project file. The default value is 50Mi

Resolves: https://github.com/Azure/brigade/issues/220


brandon-bethke-neudesic:
Second commit resolves https://github.com/Azure/brigade/issues/269
Add user configurable option in the brigade project to allow git submodules to be initialized.

PR #284, opened 2018-01-26
feat(worker): make storage class for cache and build configurable

### LINKED ISSUE #279: Feature: Add the ability to set Storage Class for PVC
Currently the 'default' storage class is used when creating a PVC. The ability to set storage class like [azurefile](https://docs.microsoft.com/en-us/azure/aks/azure-files-dynamic-pv) would be great since now you have to override the 'default' storage class in AKS & ACS. 

I would be happy to implement it :)

technosophos:
Closes #279

### DIFF
--- a/brigade-worker/test/k8s.ts
+++ b/brigade-worker/test/k8s.ts
@@ -30,12 +30,14 @@ describe("k8s", function() {
       assert.equal(p.name, "github.com/deis/test-private-testbed")
       assert.equal(p.repo.name, "deis/test-private-testbed")
       assert.equal(p.repo.cloneURL, "https://github.com/deis/empty-testbed.git")
-      assert.equal(p.repo.initGitSubmodules, "true");
+      assert.isTrue(p.repo.initGitSubmodules);
       assert.equal(p.repo.token, "pretend password\n")
       assert.equal(p.kubernetes.namespace, "default")
       assert.equal(p.kubernetes.vcsSidecar, "vcs-image:latest")
       assert.property(p.secrets, "hello")
       assert.equal(p.secrets.hello, "world")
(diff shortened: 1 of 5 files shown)


brigade-worker/test/k8s.ts
```diff
@@ -186,8 +188,8 @@ describe("k8s", function() {
           // We also have to check this for the sidecar pod. But right now this is
           // embedded in a string.
           let sidecar = jr.runner.spec.initContainers[0]
-          assert.equal(sidecar.env.length, 5)
-          assert.equal(sidecar.env[4].name,"BRIGADE_REPO_KEY", "Has BRIGADE REPO KEY as param")
+          assert.equal(sidecar.env.length, 6)
+          assert.equal(sidecar.env[5].name,"BRIGADE_REPO_KEY", "Has BRIGADE REPO KEY as param")
```
technosophos: ^^ These tests were broken on master. When a new env var was added in a recent PR, this test did not get updated.

technosophos:
@mumoshu if you have time, would you be willing to review this PR?

charts/brigade-project/values.yaml
```diff
@@ -68,3 +68,12 @@ allowPrivilegedJobs: "true"
 # OPTIONAL: Use this to allow host mounted docker sockets in your jobs.
 # This is a big security risk if your project is public-facing; enable at your own risk.
 # allowHostMounts: "true"
+
+# Kubernetes-specific configuration options.
+kubernetes: {}
```
mumoshu: Nit but this seems to work without `{}`. Then, perhaps we'd rather like to remove it, so that you can just uncomment any of the settings below to start customizing values?

PR #285, opened 2018-01-26
feat(brig): build binaries at release

technosophos:
This automates the process of building binaries for Brig and uploading
them to a GitHub release.


brigade.js
```diff
@@ -42,7 +44,61 @@ function build(e, project) {
...
+    "ls -lh",
+    "dep ensure",
+    "make build-release",
+    `github-release release -t ${tag} -n "${parts[1]} ${tag}"`
+  ]
+
+  // Upload for each target that we support
+  for (const f of ["linux-amd64", "windows-amd64", "darwin-amd64"]) {
+    const name = binName + "-"+f
+    cx.tasks.push(`github-release upload -R -f ./bin/${name} -n ${name} -t ${tag}`)
```
mumoshu: Could we extract this release-upload sequence to e.g. `make publish-release`, just for the sake of consistency with `make build-release` and a bit more testability(it will be testable in isolation with brigade)?

mumoshu: Or If you've intended to parallelize uploads with `Group.runAll` in the future, I believe we'd better leave this as-is for now.

brigade.js
```diff
@@ -42,7 +44,61 @@ function build(e, project) {
   ];
 
   // Run in parallel
-  Group.runAll([jsTest, goBuild])
+  Group.runAll([jsTest, goBuild]).then( () => {
+    const gh = JSON.parse(e.payload)
+    if (e.event == "push" && gh.ref.startsWith("refs/tags/")) {
```
mumoshu: Seems like an useful idiom for deployment/release automation 👍

brigade.js
mumoshu: Nit: As `github-release release` fails when the release already exists, foreseeing uploads to replace existing ones by `github-release upload -R` seemed unnecessary. I believe `-R` can be safely omitted.

technosophos: Fixed.

PR #286, opened 2018-01-26
WIP feat(tests): add brigade target to run functional test in cluster

technosophos:
This adds an event to brigade.js that lets us run the functional tests
inside of a cluster.

### DIFF
--- a/tests/tests_test.go
+++ b/tests/tests_test.go
@@ -11,6 +11,12 @@ import (
 )
 
 func TestFunctional(t *testing.T) {
+
+	host := os.Getenv("BRIGADE_BRIGADE_GITHUB_GW_SERVICE_HOST")
+	if host == "" {
+		host = "localhost"
+	}
+
 	githubPushFile, err := os.Open("testdata/test-repo-generated.json")
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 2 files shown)


technosophos:
If anyone has some feedback on this, it'd be appreciated.

The idea is roughly this: You should be able to run the functional tests _inside of the cluster_ by executing a Brigade command:

```
$ brig run -f ./brigade.js -e functional_test Azure/brigade
```

Right now, there is one sticking point: The name of the Brigade cluster (as installed by Helm) must be `brigade`. The reason for this is that I haven't found a good way yet to generate the service discovery key to look up the in-cluster IP address.

If anyone has some ideas for how to work around that, I'd love to hear them. Maybe the easiest would just be to pass that info in during `brig run`... I don't know.

tests/tests_test.go
```diff
@@ -11,6 +11,12 @@ import (
 )
 
 func TestFunctional(t *testing.T) {
+
+	host := os.Getenv("BRIGADE_BRIGADE_GITHUB_GW_SERVICE_HOST")
```
mumoshu: So, we can set this to e.g. `brigade-brigade-github-gw` to actually run the test against the brigade cluster? (Great!)

mumoshu:
@technosophos Code LGTM. But anything you would like to finish before this gets out of WIP?

technosophos:
The env var that I rely upon is created by Kubernetes, but only if I install Brigade with the name `brigade` (e.g. via `helm install -n brigade ...`). I think I should make it more generic/configurable.

technosophos:
Revising the process to be more accomodating, we could:

1. Do a DinD build of Brigade
2. Run a `helm install -n integration-test --namespace foo...`
3. Execute the tests (which means we need to have them do a DNS lookup of the gh gw)
4. `finally()` we need to `helm delete --purge`

radu-matei:
This is out of date, and rebasing would be rather difficult - closing, but I'd be happy to help anyone looking at adding functional tests to the new `brigade.js`.

PR #290, opened 2018-01-26
fix(chart): add rbac for vacuum

technosophos:
This adds a service account, role, and role binding for vacuum.

### DIFF
--- a/charts/brigade/templates/vacuum-cronjob.yaml
+++ b/charts/brigade/templates/vacuum-cronjob.yaml
@@ -1,10 +1,10 @@
-{{ if .Values.vacuum.enabled }}
+{{ if .Values.vacuum.enabled }}{{ $fullname := include "brigade.vacuum.fullname" .}}
 apiVersion: batch/v1beta1
 kind: CronJob
 metadata:
-  name: {{ template "brigade.fullname" . }}-vacuum
+  name: {{ $fullname }}
   labels:
-    app: {{ template "brigade.fullname" . }}-vacuum
+    app: {{ template "brigade.fullname" . }}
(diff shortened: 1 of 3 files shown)


charts/brigade/templates/vacuum-cronjob.yaml
```diff
@@ -19,6 +19,7 @@ spec:
             app: {{ template "brigade.fullname" . }}
             role: vacuum
         spec:
+          serviceAccountName: {{ $fullname }}
```
mumoshu: Note: At glance, I couldn't understand why we're explicitly setting serviceAccountName here. But this is also done elsewhere, in `gateway-cr-role.yaml` for example, and won't break anything.

technosophos:
Closes #301

charts/brigade/templates/vacuum-cronjob.yaml
technosophos: This is to apply RBACs consistently. We try to create a service account for any thing that needs to access the Kubernetes API. That way there is always a distinctive handle for people to write RBACs against.

mumoshu: Thanks for the clarification!
Made sense a lot.

For anyone interested: 
Originally, I had wanted to point out that if the serviceAccountName should be set explicitly only
when rbac.enabled, so that it looks like `{{ if .Values.vacuum.enabled }}serviceAccountName: {{ $fullnamn }}`.

However, `rbac.enabled: false` doesn't always imply RBAC is disabled on your cluster.

It is just set to false when you want to manage roles and bindings yourself, regardless
of RBAC is disabled on the cluster or not.

So, set it to false, write your own roles
and bindings against the service account always created by helm-installing brigade, and then you are done.

Never noticed even I've seen many charts authored this way. Great design!

PR #292, opened 2018-01-29
added ingress for api service

alex-egorov:
added ingress for api service

### DIFF
--- a/charts/brigade/values.yaml
+++ b/charts/brigade/values.yaml
@@ -61,12 +61,9 @@ api:
     type: ClusterIP
     externalPort: 7745
     internalPort: 7745
-  # Configure liveness probes except `httpGet` and the belongings
-  #livenessProbe:
-  #  initialDelaySeconds: 20
-  # Configure readiness probes except `httpGet` and the belongings
-  #readinessProbe:
-  #  initialDelaySeconds: 20
+  ingress:
+    enabled: false
+    hosts: []
 
 # worker is the JavaScript worker. These are created on demand by the controller.
 worker:
--- a/charts/brigade/templates/api-ingress.yaml
+++ b/charts/brigade/templates/api-ingress.yaml
@@ -1,0 +1,32 @@
+{{- if .Values.api.ingress.enabled -}}
+{{- $serviceName := include "brigade.api.fullname" . -}}
+{{- $servicePort := .Values.api.service.externalPort -}}
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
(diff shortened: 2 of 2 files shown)


charts/brigade/values.yaml
```diff
@@ -61,12 +61,9 @@ api:
     type: ClusterIP
     externalPort: 7745
     internalPort: 7745
-  # Configure liveness probes except `httpGet` and the belongings
-  #livenessProbe:
-  #  initialDelaySeconds: 20
-  # Configure readiness probes except `httpGet` and the belongings
-  #readinessProbe:
-  #  initialDelaySeconds: 20
```
mumoshu: Probably you've mistakenly removed this block?

mumoshu:
Hi @alex-egorov, thanks for your contribution!

Probably you've intended to support the use-case of https://github.com/Azure/kashti/blob/angular5/docs/install.md#with-ssltls?

If that's the case, would you mind adding an standard/example setup in `values.yaml` so that users won't innocently expose the brigade apiserver to the internet, which is a big security risk IMHO?

charts/brigade/values.yaml
alex-egorov: yes, it's my mistake

mumoshu:
@alex-egorov Please let me know if there is anything I can help

alex-egorov:
@mumoshu Thanks for help. Actually, I didn't plan to use tls. This pr is intended to share brigade apiserver through ingress, in order to use Kashti UI in public (without  any actions like `kubectl port-forward`, which require secure information), like [this](https://azure.github.io/kashti/). Security issues are really presented and should be solved by means of tokens/secrets in bgigade-kashti communications.

mumoshu:
Thanks for the explanation!

Just to sync up with you, even the kashti instance you've mentioned seems to utilize TLS for accessing brigate api from your browser.

However, I do agree that it would slightly make the problem smaller if your intended use-case is exposing it as a public pipeline.

I'd still suggest using TLS through your external LB in front of the ingress pods, though! Anyway, if youre securing kashti with TLS, your browser would complain when the ui tried to access api without TLS?

technosophos:
@mumoshu Should we merge as-is and then add TLS?

mumoshu:
@technosophos Yes, I believe so. Please take my comments as suggestion for future enhancements, not a blocker. This LGTM at
this point thanks to @alex-egorov!

// Basically, I prefer splitting a big topic into  PRs of manageable sizes like this, so that we can get the ball rolling.

PR #296, opened 2018-02-06
Do not set the storage class to "default" by default ; Destroy Build Storage PVC on error

brandon-bethke-neudesic:
Resolves Issue https://github.com/Azure/brigade/issues/287
Resolves Issue https://github.com/Azure/brigade/issues/293
Do not set the storage class to "default" by default. 
Do not set the kube storage class if it is not specified.
Destroy Build Storage PVC on error


mumoshu:
@brandon-bethke-neudesic Merged. Thanks as always for your contribution 👍

PR #299, opened 2018-02-07
feat(gateway): add support for default brigade script

### LINKED ISSUE #250: Reacting on GitHub events from a repository not containing brigade.js
Hi, thanks for maintaining this 👍 

To react events from GitHub, brigade seems to require a brigade.js in the root of each GitHub repository. This is slightly problematic to me because then I need to copy-paste or maintain a set of exactly same brigade.js across various projects.

So, I'd wanted to prepare a "default" brigade.js which is used when the repository which has emitted a github/brigade event misses its own brigade.js, so that I don't need this copy-pasting.

However, as far as I understand, brigade's GitHub-integration as of today seems to not allow this use-case out-of-box as [the GitHub gateway expects the repo emitted the event has a `brigade.js` in the repository root](https://github.com/Azure/brigade/blob/25b4e7990bcb02697ac7bbedd168beefef89c0f2/pkg/webhook/github.go#L208) and there is not way to inject brigade.js from an another source.

It seems like I can write my own brigade gateway based on the brigade's default gh gateway, which would load a default brigade.js from perhaps a global k8s secret containing it rather than the repo corresponds to the event coming from the gh gateway.

// My understanding behind this idea is that I could create a k8s secret representing [a build](https://github.com/Azure/brigade/blob/435cb75130758a3dda1638879525868c0dec1776/pkg/brigade/build.go#L20) which accepts an arbitrary "Script" to be executed by brigade, right?

But anyway, would it make sense it to support this use-case in the default GitHub gateway included in brigade? If so, any guidance/suggestion on how we should design it/other related use-cases should be covered together/etc.?

mumoshu:
Resolves #250

### DIFF
--- a/charts/brigade-project/templates/secret.yaml
+++ b/charts/brigade-project/templates/secret.yaml
@@ -16,6 +16,9 @@ data:
   sharedSecret: {{ .Values.sharedSecret | b64enc }}
   cloneURL: {{ .Values.cloneURL | b64enc }}
   initGitSubmodules: {{ default "false" .Values.initGitSubmodules | b64enc }}
+  {{ if .Values.defaultScript }}
+  defaultScript: {{.Values.defaultScript | b64enc }}
+  {{- end }}
   {{ if .Values.vcsSidecar }}
   vcsSidecar: {{ .Values.vcsSidecar | b64enc }}
   {{- end }}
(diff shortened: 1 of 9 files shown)


technosophos:
@mumoshu Now that you are an official maintainer, when other maintainers LGTM your PRs, we will leave them alone and let you merge them yourself. This way, you are always the final decision maker on your own PRs.

So feel free to merge this whenever you are ready.

mumoshu:
@technosophos Noted. Thanks for your guidance and support!

PR #306, opened 2018-02-10
fix(chart): turn vcsSidecar on by  default

### LINKED ISSUE #305: Worker does not correctly fall back to VCS when no script is provided
It appears that the init container definition is missing for the worker, so `/vcs` never gets the contents fetched by vcs-sidecar.

technosophos:
This enables the vcsSidecar param by default in values.yaml. This will
cause the worker to use the vcsSidecar as an init container. Omiting it
prevents the worker from declaring a sidecar pod, which means passing
the brigade.js file is incumbent upon the gateways.

Closes #305

### DIFF
--- a/charts/brigade-project/values.yaml
+++ b/charts/brigade-project/values.yaml
@@ -56,8 +56,12 @@ secrets:
 # you know what you are doing.
 # namespace: "default"
 
-# OPTIONAL: vcsSidecar is the image that fetches a repo from a VCS
-# vcsSidecar: "deis/git-sidecar:latest"
+# RECOMMENDED: vcsSidecar is the image that fetches a repo from a VCS
+# The default sidecar uses Git to fetch a copy of the project. Commenting this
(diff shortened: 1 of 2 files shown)


charts/brigade-project/values.yaml
```diff
@@ -56,8 +56,12 @@ secrets:
 # you know what you are doing.
 # namespace: "default"
 
-# OPTIONAL: vcsSidecar is the image that fetches a repo from a VCS
-# vcsSidecar: "deis/git-sidecar:latest"
+# RECOMMENDED: vcsSidecar is the image that fetches a repo from a VCS
+# The default sidecar uses Git to fetch a copy of the project. Commenting this
+# out will prevent the worker from using a sidecar. This may improve performance
+# very slightly, but will break some gateways or cause the default script to
+# be used.
```
mumoshu: Appreciate covering relationships with both the gateway and the default script.
I believe that this definitely helps us understand the gotachas without consuming too much time!

PR #307, opened 2018-02-12
feat(*): use git references rather than commit SHA

adamreese:
This change will fetch git references so they can be accessed during a
build. References are fetched directly into a new repo to avoid having
to perform a shallow clone.

This is a breaking change.

- The Build secret structure changed to allow for SHA and Reference.
- Environment variables for the sidecar have been renamed for consistency.


brigade-worker/src/events.ts
```diff
@@ -70,6 +65,25 @@ export class BrigadeEvent {
...
+export class Revision {
+  /**
+   * commit is the upstream VCS commit ID (revision).
+   *
+   * If it is not provided, it may be interpreted as `master`, or the head
+   * of the main branch.
+   *
+   * The default value is not guaranteed to be `master` in future versions.
+   */
+  commit?: string;
```
mumoshu: As you describe this in the comment above, I believe this should be `id` rather than `commit`.
A bit more context is that, in my mental model, `revision` almost equal to `commit`. So for example, `revision.id` sounds more natural to me than `revision.commit`.
WDYT?

brigade-worker/src/events.ts
```diff
@@ -70,6 +65,25 @@ export class BrigadeEvent {
...
+  /*
+   * Ref is the symbolic ref name. (refs/heads/master, refs/pull/12/head, refs/tags/v0.1.0)
```
mumoshu: The value examples in the comment really helps. Thanks 👍

brigade-worker/src/k8s.ts
```diff
@@ -566,14 +556,26 @@ function sidecarSpec(
...
+
   return spec;
 }
 
 function newRunnerPod(
   podname: string,
   brigadeImage: string,
   imageForcePull: boolean,
-  serviceAccount: string,
+  serviceAccount: string
```
mumoshu: Very nit-picky but I prefer the trailing comma. So that we won't necessary introduce a bit noisy diff due to just adding a new parameter to this func.

```
function newRunnerPod(
    podname: string,
    brigadeImage: string,
    imageForcePull: boolean,
 -  serviceAccount: string
 +  serviceAccount: string,
 +  aBrandNewParam: string
```

brigade-worker/src/index.ts
```diff
@@ -8,7 +8,8 @@
  * - `BRIGADE_EVENT_PROVIDER`: The name of the event provider, such as `github` or `dockerhub`
  * - `BRIGADE_PROJECT_ID`: The project ID. This is used to load the Project
  *   object from configuration.
- * - `BRIGADE_COMMIT`: The VCS commit ID (e.g. the Git commit)
+ * - `BRIGADE_COMMIT_ID`: The VCS commit ID (e.g. the Git commit)
+ * - `BRIGADE_COMMIT_REF`: The VCS full reference
```
mumoshu: Adding something like "This defaults to `refs/head/master`" would be even more helpful imho.
So that we can do understand that the value for BRIGADE_COMMIT_ID can be unset but BRIGADE_COMMIT_REF.

docs/topics/developers.md
```diff
@@ -164,7 +164,7 @@ Here is an example script for running a quick test against a locally running bri
 
 export BRIGADE_EVENT_TYPE=quicktest
 export BRIGADE_EVENT_PROVIDER=script
-export BRIGADE_COMMIT=9c75584920f1297008118915024927cc099d5dcc
+export BRIGADE_COMMIT_ID=9c75584920f1297008118915024927cc099d5dcc
```
mumoshu: Shouldn't we include `BRIGADE_COMMIT_REF` rather than `~ID` because it is `REF` which value always exists?

pkg/storage/kube/build.go
```diff
@@ -65,11 +59,13 @@ func (s *store) CreateBuild(build *brigade.Build) error {
 			"payload": build.Payload,
 		},
 		StringData: map[string]string{
-			"project_id":     build.ProjectID,
-			"event_type":     build.Type,
-			"event_provider": build.Provider,
-			"commit":         build.Commit,
+			"build_id":       buildName,
 			"build_name":     buildName,
+			"commit_id":      build.Revision.Commit,
```
mumoshu: I understand this, but on the other hand this make me think that we'd better name it `Commit` rather than `Revision`.

More concretely, I think if it looked like:

```
"commit_id": build.Commit.ID,
"commit_ref": build.Commit.Ref,
```

That's easier to read for me due to consistency(with env var naming. This is already mirrored to BRIGADE_COMMIT_ID and BRIGADE_COMMIT_REF anyways?) and less indirection hence less translation in my mental model of `Revision` to `Commit`.

docs/topics/developers.md
mumoshu: And you already seem to be aware of that in [selftest.sh](https://github.com/Azure/brigade/pull/307/files#diff-14d431b5ed62d9f4ec854c1e155124a9R16)!

.gitignore
```diff
@@ -1,13 +1,11 @@
...
+/brigade-api/rootfs/brigade-api
+/brigade-controller/rootfs/brigade-controller
+/brigade-cr-gateway/rootfs/brigade-cr-gateway
+/brigade-github-gateway/rootfs/brigade-github-gateway
+/brigade-vacuum/rootfs/brigade-vacuum
+/brigade-worker/dist/
+/brigade-worker/doc/
+/tests/testdata/*-generated.*
+/vendor/
 node_modules/
```
mumoshu: Nit: Perhaps you'd also want to chage this like `/node_modules/`?

brig/cmd/brig/commands/run.go
```diff
@@ -120,20 +120,18 @@ func (a *scriptRunner) send(projectName string, data []byte) error {
...
 	if err := a.store.CreateBuild(b); err != nil {
 		return err
 	}
 
-	shortRef := b.Commit
-	if len(shortRef) > 8 {
-		shortRef = shortRef[0:8]
-	}
-	podName := fmt.Sprintf("brigade-worker-%s-%s", b.ID, shortRef)
+	podName := fmt.Sprintf("brigade-worker-%s", b.ID)
```
mumoshu: Sounds OK becuase I have never relied on the trailing shortRef in the podName.

Maybe It wasn't necessary at all?

.gitignore
technosophos: I think that without the slash, it matches any `node_modules/` directory, while with the slash it just matches a directory at the root.

brig/cmd/brig/commands/run.go
technosophos: The shortRef was causing problems in the cases where it was not a valid Kubernetes DNS string. And since a ULID is sortable by time, this made the whole thing more compact. So I like this change because it fixes a bug while also making our use of ULID a little more uniform.

brigade-worker/src/events.ts
technosophos: That's a good point. I wonder what the Git documentation calls these.

brigade-worker/src/k8s.ts
adamreese: This is a result of running `yarn format`

pkg/storage/kube/build.go
adamreese: I have named this revision to keep it generic in case we add more fields (`base_ref`, `base_sha`, `pull_request`, `branch`, `head_remote_url`).  Also a build could be multiple commits.

.gitignore
adamreese: yep, without the `/` it is recursive.

brigade-worker/src/events.ts
mumoshu: Actually, it is called just "SHA-1" throughtout man-pages and the textbook at git-scm.com as far as I know.

Example: https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection

Based on that and assuming a "revision" has up to 1 git-commit in it, I'm also ok with `revision.sha1` rather than `revision.commit`.

Btw, `revision.ref` makes sense to me, because it refers to a git-ref, which I believe that it is already used widely/officially in git.

WDYT?

brigade-worker/src/k8s.ts
mumoshu: Got it. LGTM then. Thanks for the explanation 👍

pkg/storage/kube/build.go
mumoshu: Thanks for the explanation!

I understood that you meant that a "revision"(and yeah, I got to know that git-scm.com does mention about the word "revision" many times) in Git can be represented by many ways like you've noted, right?

That makes sense to me.
However then, I'd prefer `build.Revision.SHA1` or maybe `build.Revision.SHA` to `build.Revision.Commit`.
If you name it `Commit`, I doubt if it conflicts with `base_sha`, which also refers to a specific "commit" by a sha1 hash.

.gitignore
mumoshu: I was too blind! Totally understood. Thank you very much for the explanation.

pkg/storage/kube/build.go
mumoshu: > Also a build could be multiple commits.

This is certainly what I had missed in my mind. Thanks 👍 

Probably we'd better have `revisions[].commit` rather than `revison.commits` in that case, though?

At least in git's terminology, a "revision" seems like representing a single state in the history, but not a "range between two states". If that's the case, IMHO, `revision.commits` sounds a bit unnatural.

git-sidecar/rootfs/clone.sh
```diff
@@ -1,23 +1,35 @@
...
-: "${VCS_REVISION:?}"
-: "${INIT_GIT_SUBMODULES:?}"
+# The Git reference. Required.
+#
+# Also can be a full hex object name.
+# master, v0.1.0, refs/pulls/1/head, 589e15029e1e44dee48de4800daf1f78e64287c0
+: "${BRIGADE_COMMIT_REF:?}"
+
+# The Git SHA1 of the revision.
+: "${BRIGADE_COMMIT_ID:=}"
```
mumoshu: After spending some time on reviewing this PR, I realized the clear distinctions of "commit id" vs "commit ref" vs "revision" and concise comments on these.

Even though I still have some nits/confusion between "commit" and "revision", merging this PR will indeed be a significant step forward to make brigade a more general solution as a CI system.

Thank you very much for your efforts 👍

PR #309, opened 2018-02-13
fix(controller, worker): remove extra tsc

### LINKED ISSUE #303: Remove extra `tsc`
Each time the worker is started up, `tsc` is run:

```
Started build 01c5wf0xe78g3b49a7w751xjsq as "brigade-worker-01c5wf0xe78g3b49a7w751xjsq-master"
yarn run v1.3.2
$ node prestart.js
prestart: src/brigade.js written
$ yarn build && node --no-deprecation ./dist/index.js
$ tsc
```

I removed this once before, but it seems to have crept back in. And it incurs as much as a 30 second penalty on my system.

technosophos:
But also allow the worker command to be customized so that users can
do things like increase verbosity or choose to have tsc run again.

Closes #303


technosophos:
Oh, I also added the `-s` flag by default, which makes the output of yarn less verbose.

charts/brigade-project/values.yaml
```diff
@@ -92,3 +92,12 @@ kubernetes:
...
   # point to a readWriteMany volume provisioner.
   #buildStorageClass: default
+
+###
+# ADAVANCED
+###
+
+# OPTIONAL: specify an alternative command that the worker pod should run on startup.
+# For example, if you want the typescript compiler to re-run, replace this with
+# 'yarn build-start'
```
mumoshu: Just my two cents but also mentioning the two points below would be even more helpful:

- "The default is `yarn -s start` which suppresses log messages but the script output"
- "For extra log verbosity, just uncomment the line below"

brigade-controller/cmd/brigade-controller/controller/controller_test.go
```diff
@@ -284,3 +291,81 @@ func TestController_NoSidecar(t *testing.T) {
...
+	client := fake.NewSimpleClientset()
+	client.PrependReactor("create", "pods", func(action core.Action) (bool, runtime.Object, error) {
+		createdPod = true
+		t.Log("creating pod")
+		return false, nil, nil
+	})
+
+	config := &Config{
+		Namespace:        v1.NamespaceDefault,
+		WorkerImage:      "deis/brgiade-worker:latest",
```
mumoshu: Nit: `brgiade` -> `brigade`

technosophos:
Fixed those few nits. Will merge tomorrow (for me). Thanks!

PR #310, opened 2018-02-13
fix(github): support using GitHub Enterprise

### LINKED ISSUE #241: Github integration always defaults to github.com ?
Greetings,

I try to connect brigade to a private github instance. (not github.com but github.my-company.corp). In my project definition, I have my private github instance showing up in the clone url and the repository name. However, when I test the setup, I always see:

```
[GIN] 2017/12/28 - 08:22:27 | 200 |    5.873496ms |      10.132.0.4 |  POST     /events/github
Creating Build failed: GET https://api.github.com/repos/D057760/ubiquitous-journey/contents/?ref=522f56e37609fc70975137de8144da7c16b61b4e: 401 Bad credentials []
Error setting status to failure: POST https://api.github.com/repos/D057760/ubiquitous-journey/statuses/522f56e37609fc70975137de8144da7c16b61b4e: 401 Bad credentials []
```

The problem being that "api.github.com" is used.That is the default form the go github api client. Instead I would expect to see something pointing to my private github instance.

Looking at the code in [client.go](/Azure/brigade/blob/master/pkg/webhook/client.go), I do not see any way to override this.

technosophos:
This adds support for specifying the API endpoint URL for GitHub
Enterprise. If none is set, it defaults to regular GitHub API.

Closes #241

### DIFF
--- a/charts/brigade-project/values.yaml
+++ b/charts/brigade-project/values.yaml
@@ -27,6 +27,9 @@ sharedSecret: "IBrakeForSeaBeasts"
 # This is REQUIRED for the GitHub gateway, but optional otherwise.
 github:
 #   token: "github oauth token"
+#   ENTERPRISE: For enterprise GitHub customers, set both of these as well.
+#   baseURL: "https://internal.github.url/foo"
+#   uploadURL: "https://internal.github.url/foo"
 
 # OPTIONAL: Use this for private repositories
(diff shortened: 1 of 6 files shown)


technosophos:
@holgpar I don't have a GHE that I can test this on. But I think this should now work to set GitHub Enterprise URLs.

PR #312, opened 2018-02-13
fix(worker): die on non-existent job image

### LINKED ISSUE #193: Container Image Verify
Running a Job with a container image that doesn't exist errors with a non-useful message:
`FATAL: TypeError: First argument must be a string, Buffer, ArrayBuffer, Array, or array-like object. (rejection)`

How can we verify the image exists or improve error messaging before proceeding?

technosophos:
This causes the worker to fail if a Job's image cannot be fetched by
Kubernetes.

Closes #193

### DIFF
--- a/brigade-worker/src/k8s.ts
+++ b/brigade-worker/src/k8s.ts
@@ -458,6 +458,15 @@ export class JobRunner implements jobs.JobRunner {
             } else if (phase == "Failed") {
               clearTimers();
               reject(new Error(`Pod ${name} failed to run to completion`));
+            } else if (phase == "Pending") {
+              // Trap image pull errors and consider them fatal.
+              let cs = pod.status.containerStatuses;
+              if (cs && cs.length > 0 && cs[0].state.waiting && cs[0].state.waiting.reason == "ErrImagePull") {
+                k.deleteNamespacedPod(name, ns, new kubernetes.V1DeleteOptions())
+                  .catch(e => logger.error(e))
+                clearTimers()
+                reject(new Error(cs[0].state.waiting.message));
+              }
             }
             logger.log(
               `${pod.metadata.namespace}/${pod.metadata.name} phase ${
@@ -479,7 +488,7 @@ export class JobRunner implements jobs.JobRunner {
           clearTimeout(waiter);
           return;
         }
-        pollOnce(name, ns, interval);
+        pollOnce(name, ns, interval)
       }, 2000);
       let clearTimers = () => {
         clearInterval(interval);


brigade-worker/src/k8s.ts
```diff
@@ -458,6 +458,15 @@ export class JobRunner implements jobs.JobRunner {
...
               clearTimers();
               reject(new Error(`Pod ${name} failed to run to completion`));
+            } else if (phase == "Pending") {
+              // Trap image pull errors and consider them fatal.
+              let cs = pod.status.containerStatuses;
+              if (cs && cs.length > 0 && cs[0].state.waiting && cs[0].state.waiting.reason == "ErrImagePull") {
+                k.deleteNamespacedPod(name, ns, new kubernetes.V1DeleteOptions())
+                  .then(() => console.log("deleted " + name))
+                  .catch(e => console.log(e))
+                reject(new Error(cs[0].state.waiting.message));
```
mumoshu: Honestly I'm not a modern JS writer.
But I believe this should be `.finally(() => reject(new Error(cs[0].state.waiting.message)))`, so that we won't end up rejecting before writing useful logs regarding the pod deletion?

technosophos: My intention here was to delete in the background and reject right away so that the script could continue running. I'm going to test it with the `finally` and see how much of a performance hit that is.

technosophos: First, after taking a second look, I realized I'd missed a call to clearTimers()! Glad you had me take a second look at this section.

Second, if I move this to a `finally`, there's a weird race condition. A race condition can occur if the deletion is slow:

- If deleteNamespacedPod() takes too long (slow k8s API server), the `setInterval` will trigger
- If the interval triggers again, the container _may_ still be in ErrImagePull (or it may toggle to ImagePullBackoff)
  - If it's in ErrImagePull, it will start a second delete operation, and then send a second rejection

It doesn't look like anything really bad happens as a result. But I feel a little more comfortable leaving the rejection outside of a `finally()`. In that case, it will reject on the current tick and the interval will not be triggered again.

Here's the script I tested with:

```javascript
const {events, Job} = require("brigadier");

events.on("exec", () => {
  const job = new Job("fake", "nosuchimagebanana:latest")
  job.run().catch(() => {
    const recover = new Job("recover", "alpine:3.6", ["echo recovering"])
    recover.run()
  })
})

events.on("fail", () => {
  (new Job("fake", "nosuchimagebanana:latest")).run()
});
```

mumoshu: > I realized I'd missed a call to clearTimers()! Glad you had me take a second look at this section.

Glad it helped anyway! 😆 

> But I feel a little more comfortable leaving the rejection outside of a finally(). In that case, it will reject on the current tick and the interval will not be triggered again.

Understood & LGTM. Thank you very much for taking your time 👍 

---

Btw, just wanted to make note to myself but - would the race disappears once we move `clearTimers()` in front of `deleteNamespacedPod` in theory?

More concretely, I mean:

```
clearTimers()
k.deleteNamespacedPod(name, ns, new kubernetes.V1DeleteOptions())
  .then(() => console.log("deleted " + name))
  .catch(e => console.log(e))
  .finally(() => reject(new Error(cs[0].state.waiting.message)))
```

Really no need to fix anything(even if it did fix the race).

I just wanted to understand what we are actually doing, so that I won't make considerable mistake(s) in the future.

And anyway, we'd better stick with your current approach, as it is much simpler.
We don't seem to gain much benefit by tackling the race here. It is just about ordering of log messages 😉

technosophos: Yeah, that would work too, I think.

PR #313, opened 2018-02-14
docs(install,developer): update installation info

technosophos:
This updates the installation docs to (a) make the install docs more
reliable, and (b) make the developer docs clearer.

### DIFF
--- a/docs/topics/developers.md
+++ b/docs/topics/developers.md
@@ -83,13 +83,21 @@ Docker daemon:
 $ eval $(minikube docker-env)
 ```
 
-Running `make docker-build docker-push` will push the Brigade images to the Minikube Docker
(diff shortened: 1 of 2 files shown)


docs/topics/developers.md
```diff
@@ -83,13 +83,21 @@ Docker daemon:
 $ eval $(minikube docker-env)
 ```
 
-Running `make docker-build docker-push` will push the Brigade images to the Minikube Docker
+Running `VERSION=latest make docker-build docker-push` will push the Brigade images to the Minikube Docker
```
mumoshu: Ah, nice. I had been blindly building brigade components without the `VERSION=latest` and setting `controller.tag`, `api.tag`, `worker.tag` and so on while installing the brigade 😅

PR #317, opened 2018-02-14
fix(chart): fix broken for loop in brigade-project chart.

technosophos:
Fix a syntax bug introduced in a recent change.


charts/brigade-project/templates/secret.yaml
```diff
@@ -44,7 +44,7 @@ data:
   allowHostMounts: {{ b64enc "false" }}
   {{- end }}
   {{ if .Values.worker -}}
```
adamreese: Can this `if` statement be removed?

technosophos: Yup

PR #328, opened 2018-02-25
charts/gateway: use enable flag on all resources

pawelprazak:
- cr.enable was being ignored in the service and role
- gw.enable was being ignored in the role


pawelprazak:
Yes, it looks like someone has done it the second time.

technosophos:
Merging this to get the improved comments.

PR #329, opened 2018-02-25
chore(Makefile): remove unused dependencies from bootstrap

adamreese:
- remove unused dependencies from bootstrap
- automagically vendor when missing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,13 +22,13 @@ CX_ARCHS = amd64
 build: $(BINS)
 
 .PHONY: $(BINS)
-$(BINS):
+$(BINS): vendor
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -136,11 +136,9 @@ format:
 	test -z "$$(find . -path ./vendor -prune -type f -o -name '*.go' -exec gofmt -w {} + | tee /dev/stderr)"
 
 HAS_NPM          := $(shell command -v npm;)
```
adamreese: Can we remove npm from here also?

pawelprazak:
btw I've filed a partially related issue #330

Makefile
mumoshu: Seems so to me, and the mention of npm in developers.md, too!

PR #333, opened 2018-02-26
Add brig proxy command to start tunnels to api and kashti

radu-matei:
WIP implementation of `brig proxy` that currently starts tunnels to the API and Kashti.

This way, you don't have to expose the dashboard publicly, and connecting is done through a single command.

### DIFF
--- a/brig/cmd/brig/commands/proxy.go
+++ b/brig/cmd/brig/commands/proxy.go
@@ -1,0 +1,77 @@
+package commands
+
+import (
+	"fmt"
+	"os"
+	"os/signal"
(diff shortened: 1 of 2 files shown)


technosophos:
This is a great idea, and a very straightforward implementation. Whenever you're done, take of the WIP and I'll LGTM and merge!

radu-matei:
Also added `--api-port` flag for the cases when Kashti was deployed with a different port for the API server (`helm install -n kashti ./charts/kashti --set brigade.apiServer=http://localhost:<other-port>`)

PR #337, opened 2018-02-27
Only create service & RBAC for components when enabled

### LINKED ISSUE #336: Brigade Chart: container registry service created when disabled
Very simple one, in the helm chart for Brigade, the container registry service is  created even when it is disabled for a release.

TimJones:
Fixes #336 

**NOTE** I noticed that RBAC resources are created for each service  too, even when the service is disabled as well. I could add a check for that to this PR quickly if needed.

### DIFF
--- a/charts/brigade/templates/api-role.yaml
+++ b/charts/brigade/templates/api-role.yaml
@@ -1,3 +1,4 @@
+{{ if .Values.api.enabled }}
 {{ $fname := include "brigade.api.fullname" . }}
 ---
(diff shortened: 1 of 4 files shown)


adamreese:
Thanks @TimJones.  It would be great to add this to the RBAC resources too.

TimJones:
@adamreese Updated to only create ServiceAccount\Roles\RoleBindings for each component only when it is enabled.

adamreese:
Thanks!

PR #339, opened 2018-03-01
ref(*): remove label for commit_id

adamreese:
This is unused and can be confusing.


adamreese:
as discussed with @technosophos

PR #341, opened 2018-03-02
docs(topics): Add document on customizing workers

technosophos:
This adds documentation explaining how to write and use custom worker
images with custom JavaScript libraries.

### DIFF
--- a/docs/topics/workers.md
+++ b/docs/topics/workers.md
@@ -1,0 +1,224 @@
+# Building Custom Workers
+
+A worker is the component in Brigade that executes a `brigade.js`. Brigade ships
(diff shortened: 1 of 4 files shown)


docs/topics/workers.md
```diff
@@ -0,0 +1,224 @@
...
+
+```
+
+Now we've added a few new lines to the script. We import `alpineJob` from our
+`./mylib` module at the top. Then, late in the script, we call `alpineJob` to
+create a new job for us.
+
+## Best Practices
+
+As we have seen, it is easy to add new funcitonality to the Brigade worker. But
```
adamreese: s/funcitonality/functionality/

PR #344, opened 2018-03-05
docs(install): document using NFS for volume storage

technosophos:
This documents how to use nfs-provisioner to back storage and cache to
an NFS instance.

### DIFF
--- a/docs/topics/install.md
+++ b/docs/topics/install.md
@@ -87,8 +87,111 @@ Brigade creates Persistent Volume Claims on the fly. By default it will create
 PVCs using the default persistent volume class. Changing your cluster's default
 storage class will change what Brigade creates.
 
-We recommend the following file system types for these distributions or platforms:
+The following volume backends have been tested:
 
 - Azure: AzureFile
 - Minikube: 9p (the default)
-- General: A storage type that supports ReadWriteMany.
+- Any: NFS
+
+The NFS solution is explained below.
+
+#### Using an NFS Server
+
+Brigade uses storage for caching and short-term file sharing. Because of this,
+it is often convenient to use storage backends that are optimized for short-term
+ephemeral storage.
+
+NFS (Network File System) is one protocol that works well for Brigade. You can
+use the [NFS Provisioner](https://github.com/IlyaSemenov/nfs-provisioner-chart)
+chart to easily install an NFS server.
+
+```console
+$ helm repo add nfs-provisioner https://raw.githubusercontent.com/IlyaSemenov/nfs-provisioner-chart/master/repo
+$ helm install --name nfs nfs-provisioner/nfs-provisioner --set hostPath=/var/run/nfs-provisioner
+```
+
+(Note that RBAC is enabled by default. To turn it off, use `--set rbac.enabled=false`.)
+
+To use an emptyDir instead of a host mount, set `--hostPath=""`.
+
+
+```console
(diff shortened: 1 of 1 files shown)


docs/topics/install.md
```diff
@@ -87,8 +87,100 @@ Brigade creates Persistent Volume Claims on the fly. By default it will create
...
+I0305 21:25:50.994219       1 controller.go:803] volume "pvc-c746f068-20bb-11e8-a31a-080027a443a9" for claim "default/github-com-deis-empty-testbed-three" saved
+I0305 21:25:50.994237       1 controller.go:839] volume "pvc-c746f068-20bb-11e8-a31a-080027a443a9" provisioned for claim "default/github-com-deis-empty-testbed-three"
+I0305 21:25:56.974297       1 controller.go:1028] volume "pvc-c13e95f0-20bb-11e8-a31a-080027a443a9" deleted
+I0305 21:25:56.985432       1 controller.go:1039] volume "pvc-c13e95f0-20bb-11e8-a31a-080027a443a9" deleted from database
+```
+
+Implementation details of note:
+
+- The NFS server used is [NFS-Ganesha](https://github.com/nfs-ganesha/nfs-ganesha)
+- The Kubernetes proisioner is part of [kubernetes-incubator/external-storage](https://github.com/kubernetes-incubator/external-storage/tree/master/nfs)
```
adamreese: provisioner

docs/topics/install.md
```diff
@@ -87,8 +87,100 @@ Brigade creates Persistent Volume Claims on the fly. By default it will create
...
+#### Using an NFS Server
+
+Brigade uses storage for caching and short-term file sharing. Because of this,
+it is often convenient to use storage backends that are optimized for short-term
+ephemeral storage.
+
+NFS (Network File System) is one protocol that works well for Brigade. You can
+use the [NFS Provisioner](https://github.com/IlyaSemenov/nfs-provisioner-chart)
+chart to easily install an NFS server.
+
```
lukepatrick: Note, one assumption is the host has nfs-common. Rare not to have it but we hit that once before.
If not already set up on your host, add nfs-common
`$ apt-get install nfs-common`

docs/topics/install.md
```diff
@@ -87,8 +87,100 @@ Brigade creates Persistent Volume Claims on the fly. By default it will create
...
+If you have plenty of memory to spare, and are more concerned with fast storage,
+you can configure the provisioner to use a tmpfs in-memory filesystem.
+
+```console
+$ helm install --name nfs nfs-provisioner/nfs-provisioner --set hostPath="" --set useTmpfs=true
+```
+
+This chart installs a `StorageClass` named `local-nfs`. Brigade projects can
+each declare which storage classes they want to use. And there are two storage
+class settings:
```
lukepatrick: In lieu of a named StorageClass, this can be installed as a default storage class. Though an explicit name is recommended.
`$ helm install --name nfs --namespace nfs-provisioner nfs-provisioner/nfs-provisioner --set defaultClass=true`

docs/topics/install.md
technosophos: That is _very_ good to know. Thanks!

docs/topics/install.md
technosophos: That's a good idea.

docs/topics/install.md
technosophos: Good catch. Fixing.

PR #348, opened 2018-03-06
fix(worker) use sshKey for private repo, not brigadeSSHKey.

lukepatrick:
fix(worker) use sshKey for private repo. There is no brigadeSSHKey in the Secret data.

### DIFF
--- a/brigade-worker/src/k8s.ts
+++ b/brigade-worker/src/k8s.ts
@@ -567,7 +567,7 @@ function sidecarSpec(
       name: "BRIGADE_REPO_KEY",
       valueFrom: {
         secretKeyRef: {
-          key: "brigadeSSHKey",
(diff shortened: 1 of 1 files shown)


brigade-worker/src/k8s.ts
```diff
@@ -567,7 +567,7 @@ function sidecarSpec(
       name: "BRIGADE_REPO_KEY",
       valueFrom: {
         secretKeyRef: {
-          key: "brigadeSSHKey",
+          key: "sshKey",
```
mumoshu: Double-checked that this matches [what we expect the key to be in another part of `k8s.ts`](https://github.com/Azure/brigade/pull/348/files#diff-d54f19b6bbea870b89fa594044af83d9R728).

Issue #342, opened 2018-03-02
vcsSidecar cleanup / pull policy


lukepatrick:
Recently the default vcsSidecar `deis/git-sidecar` was updated. The image pull policy for the sidecar uses the **worker** policy, default `IfNotPresent`.

If you had the git-sidecar in your local docker registry, a new install of Brigade would not pull the latest image. In the case of going from `v0.9.0` to `v0.10.0` you would see the error:
```
+ : /src
/clone.sh: line 6: VCS_REPO: parameter not set or null
```

Some ideas:

- A clean up or intentional pull of the sidecar image at delete/install of Brigade
- Default the brigade-project _values.yaml_ `vcsSidecar: "deis/git-sidecar:latest"` with a tag name (v0.10.0)
- Add a pullPolicy property to brigade-project _values.yaml_ for the sidecar. Not ideal to pull every time a build happens...

technosophos:
Oh! Ouch! You're right! We were supposed to set that to the latest image tag. I've been meaning to fix that for ages.

technosophos:
Fixed via #349

PR #349, opened 2018-03-07
fix(chart): BREAKING set vcsSidecar tag to version

technosophos:
Previously, this always used `:latest` for the vcs sidecar tag. Now it
uses .Chart.Version. In order to not use any vcsSidecar, you now have
to set the version to NONE, which is breaking, but probably an unused
feature anyway.


lukepatrick:
fixes #342

PR #358, opened 2018-03-12
ref(worker): move env var processing to the top

technosophos:
This is a minor refactor to move an env var reference to the top.

### DIFF
--- a/brigade-worker/src/k8s.ts
+++ b/brigade-worker/src/k8s.ts
@@ -21,9 +21,33 @@ const expiresInMSec = 1000 * 60 * 60 * 24 * 30;
 
 const defaultClient = kubernetes.Config.defaultClient();
 
-const defaultServiceAccount = "brigade-worker";
-
 const logger = new ContextLogger("k8s");
+
+/**
+ * options is the set of configuration options for the library.
+ *
+ * The k8s library provides a backend for the brigade.js objects. But it needs
+ * some configuration that is to be passed directly to the library, not via the
+ * brigade.js. To allow for this plus overrides (e.g. by Project or Job objects),
+ * we maintain a top-level singleton object that holds configuration.
+ *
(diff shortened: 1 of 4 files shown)


technosophos:
I'm trying to figure out how to best bubble this env var reference up to the `index.js` file. This feels pretty hacky, so if you have any clever ideas, I'd be happy to make this better.

/cc @jchanam and @adamreese

jchanam:
I think this is a much better approach. The only thing that I don't know is: Why to have the default value defined twice?

This is what I tried to do on my first approach, but couldn't achieve it :(

mumoshu:
@technosophos Just trying to figure out your intension -
 maybe you're going to update this line: https://github.com/Azure/brigade/pull/358/files#diff-d54f19b6bbea870b89fa594044af83d9R158

from

```
this.serviceAccount = job.serviceAccount || process.env.BRIGADE_SERVICE_ACCOUNT || defaultServiceAccount;
```

to

```
this.serviceAccount = job.serviceAccount || defaultServiceAccount;
```

on the other hand of your current change?

technosophos:
Ugh... that was my mistake, @mumoshu. I need to fix that.

technosophos:
Okay, refactored this based on some OTS input from @adamreese . This  new version should be a little bit more future-proof, and is based on an idea he found in `ts-node`.

brigade-worker/src/k8s.ts
```diff
@@ -142,18 +166,18 @@ export class JobRunner implements jobs.JobRunner {
   event: BrigadeEvent;
   job: jobs.Job;
   client: kubernetes.Core_v1Api;
+  localOptions: KubernetesOptions;
```
adamreese: i would just call this `options`

technosophos: Fixed

Issue #368, opened 2018-03-14
brigade / brigade-project version incompatibilities are somewhat hidden


blimmer:
I'm testing out a brigade setup in minikube. I've got a private repo set up that I want to test with and I'm `curl`-ing the cr gateway to trigger the build. However, I'm met with an instant failure and no other information as to what's wrong:

```
prestart: empty script found. Falling back to VCS script
prestart: no script override
error Command failed with exit code 1.
```

I'm not seeing the VCS sidecar come up, so maybe there's a problem there?

I also can't run `brig run` even with the verbose flag, it just hangs with no output

```
brig run --verbose -f brigade.js blimmer/crawler
# ... nothing
```

Is there additional verbosity I can pass through to the prestart or something? I'm just not sure how to further debug the issue.

lukepatrick:
Can you get the vcs sidecar logs?
```$ kubectl logs brigade-worker-1234abcd vcs-sidecar```
or output of describe on the worker?
```$ kubectl decribe pod brigade-worker-1234abcd```

**Comment by technosophos** [comment]
If `brig run` hangs with no output, it usually means that it had trouble with something in Kubernetes. It could be that it's not pointed at the cluster (but in that case, `brig project list` will also fail). It could also be that the event that fired was never received by the controller. Usually there is at least some output in that case, but it _could_ happen if there is something wrong when trying to read the `blimmer/crawler` project.

The best way to get debug data for that is by looking at the logs of the controller. I think something like this should work:

```
$ kubectl logs $(kubectl get pods | grep brigade-ctrl | awk '{ print $1 }')
```

That might tell you why the event secret was not processed.

**Comment by blimmer** [comment]
I think the problem is that the vcs-sidecar isn't being instantiated at all - the command `brigade-worker-<hash> vcs-sidecar` reports that it's invalid.

```
kubectl describe pod brigade-worker-01c8kbnfqmvfmpne28bmg0y48k
Name:         brigade-worker-01c8kbnfqmvfmpne28bmg0y48k
Namespace:    default
Node:         minikube/192.168.99.100
Start Time:   Wed, 14 Mar 2018 16:56:35 -0600
Labels:       build=01c8kbnfqmvfmpne28bmg0y48k
              component=build
              heritage=brigade
              project=brigade-2f0fb8be11807675c14b07d6d9980e43b94a02e890b35ddd02eb4f
Annotations:  <none>
Status:       Failed
IP:           172.17.0.2
Containers:
  brigade-runner:
[...]

PR #375, opened 2018-03-15
fix(brig): add verbose output for waiting

technosophos:
One of the common Brig problems occurs when Brig creates an event, but
the controller does not start the worker. This adds a single line of
verbose output indicating (a) when Brig starts waiting for the worker,
and (b) what pod it expects to start as a worker.

Partially addresses #368


brig/cmd/brig/commands/run.go
```diff
@@ -144,6 +144,10 @@ func (a *scriptRunner) send(projectName string, data []byte) error {
 
 	podName := fmt.Sprintf("brigade-worker-%s", b.ID)
 
+	if globalVerbose {
```
adamreese: Why is this behind a verbose flag?

PR #393, opened 2018-03-22
feat(brig): add 'brig build logs'

technosophos:
This gives Brig a new `build logs` subcommand for dumping logs.

Along the way, I backfilled a bunch of tests, too.

### DIFF
--- a/brig/cmd/brig/commands/build_logs.go
+++ b/brig/cmd/brig/commands/build_logs.go
@@ -1,0 +1,121 @@
+package commands
+
+import (
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/spf13/cobra"
+
+	"github.com/Azure/brigade/pkg/brigade"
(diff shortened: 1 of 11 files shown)


brig/cmd/brig/commands/build_logs.go
```diff
@@ -0,0 +1,115 @@
...
+
+Print the logs for a build, and (optionally) the jobs executed as part of the
+build.
+`
+)
+
+var logsAll bool
+
+func init() {
+	buildLogs.Flags().BoolVarP(&logsAll, "all", "a", false, "Show job logs as well as the worker log")
```
adamreese: This flag name should probably be `--jobs`.

`--all` seems like I should get logs for all workers `brig build logs --all`

technosophos: Fixed

technosophos:
@adamreese Fixed

PR #395, opened 2018-03-23
Fix: ACI connector makes it work

TsuyoshiUshio:
Currently, ACI connector doesn't work for the Brigade. I create a pull request for fixing this.

Also, when I tried locally, I found one issue. 

```
     message: 'Secret "one-01C99SZ7D3E9B8W60EM6K731ZT" is invalid: metadata.name: Invalid value: "one-01C99SZ7D3E9B8W60EM6K731ZT": a DNS-1123 subdomain must consist of lower case alphanumeric characters, \'-\' or \'.\', and must start and end with an alphanumeric character (e.g. \'example.com\', regex used for validation is \'[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*\')',
```
The first commit fix this issue for local execution. 

The second commit enable us to specify request cpu and memory. If we don't specify these two, ACI connector doesn't work now. 

After this change, We can write `brigade.js` like this

```
const { events, Job } = require('brigadier')

events.on("exec", (brigadeEvent, project) => {
  console.log("******** start working **********")
  var one = new Job("one")
  one.image = "alpine:3.4"
  one.tasks = ["echo hello aci"]
  one.host.os = "linux"
  one.host.name = "virtual-kubelet"
  one.requests.cpu = "1"
  one.requests.memory = "1G"
  one.run()
  console.log("******** end container **********")
})
```

### DIFF
--- a/brigade-worker/src/index.ts
+++ b/brigade-worker/src/index.ts
@@ -50,7 +50,7 @@ const requiredEnvVar = (name: string): string => {
 
 const projectID: string = requiredEnvVar("BRIGADE_PROJECT_ID");
 const projectNamespace: string = requiredEnvVar("BRIGADE_PROJECT_NAMESPACE");
-const defaultULID = ulid();
+const defaultULID = ulid().toLocaleLowerCase();
 let e: events.BrigadeEvent = {
   buildID: process.env.BRIGADE_BUILD_ID || defaultULID,
   workerID: process.env.BRIGADE_BUILD_NAME || `unknown-${defaultULID}`,
--- a/brigade-worker/src/job.ts
+++ b/brigade-worker/src/job.ts
@@ -126,6 +126,16 @@ export class JobDockerMount {
 }
 
 /**
+ * Request represents request of the resources
+ */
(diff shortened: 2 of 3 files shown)


TsuyoshiUshio:
Hi @technosophos ,
I change the name of the resource and variables. 

the sample brigade.js what I execute.

```
const { events, Job } = require('brigadier')

events.on("exec", (brigadeEvent, project) => {
  console.log("******** start working **********")
  var one = new Job("one")
  one.image = "alpine:3.4"
  one.tasks = ["echo hello aci"]
  one.host.os = "linux"
  one.host.name = "virtual-kubelet"
  one.resourceRequests.cpu = "1"
  one.resourceRequests.memory = "1G"
  one.run()
  console.log("******** end container **********")
})
```

It worked as I expected

```
$ az container logs --container-name brigaderun -g aci-group --name default-one-01c9r9ma005etakm6ry4cbn5c9
hello aci
```

PR #396, opened 2018-03-25
fix(worker): Allow secretKeyRef and configMapKeyRef as Job env vars

jdotjdot:
Currently, `Job`s in `brigade.js` files can only have key/value pairs in the `env` object. This means that accessing any secrets either needs to dynamically pulled from a storage system like Vault inside the container, or needs to be saved directly in the project in a `values.yaml` file that can't be committed to git due to containing secrets.  Many clusters will already contain a set of secrets for use across a number of deployments, especially if Brigade is being used for CI--many CI pipelines will reuse the same set of secrets.

This PR allows `secretKeyRef`, `configMapKeyRef`, and the other Kubernetes value references to be used in a job's `env` object.  This also comes with documentation updates and passing tests.  We are using this code right now internally and it works fine.

Note:
The one line I'm uncertain of is [this one](https://github.com/Azure/brigade/pull/396/files#diff-d54f19b6bbea870b89fa594044af83d9R214); ideally I'd like a better way to detect the type of the object that's being passed beyond just "is it a string or not," but unfortunately that pushes the boundaries of my Typescript knowledge.  I'm hoping someone on the Brigade team can advise on the best way to handle that.

### DIFF
--- a/brigade-worker/src/job.ts
+++ b/brigade-worker/src/job.ts
@@ -7,6 +7,8 @@
 
 /** */
 
+import {V1EnvVarSource} from "@kubernetes/client-node/api";
+
 /**
  * The default shell for the job.
  */
@@ -147,7 +149,7 @@ export abstract class Job {
   /** tasks is a list of tasks run inside of the shell*/
   public tasks: string[];
   /** env is the environment variables for the job*/
-  public env: { [key: string]: string };
+  public env: { [key: string]: string | V1EnvVarSource };
   /** image is the container image to be run*/
   public image: string = brigadeImage;
   /** imageForcePull defines the container image pull policy: Always if true or IfNotPresent if false */
--- a/brigade-worker/src/k8s.ts
+++ b/brigade-worker/src/k8s.ts
@@ -210,18 +210,31 @@ export class JobRunner implements jobs.JobRunner {
     let envVars: kubernetes.V1EnvVar[] = [];
     for (let key in job.env) {
       let val = job.env[key];
-      this.secret.data[key] = b64enc(val);
 
-      // Add reference to pod
-      envVars.push({
-        name: key,
-        valueFrom: {
-          secretKeyRef: {
-            name: secName,
-            key: key
+      if (typeof val === 'string') {
+        // For environmental variables that are submitted as strings,
+        // add to the job's secret and add a reference.
+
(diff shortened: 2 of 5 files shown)


technosophos:
This looks really cool! I'm going to manually test it and make sure this doesn't cause any security problems as a side effect. But this looks like it will  solve some of our use cases really nicely!

technosophos:
Okay, so the main thing we are worried about is that it appears that using this mechanism, one could load any other arbitrary secret (including other projects), which has some security implications for at least some Brigade users.

Could we add a configuration switch around this to allow the behavior to be turned off/on from `helm install brigade`? I think it'd just be adding another parameter to the worker config, and then passing that all the way through.

So, for example, I'd do `helm install brigade/brigade --set worker.allowExternalSecrets=true`, and that config would be passed all the way to the Worker on startup.

I looked at doing this at the project level, but since this is the sort of thing that will impact all projects if enabled on even one project, we should probably make it a top-level config.

jdotjdot:
While we could add a configuration switch from `helm install brigade`, I don't see how that would actually solve the security issue.  All of the additional code here is just Javascript executed by the worker container.  The worker container is already able to access those secrets, or the `secretKeyRef` wouldn't work.  As such, even if this PR weren't merged, anyone could just add a lot of Javascript in their `brigade.js` file to reproduce this same functionality, and they'd still have access to the secrets.

I'm not as much of a Kubernetes expert as you all, but I'd imagine that Kubernetes `Secrets` are a bit of a fragile place to store truly secret information.

technosophos:
Yeah you are right. I mean, any "security win" we'd get from making this configurable is basically just making it harder for a dev to use the API. As you point out, they could work around this fairly easily by just importing the k8s code and directly loading secrets. Sometimes I have to remind myself that when we give users a full scripting environment, we're implicitly giving them permission to do dangerous things.

Since the beginning, we've said that we assumed that users who were allowed to create scripts were allowed to work with Kubernetes objects to the extent that RBAC allowed them. (In other words, "don't want the scripter to see a secret? Block it with RBAC") So... if we stay true to that, then putting a switch on this particular piece of functionality does seem rather arbitrary.

Alright, I'll run this by @adamreese during our PR review today, and we'll take a more circumspect view. Thanks for the thoughtful reply.

PR #399, opened 2018-03-27
feat(brig): add progress meter to run output

technosophos:
This adds a progress indicator on 'brig run' output by collapsing repeat
entries.

Progress


adamreese:
so purdy

PR #401, opened 2018-03-27
docs(README): added a few other gateways

technosophos:
Adds:
- BitBucket gateway
- Trello gateway
- k8s gateway


radu-matei:
In what state does a gateway have to be in order to be here?

https://github.com/search?utf8=%E2%9C%93&q=brigade+gateway&type=

technosophos:
@radu-matei I'd say that if the author is comfortable with including it, or if someone else has tested it and gotten it to work, that is good enough for me.

lukepatrick:
Trying to ride in on this PR with another (not sure if effective GH etiquette..)
https://github.com/technosophos/brigade/pull/1

radu-matei:
Do you think it makes sense to also add the [Azure EventGrid Gateway](https://github.com/radu-matei/brigade-eventgrid-gateway)?

PR #409, opened 2018-04-01
feat(brig): add brig container

technosophos:
This adds the necessary Dockerfile and Makefile changes so that we can
also build a containerized Brig, which is necessary for our CronJob
gateway.


radu-matei:
I think you should also add `brig/rootfs` to `.gitignore`

technosophos:
Thanks, @radu-matei. Fixed

PR #410, opened 2018-04-02
 feat(brigade.js): build docker images on tag event.

technosophos:
This adds a new target that builds and releases to a Docker registry.

Wait to merge until after #409

### DIFF
--- a/brigade.js
+++ b/brigade.js
@@ -65,7 +65,7 @@ function build(e, project) {
   });
 }
 
-function release(e, p, tag) {
+function releaseBrig(e, p, tag) {
   if (!p.secrets.ghToken) {
     throw new Error("Project must have 'secrets.ghToken' set")
   }
@@ -105,13 +105,11 @@ function release(e, p, tag) {
   // Upload for each target that we support
   for (const f of ["linux-amd64", "windows-amd64", "darwin-amd64"]) {
     const name = binName + "-"+f
-    cx.tasks.push(`github-release upload -f ./bin/${name} -n ${name} -t ${tag}`)
+    cx.tasks.push(`github-release upload -f ./bin/${name} -n ${name} -t ${tag}`)  
   }
-
-  console.log(cx.tasks)
-  cx.run().then( res => {
-    console.log(`releases at https://github.com/${p.repo.name}/releases/tag/${tag}`);
-  })
+  console.log(cx.tasks);
+  console.log(`releases at https://github.com/${p.repo.name}/releases/tag/${tag}`);
+  return cx.run();
 }
 
 function ghNotify(state, msg, e, project) {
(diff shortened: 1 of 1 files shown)


brigade.js
```diff
@@ -127,10 +127,171 @@ function ghNotify(state, msg, e, project) {
...
+  // Docker builder
+  const dind = new Job("dind", "docker:stable-dind")
+  dind.storage.enabled = true
+  dind.privileged = true
+  dind.env = {
+    DOCKER_DRIVER: "overlay"
+  }
+  dind.tasks = [
+    "dockerd-entrypoint.sh &",
+    "echo waiting for Docker && sleep 20",
```
mumoshu: For no flakiness, `printf "waiting for docker daemon"; while ! docker info >/dev/null 2>&1; do printf .; sleep 1; done; echo` could be used instead, if you like.

technosophos: :hugs: This works great!

brigade.js
```diff
@@ -127,10 +128,169 @@ function ghNotify(state, msg, e, project) {
...
+
+  return Group.runEach([goBuild, dind]);
+}
+
+function acrBuild(e, project, tag) {
+  const gopath = "/go"
+  const localPath = gopath + "/src/github.com/" + project.repo.name;
+  const org = "deis";
+  const images = [
+    //"brig", // Uncomment this after 0.13.0.
```
adamreese: 0.13 is shipped

brigade.js
```diff
@@ -127,10 +128,169 @@ function ghNotify(state, msg, e, project) {
...
+
+  // For each image we want to build, build it, then tag it latest, then post it to registry.
+  for (let i of images) {
+    let imgName = registryHost+"/"+i+":"+tag;
+    let latest = registryHost+"/"+i+":latest";
+    dind.tasks.push(
+      `cd ${i}`,
+      `echo '========> Building ${i}'`,
+      `cp -a /mnt/brigade/share/${i}/rootfs ./rootfs`,
+      // Total hack to work around something weird with brigade-github-gateway
```
adamreese: what does this fix?

brigade.js
```diff
@@ -127,10 +128,169 @@ function ghNotify(state, msg, e, project) {
...
+      `[ ! -f /mnt/brigade/share/${i}/rootfs/${i} ] || cp /mnt/brigade/share/${i}/rootfs/${i} ./rootfs/`,
+      // TODO: Fix the Makefile to make this easier.
```
adamreese: `make docker-build docker-push IMAGE_TAG=latest`

technosophos: I'll update the comment on that. But the problem with `make docker-....` is that I always have to rebuild from source, which means always having to load the Go toolchain. Here, I am copying the binary from another location so I don't have to recompile.

brigade.js
technosophos: It fixes this peculiar case where the brigade-github-gateway bin is not copied by the `cp` above. I haven't been able to trace it down. There are no errors... it's just skipped.

brigade.js
technosophos: I'll fix

brigade.js
```diff
@@ -127,10 +128,169 @@ function ghNotify(state, msg, e, project) {
...
+  for (let i of images) {
+    let imgName = org +"/"+i+":"+tag;
+    let latest = org +"/"+i+":latest";
+    builder.tasks.push(
+      `cd ${i}`,
+      `echo '========> Building ${i}'`,
+      `cp -a /mnt/brigade/share/${i}/rootfs ./rootfs`,
+      // Total hack to work around something weird with brigade-github-gateway
+      `[ ! -f /mnt/brigade/share/${i}/rootfs/${i} ] || cp /mnt/brigade/share/${i}/rootfs/${i} ./rootfs/`,
+      // Currently, it appears this is the only way to add two tags.
```
ehotinger: 💄 You can remove this comment since it's not applicable anymore. I forgot to delete it from my previous PR.

PR #415, opened 2018-04-04
feat(brig): Exit with error code if build fails

jdotjdot:
The `brig` CLI will exit 0 even if the build fails.  Currently, it exits 0 even if there is an error thrown inside the `brigade.js` file. For our CI pipelines, this isn't desirable behavior, as we want `brig` to properly reflect that the build was unsuccessful and exit 1 so that we can fail the CI pipeline that is calling it.

This PR adds an `--exit-code` option that tells `brig` to exit with an error code when the build fails. This is done by checking the worker container's pod phase after completion of the build to see if it was successful or failed.  This relies on the fact that the worker container terminates as a failure when one of the `Jobs` within the build fails (unless caught and handled by a `.catch()`, as you'd expect).   The `--exit-code` naming convention was inspired by what Heroku includes for their `heroku run --exit-code` flag, but I'd be totally open to changing the name.

### DIFF
--- a/brig/cmd/brig/commands/rerun.go
+++ b/brig/cmd/brig/commands/rerun.go
@@ -3,7 +3,6 @@ package commands
 import (
 	"errors"
 	"io/ioutil"
-
 	"os"
 
 	"github.com/spf13/cobra"
@@ -46,10 +45,11 @@ var rerun = &cobra.Command{
 			return err
 		}
 
-		a, err := script.NewDelegatedRunner(kc, os.Stdout, globalNamespace)
+		a, err := script.NewDelegatedRunner(kc, globalNamespace)
 		if err != nil {
 			return err
 		}
+		a.ScriptLogDestination = os.Stdout
 		a.NoProgress = runNoProgress
 		a.Background = runBackground
 		a.Verbose = globalVerbose
--- a/brig/cmd/brig/commands/run.go
+++ b/brig/cmd/brig/commands/run.go
@@ -102,10 +102,11 @@ var run = &cobra.Command{
 			return err
 		}
 
-		runner, err := script.NewDelegatedRunner(kc, destination, globalNamespace)
+		runner, err := script.NewDelegatedRunner(kc, globalNamespace)
 		if err != nil {
 			return err
 		}
+		runner.ScriptLogDestination = destination
 		runner.NoProgress = runNoProgress
 		runner.Background = runBackground
 		runner.Verbose = globalVerbose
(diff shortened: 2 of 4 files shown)


technosophos:
Should we just make exiting with an error code the default?

jdotjdot:
I'd agree with that.  The one issue currently is that if it exits for other reasons than a worker failing, the behavior of Cobra's `RunE` is to print out the full CLI help, which in most cases we don't want.  However, there are some cases when you _do_ want it.

The only way I've seen to really solve this is to use `os.Exit` as the exit point, which doesn't feel great.

jdotjdot:
That said, I imagine there's a reason why Heroku has theirs not exit 0 by default and then offers an `--exit-code` switch.  Maybe we can get feedback from others on the team to see what they'd think is most natural?

technosophos:
Yeah, I can see both of those points.

The usage info is really not the approporiate output when a build fails. It would be good to be able to strategically suppress that.

If you're okay with leaving this hanging open for a short while, i think it is a good idea to ask others for some opinions.

jdotjdot:
I already have our Brigade setup running entirely off of my own set of forked containers anyway, so while not ideal, it's fine.

Just note that this PR does actually suppress the usage info for a build failure&mdash;this code here checks if the error returned is a `BrigadeFailure` and then suppresses the Cobra usage info if it is.
https://github.com/jdotjdot/brigade/blob/bd2dc9cc0bd546b7afa463502e1ea41aa76f5611/brig/cmd/brig/commands/run.go#L105

If an error happens _before_ that point (e.g., in the `ScriptRunner.send(...)` method), then currently it will still display the usage output, but that was the already existing behavior.

adamreese:
Thanks for the PR @jdotjdot! I think that this change should be the default behavior and remove the flag.  We can return a different error code for build vrs internal errors.

technosophos:
After a week of letting this sit, I agree. I think this should be default behavior, but that errors from builds should return an exit code like 2 or 3 if possible.

jdotjdot:
It's totally fine.  I can, I just have a number of other things going on right now and can't refactor immediately.  Happy for someone else to take it over, or I'll do it when I can.

technosophos:
Okay, cool.

jdotjdot:
I promise I will get this done at some point if no one else picks it up 😓

mumoshu:
This looks awesome. I'll try to make the change to move this forward.

mumoshu:
I intentionally left the dirty commits history as-is, to keep credit to @jdotjdot, and for the purpose of sharing what I've adjusted.

Once LGTMed, I'll squash before merging to keep the history clean on the master branch.

PR #417, opened 2018-04-05
fix(worker): allow jobs during error handler

### LINKED ISSUE #364: Execute termination actions on `error` lifecycle event
At our pipelines, we try to use the `after` and `error` lifecycle events in order to set Github statuses. The problem is that the `after` ones are correctly set, but the `error` ones no, Is like if this event handler doesn't wait until everything is finished (in very few cases the GitHub status job pod triggers and when this happens the github status pod finishes after the worker has finished). I guess that has something to do with this.

https://github.com/Azure/brigade/blob/a140145cd4200e9be358165978f0344c3a374a41/brigade-worker/src/app.ts#L106-L106

We are not "catching" errors when executing jobs or groups because this could make repetitive to handle in all different kind of event flows (push, pull_request, exec...). 

Are we doing something wrong?

technosophos:
The error handler was not waiting for other events to finish. This
removes the process error code handling to the "exit" event, and allows
the "error" event to schedule jobs and other things that will require a
next-tick. It also intentionally bypasses the "after" event once an
error code is set. This ensures that if "error" triggers, "after" won't.

Closes #364

### DIFF
--- a/brigade-worker/src/app.ts
+++ b/brigade-worker/src/app.ts
@@ -61,6 +61,8 @@ export class App {
    */
   public buildStorage: BuildStorage = new k8s.BuildStorage();
 
+  protected exitCode: number = 0;
+
   /**
    * Create a new App.
    *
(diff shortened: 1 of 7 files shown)


brigade-worker/src/app.ts
```diff
@@ -127,8 +133,19 @@ export class App {
       this.fireError(reason, "unhandledRejection");
     });
 
+    process.on("exit", code => {
+      if (this.exitCode!= 0) {
```
adamreese: nit formatting

PR #422, opened 2018-04-10
Link to PVC Access Mode Table

blimmer:
I didn't realize that I couldn't rely on PVC in my cluster (AWS EBS backed). This documentation change links out to the k8s support table and warns people that, unless you use NFS or are on Azure, you probably will have a bad time.


jdotjdot:
I also have this problem

technosophos:
Wow... that's bleak. I didn't notice how few readWriteMany there are.

Minikube as a 9P implementation that is also read-write-many. Guess I should also test Rook and see if that's an offering.

PR #423, opened 2018-04-11
feat(brigade.js): Retrieve logs from Job run

jdotjdot:
We needed a way to output the logs of a Job after the fact&mdash;in particular, for reporting on failed jobs.  This adds a `logs()` method to the Typescript `Job`.

Note that this does not seem to be fully working, even though the tests pass.  When I run this code:
```javascript
    let slack = new Job("slack-notify", "technosophos/slack-notify:latest", ["/slack-notify"]);
    slack.storage.enabled = false;
    slack.useSource = false;
    slack.env = {
        SLACK_WEBHOOK: project.secrets.SLACK_WEBHOOK,
        SLACK_COLOR: "red",
        SLACK_CHANNEL: "brigade-failures",
        SLACK_TITLE: `Build failure`,
        SLACK_MESSAGE: `Build ${event.buildID} failed to build.`
    };

    await slack.run()
    let logs = await slack.logs();
    console.info(`This is a test: ${logs}`);
```

I get this output:
```
Event created. Waiting for worker pod named "brigade-worker-01cav4jcjea8wzqsvs1bycedeq".
Started build 01cav4jcjea8wzqsvs1bycedeq as "brigade-worker-01cav4jcjea8wzqsvs1bycedeq"
prestart: src/brigade.js written
error An unexpected error occurred: "Command failed.
Exit code: 1
Command: sh
Arguments: -c node --no-deprecation ./dist/index.js
Directory: /home/src
Output:
".
```

### DIFF
--- a/brigade-worker/src/brigadier.ts
+++ b/brigade-worker/src/brigadier.ts
@@ -49,15 +49,21 @@ export function fire(e: eventsImpl.BrigadeEvent, p: eventsImpl.Project) {
  * executed as-is.
  */
 export class Job extends jobImpl.Job {
+  jr: JobRunner;
+
   run(): Promise<jobImpl.Result> {
-    let jr = new JobRunner(this, currentEvent, currentProject);
-    this._podName = jr.name;
-    return jr.run().catch(err => {
+    this.jr = new JobRunner(this, currentEvent, currentProject);
+    this._podName = this.jr.name;
+    return this.jr.run().catch(err => {
       // Wrap the message to give clear context.
       console.error(err);
-      let msg = `job ${this.name}(${jr.name}): ${err}`;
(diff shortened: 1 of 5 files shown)


technosophos:
Can you wrap each of the `slack.run()` and `slack.logs()` files in `try/catch` blocks to see which of those is erroring out? I am guessing that there is some initialization issue lurking beneath the surface, but I can't find it yet.

technosophos:
Okay, I finished manually testing this.

```javascript

events.on("exec", () => {
  const j = new Job("a", "alpine:3.7");
  j.tasks= [ "echo hello", "echo goodbye" ];
  j.run().then(() => { return j.logs() }).then( (rez) => {console.log(rez)})
});
```

This seems to work as designed.

Issue #421, opened 2018-04-10
Allow Annotation of Worker Pods


blimmer:
We use [kube2iam](https://github.com/jtblin/kube2iam) to manage the roles assumed by our services. This is used along with KMS keys to allow pods to decrypt secrets. We'd prefer to use this method, opposed to Brigade Secrets + injecting secrets via the Job ENV variable.

To accomplish this, we'd need to be able to pass an annotation to at least the Job pods. In straight-up k8s it would look something like this ([docs](https://github.com/jtblin/kube2iam#kubernetes-annotation)):

```
apiVersion: v1
kind: Pod
metadata:
  name: aws-cli
  labels:
    name: aws-cli
  annotations:
    iam.amazonaws.com/role: role-arn
spec:
  containers:
  - image: fstab/aws-cli
    command:
      - "/home/aws/aws/env/bin/aws"
      - "s3"
      - "ls"
      - "some-bucket"
    name: aws-cli
```

I think we could go a couple of ways with this:

## Define at the Project Level
This would expose the option on the deployed `brigade/brigade-project` values.yml file. I think in this case, the worker and all jobs would get this annotation.

## Define on the `new Job` in JS
This would define per-pod job. e.g.

```javascript
let job = new Job('something', 'someimage');
job.annotations = {  'iam.amazonaws.com/role': 'role-arn' };
```

Thoughts?

technosophos:
Closing now that #425 is merged.

PR #425, opened 2018-04-13
Add job annotations

onyxraven:
This adds the ability to add annotations to brigade job pods from the brigade.js brigade-worker.

We ran across the need for this to be able to run workloads in ec2 and take advantage of [kube2iam](https://github.com/jtblin/kube2iam#kubernetes-annotation), which needs an annotation.

I tested this in minikube, adding a dummy annotation to my brigade pod:

```javascript
  // Create a new job
  var node = new Job("copy-staging-db");

  // snip

  node.tasks = [
    'env',
    'hostname'
  ];

  node.annotations = {
    'test/annotation': "annotations/go/here"
  };

  // We're done configuring, so we run the job
  node.run(); 
```
and saw the annotation applied on the pod via the dashboard
screen shot 2018-04-13 at 8 30 37 am

I'm not currently sure this adversely affects anything else - it doesnt seem like it does, the tests were testing to ensure that annotations were not set except for the init container, and those just changed to asserting an empty object.

Happy to answer any other questions.

### DIFF
--- a/brigade-worker/test/job.ts
+++ b/brigade-worker/test/job.ts
@@ -138,7 +138,7 @@ describe("job", function() {
       beforeEach(function() {
         j = new mock.MockJob("my-job");
       });
-      it("is a empty list that can be written", function() {
+      it("is an empty list that can be written", function() {
         assert.deepEqual(j.annotations, {});
         j.annotations['some_kubetoiam/thing'] = 'my/path';
         assert.deepEqual(j.annotations, { 'some_kubetoiam/thing': 'my/path' });
--- a/brigade-worker/test/job.ts
+++ b/brigade-worker/test/job.ts
@@ -134,5 +134,15 @@ describe("job", function() {
         assert.isFalse(j.storage.enabled);
       });
     });
+    describe("#annotations", function() {
+      beforeEach(function() {
+        j = new mock.MockJob("my-job");
+      });
+      it("is a empty list that can be written", function() {
+        assert.deepEqual(j.annotations, {});
+        j.annotations['some_kubetoiam/thing'] = 'my/path';
+        assert.deepEqual(j.annotations, { 'some_kubetoiam/thing': 'my/path' });
+      });
+    });
   });
 });
--- a/brigade-worker/src/job.ts
+++ b/brigade-worker/src/job.ts
@@ -217,6 +217,11 @@ export abstract class Job {
    */
   public docker: JobDockerMount;
 
+  /**
+   * pod annotations for the job
+   */
+  public annotations: { [key: string]: string; } = {};
+
   /** _podName is set by the runtime. It is the name of the pod.*/
   protected _podName: string;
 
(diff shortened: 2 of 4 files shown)


onyxraven:
Yep, we'll look into it. 

for the CLA, I'm awaiting corporate approval (I can't imagine there will be a problem).

technosophos:
I am dying with laughter (and/or embarrassment) right now. I just clicked on your user name, and saw that it was you, Justin!

jdotjdot:
I think it's worth noting that bit by bit I'm observing Javascript features being added that just replicate (or half-replicate) all of the features that already exist in the Kubernetes YAML.

Would it be worth just adding an option that lets Brigade simply grab a YAML declaration and run it?  The current `Job` structure can be kept to make it easier for people who don't know (or what to learn) complex Kubernetes set-ups, but for those who want that control or have pre-existing container configs they want to run, it could make all of this a lot easier.

onyxraven:
@jdotjdot that's interesting, but I wonder if that exposes too much of the internals of how the job actually runs - for example, if it were to change from a Pod to a k8s Job.  In this case, the args we're specifying (args, tasks, annotations) are common among the final outputs.

@technosophos What are you looking for as far as real testing? In minikube, it applied the annotations I supplied, or are you looking to ensure it all works on a full k8s install?

technosophos:
@jdotjdot we're trying to be careful about how much of Kubernetes to expose. The main reason is that we need to be able to effectively manage the job objects, and that means doing a certain amount of stuff "for you", and making it really hard to break architectural features. And at the same time, we're trying to make it easy enough to use that a user needn't learn the details of Kubernetes to write a basic script.

That said, we're also open to exposing deeper parts of Kubernetes as we discover the need (provided doing so is "safe"). This is why I am open to adding many of the things you and others have been asking for.

In short, we're just shooting for a reasonable abstraction that hides our own implementation details, protects certain assumptions, and still meets the needs of developers.

@onyxraven I think I'll give this one a spin on my minikube cluster tomorrow. I was assuming that this needed some AWS-specific testing, but it looks like that's actually just incidental to the feature.

onyxraven:
@technosophos signed. Let me know if theres anything I can help with testing

blimmer:
Btw this fixes #421

brigade-worker/test/job.ts
```diff
@@ -134,5 +134,15 @@ describe("job", function() {
         assert.isFalse(j.storage.enabled);
       });
     });
+    describe("#annotations", function() {
+      beforeEach(function() {
+        j = new mock.MockJob("my-job");
+      });
+      it("is a empty list that can be written", function() {
```
technosophos: Very minor non-blocking nit: "is an empty list"

onyxraven: lol, i can fix that up.

blimmer:
Thanks @onyxraven , this will help to unblock our kube2iam usage :-)

PR #428, opened 2018-04-13
feat(Job): add keepCommand option to Jobs

jdotjdot:
The `JobRunner` currently always sets the `job.tasks` array to the `command` option, which maps to Docker's `ENTRYPOINT`.  This can be a problem for some containers that have entrypoints that take actions before executing the intended command.  (For example, for us, we have an entrypoint script that authenticates the container to Vault and grabs secrets.)

This PR adds a `keepCommand` option that permits the user to specify on a per-job basis that Brigade should instead put the content of `tasks` into `args` (Docker's `CMD`) rather than `command`.  We're using this in production right now and it seems to work.

### DIFF
--- a/brigade-worker/src/job.ts
+++ b/brigade-worker/src/job.ts
@@ -156,6 +156,10 @@ export abstract class Job {
   public name: string;
   /** shell that will be used by default in this job*/
   public shell: string = defaultShell;
+  /** keepCommand determines whether Brigade runs the job
+   *  with the default container command (entrypoint) or whether it overrides and
+   *  removes it.*/
+  public keepCommand: boolean = false;
   /** tasks is a list of tasks run inside of the shell*/
   public tasks: string[];
   /** env is the environment variables for the job*/
--- a/brigade-worker/src/k8s.ts
+++ b/brigade-worker/src/k8s.ts
@@ -193,7 +193,8 @@ export class JobRunner implements jobs.JobRunner {
       job.image,
       job.imageForcePull,
       this.serviceAccount,
-      job.resourceRequests
+      job.resourceRequests,
+      job.keepCommand
     );
 
     // Experimenting with setting a deadline field after which something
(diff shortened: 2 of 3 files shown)


technosophos:
Uh-oh... I think we were both working on a similar problem. See #429, which allows you to explicitly set `Job.args` (and which I released in 0.13).  We can chat on Monday and see if I screwed things up with #429.

jdotjdot:
Haha it would be a bit strong to say you screwed anything up, even if it did interfere with this!

#429 may solve part of what I was looking for, but more as a byproduct than anything else. The issue we have is that we're pre-baking Docker containers with `ENTRYPOINT`s that we want to keep&mdash;and by running `Job`s in Kubernetes with setting `command`, we lose the default `ENTRYPOINT`.  I very specifically want our pipeline authors to have the option of having the benefit of the default entrypoint (if they want it), and only setting the Kubernetes `args` and leaving `command` alone.

With #429, users can now edit `args`, and it happens to be that Brigade will no longer override `command` in that scenario&mdash;but really it should be more explicit, and you should get the choice of (a) do I want to keep the container's `ENTRYPOINT` or get rid of it, and (b) do I want to use the `tasks` scripting or use `args` instead.   (Or maybe I want to use both? I don't see why they need to be mutually exclusive.)

technosophos:
As of #429, this is what happens if you specify tasks and args:

- tasks are combined into a shell script (as always) and then stored in the secret as `main.sh`
- command is set to `["/bin/sh", "/hook/main.sh"]` (which is where the above is mounted on the pod)
- args (on the container) are set to the contents of `Job.args`

The Kubernetes documentation is not entirely clear on how it resolves this situation. I guess I should test it.

jdotjdot:
In that situation, what I believe will happen is `command` will override
`ENTRYPOINT` and `args` will override `CMD`, which would cause my problem
to still be present.

radu-matei:
Hi, @jdotjdot!
I somehow seem to have missed this PR, I apologize!

A couple of questions:
- is the behaviour described here still an issue for your use case?
- would you be willing to rebase off `master`?

Note that the TypeScript library for Brigade is in a separate repo - see https://github.com/Azure/brigadier

Thanks!

jdotjdot:
We're not using Brigade anymore internally, so I'm happy to have this PR repurposed however is best for you all.  It's not an issue for our use case anymore since we haven't used Brigade in a while.

krancour:
This is quite stale, so I am closing, but all of this _has_ been addressed in v2.

PR #429, opened 2018-04-13
feat(worker): support passing args through jobs

### LINKED ISSUE #411: How to use entrypoint
I am using brigade to launch a docker container.  In my dockerfile, I use the ENTRYPOINT keyword like so:

ENTRYPOINT ["python3", "runMyScript.py"]

If I were to run this container via the command-line, I would do something like this to specify a config file parameter:

sudo docker run -it containerName--ConfigFile "config/config.txt"

I noticed that if I add "--ConfigFile config/exampleconfig.txt" to the tasks of my brigade job, brigade complains because it doesn't know what to do with this string; it is expecting something like "python3 runMyScript.py --ConfigFile=config/exampleconfig.txt".

We want to be able to bake the entrypoint into our dockerfile and pass command-line parameters for the python script through the brigade.js file to customize the run.  I.E. I want to be able to mimic the behavior of the "sudo docker run -it containerName --ConfigFile config/config.txt" command in my brigade file.

Does brigade support this behavior?

technosophos:
Adds `Job.args: []string` so that users can set `args` on a container directly, bypassing the `tasks` scripting behavior.

Also fixes a few test errors. 

Closes #411

### DIFF
--- a/brig/cmd/brig/commands/run.go
+++ b/brig/cmd/brig/commands/run.go
@@ -33,7 +33,7 @@ var (
 	runNoProgress bool
 )
 
-var logPattern = regexp.MustCompile("\\[brigade:k8s\\]\\s[a-zA-Z0-9-]+/[a-zA-Z0-9-]+ phase \\w+")
+var logPattern = regexp.MustCompile(`\[brigade:k8s\]\s[a-zA-Z0-9-]+/[a-zA-Z0-9-]+ phase \w+`)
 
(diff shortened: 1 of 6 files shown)


technosophos:
Example usage:

```javascript

events.on("exec", () => {
  const j = new Job("a", "alpine:3.7");
  j.args = [ "echo", "hello" ];
  j.run();
});
```

PR #432, opened 2018-04-18
Use .stringData to avoid base64 encoding & decoding.

julianvmodesto:
Related to https://github.com/Azure/brigade/issues/419.

Using `.stringData` makes working with Secret data simpler because clients don't need to base64 encode values back-and-forth when editing.

As a Brigade user using Helm, this improves readability of the templated Helm chart.

As a Brigade user not using Helm, I can copy templates/secret.yaml to create a new project and easily edit my project configuration in plain text.

Tested by uncommenting most variables in the default values.yaml and running:

```
helm template charts/brigade-project -n myProject | kubectl apply --dry-run --validate -f -
```

### DIFF
--- a/charts/brigade-project/templates/secret.yaml
+++ b/charts/brigade-project/templates/secret.yaml
@@ -11,44 +11,45 @@ metadata:
   annotations:
     projectName: {{ .Values.project | quote }}
 type: "brigade.sh/project"
-data:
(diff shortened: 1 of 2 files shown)


bacongobbler:
background history on the difference between `data` and `secretData`, in case anyone is as ignorant on the subject as I am: https://github.com/kubernetes/kubernetes/issues/19575

Issue #433, opened 2018-04-20
Add caching in Brigade API server


technosophos:
Some of the calls on Brigade API take way too long to execute every single page load, and are amenable to caching.

Prime example: `/projects-build`

walialu:
I second that! We have about 30 projects (or so) so far and the brig-api takes about 10 seconds to load the `/projects-builds` endpoint.

jensneuse:
I'm currently in the works of solving the problem:
https://github.com/jensneuse/brigade/tree/improve-brig-api-performance

I've introduced a cache.NewListWatchFromClient for both secrets and pods which drops the request time for `/projects-builds` to <100ms.

Currently everything works fine running locally on my local machine. However I'm having issues passing all tests. Aside from that it looks like the current implementation of fake.NewSimpleClientset() doesn't work as expected for ListWatchers which makes it hard to unit test this scenario. (Maybe I oversaw something) Help is highly appreciated.

I'm looking forward to getting feedback on whats needed to get this into the upstream.

jensneuse:
I was able to resolve all the issues, tests are passing now. I'd be happy to get feedback on the PR. Cheers

PR #446, opened 2018-04-29
Improve brig api performance as seen in #433 by introducing caching

jensneuse:
This PR introduces api caching using the "k8s.io/client-go/tools/cache" package.
As a result of #433 I saw the necessity to optimize the route "/projects-build".

After not seeing any improvements via the "naive" implementation I've setup some benchmarks to analyse the problem.

I figured out that there were way too many requests towards the kubernetes api to fetch secrets, each of these was taking about 200ms.

The solution I'd like to introduce is caching all secrets of the brigade namespace for some reasonable time. This way there are zero network requests left during the invocation of "projects-build".

All in all this reduces the response time for our environment with a dozen pipelines from 10 seconds to 100ms.

Everything I touched is 100% covered.

I'm looking forward to getting this PR into the upstream.

### DIFF
--- a/pkg/api/project.go
+++ b/pkg/api/project.go
@@ -32,6 +32,7 @@ type ProjectBuildSummary struct {
 
 // ListWithLatestBuild lists the projects with the latest builds attached.
 func (api Project) ListWithLatestBuild(c *gin.Context) {
+
 	projects, err := api.store.GetProjects()
 	if err != nil {
 		c.JSON(http.StatusNotFound, struct{}{})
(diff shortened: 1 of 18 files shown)


technosophos:
This is what I am getting with `make test-style` right now:

```
gometalinter --config ./gometalinter.json ./...
pkg/storage/kube/apicache/liststore.go:13:1:warning: cacheStoreFactory is unused (deadcode)
pkg/storage/kube/apicache/liststore.go:13:6:warning: type cacheStoreFactory is unused (U1000) (unused)
pkg/storage/kube/apicache/liststore.go:14:2:warning: func cacheStoreFactory.new is unused (U1000) (unused)
pkg/storage/kube/store_test.go:145:6:warning: type fakeWatcher is unused (U1000) (unused)
make: *** [test-style] Error 1
```

I'm going to go ahead and build a test locally while you fix that stuff. We're really close at this point

technosophos:
Manually tested with Kashti.

WOW, that's fast!

PR #456, opened 2018-05-16
Feature: Configmap as scriptstore

FrodeHus:
**Use case**: Keep generic brigade.js scripts in a common store when you wish to avoid duplicate defaultScript entries in multiple project configurations as well as in each repo.

**Solution**: Add optional property _defaultScriptName_ that refers to the name of a config map containing the brigade.js.
Order of precedence: repo > configmap > defaultscript

Scripts can then be stored by executing the following command:
```kubectl create cm mycommonscript --from-file=./brigade.js```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ node_modules/
 
 # Developer
 pvt/
+debug.test
--- a/brigade-controller/cmd/brigade-controller/controller/handler.go
+++ b/brigade-controller/cmd/brigade-controller/controller/handler.go
@@ -1,6 +1,7 @@
 package controller
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"log"
@@ -12,6 +13,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/Azure/brigade/pkg/storage/kube"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
 )
 
 var (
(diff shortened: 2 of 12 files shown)


FrodeHus:
Sounds like a much better and generic approach :) I will look it and update my PR.

FrodeHus:
I haven’t gone in-depth, but it seems gateways bail out if they can’t find brigade.js. 
The controller takes over when the build is created so my thoughts right now is that the gateways could check if the controller can handle it (s.store.IsDefaultScriptPresent or the likes) 

Would it be acceptable if the gateways are rewritten to do this check? 
As an example - Github GW checks if there is a script in the repo and then if one exists in the project definition. If none are there; it doesnt create the build and the controller isn’t involved at all.

Or:
How about allowing anyone to create a build, but if the controller deems it invalid (no script added by the gateway and none found by the controller the build is tagged as invalid)?

FrodeHus:
Opted for a much simpler variant: As it seems the gateways use project.DefaultScript directly I changed the implementation so that if DefaultScriptName exists, it will load the script from configmap and place it in DefaultScript. If it fails, it will merely use whatever DefaultScript already had.

technosophos:
Thank you again for all the work you put in on this. And I especially appreciate the fact that you've refactored this a few times already. I'm sorry we ended up in the situation where we had two alternative implementations. I marked #467 as okay to merge, largely because I felt more comfortable with pushing the config map loading down to Kubernetes, and because it didn't change the storage interface. But I wanted to make it clear that this just ends up being one of those uncomfortable situations where two people have both done outstanding work. Thanks for submitting this, and I hope you will continue to submit PRs for improvements.

PR #461, opened 2018-05-17
Remove az extension add for ACR build

### LINKED ISSUE #460: Remove az extension add for ACR Build
ACR Build has been merged into the official Azure CLI, so we can remove the extensions since they're no longer supported. :tada:

There's one breaking change for this, `--context` has been removed in favor of positional arguments. This follows the same pattern Docker does, i.e., you can do `az acr build -f Dockerfile -t hello -t world .` now instead of having to do `--context .`

ehotinger:
Fixes #460 

@technosophos can you test this out?


technosophos:
Yup! Thanks.

PR #467, opened 2018-05-22
feat(*): add project option for fallback script

adamreese:
Add a global script as a fallback if the build is not provided one
directly or checked in to VCS.  The script is stored as a configmap and
specified in project.defaultScriptName.

### DIFF
--- a/brigade-controller/cmd/brigade-controller/controller/controller_test.go
+++ b/brigade-controller/cmd/brigade-controller/controller/controller_test.go
@@ -190,7 +190,8 @@ func TestController_WithScript(t *testing.T) {
 		},
 		// This and the missing 'script' will trigger an initContainer
 		Data: map[string][]byte{
-			"vcsSidecar": []byte(sidecarImage),
+			"vcsSidecar":        []byte(sidecarImage),
+			"defaultScriptName": []byte("my-script"),
 		},
 	}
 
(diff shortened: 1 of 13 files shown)


FrodeHus:
#456 offers two ways of handling it to provide a backwards compatibility for existing gateways by project.DefaultScript containing either DefaultScript or ConfigMap referenced by DefaultScriptName.
In addition, if no script is provided by the build - the controller will look for either DefaultScript in the project or the ConfigMap specified and update the build before creating the worker.

This should allow the feature to work with existing gateways that may use DefaultScript without code change or let the controller handle it.

technosophos:
At this point, I have audited all of the gateways I know about to see what the potential impacts and drawbacks would be. Both this PR and #456 are converting on the same solution. I think we should merge this one simply because it lets Kubernetes do the main work of loading the configmap.

FrodeHus:
True. My only thought was by updating the build instead of mounting the configmap, you can review which script was run at the time of build. 
By mounting the configmap, you can only view the current version of the script and not which one was actually there when the build ran.

technosophos:
That is a really good point. I guess to some extent that is already true of the VCS sidecar if it is tracking a branch (e.g. refs/heads/master), but at least there you have the vcs itself to fall back on.

technosophos:
Alright. I have been trying to think through the various implications. We could still pull in @FrodeHus 's method of copying the script from the ConfigMap into the event if that is looking like the best way. Now that we merged the `brig rerun` command, this might be something we should seriously consider.

PR #474, opened 2018-05-23
feat(brig): add rerun command

### LINKED ISSUE #463: how do you re-run a previous build with brig CLI?
Scenario: builds are executing from git pushes to github or bitbucket. These builds are CI/CD and at some point we need to roll-back a deployment to a previous one using a specific build #.

How do I re-run a previous brigade build through the CLI using a specific brigade build #?

This seems to be missing:
e.g. `brig run myproject --buildId 01ce2w3w016kz6hqf9k8tmb3dn`
In this scenario I would expect brigade to lookup the buildId, gather the Event data from it and start a new build with that event data.

technosophos:
This adds `brig rerun BUILD_ID`, which allows a replay of an older
build.

Closes #463

### DIFF
--- a/brig/cmd/brig/commands/build.go
+++ b/brig/cmd/brig/commands/build.go
@@ -4,7 +4,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const buildUsage = `manage builds
+const buildUsage = `Manage builds
 
 Work with Brigade builds.
 `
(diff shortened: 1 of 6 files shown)


technosophos:
I verified that the tests passed, and that GitHub was notified of the success. Not sure why this still shows as pending

Strandedpirate:
@technosophos can you do an official release that includes this functionality?

https://github.com/Azure/brigade/releases/tag/v0.14.0 doesn't have these bits.

Issue #473, opened 2018-05-23
Failed jobs are re-run on brigade upgrade


blimmer:
I upgraded today to brigade 0.14.0 and I noticed a ton of runners kicked off and failed right away:

![](https://i.imgur.com/nOIdL1D.png)

From the hashes, you can tell that these secrets were all created inside the window under which the vacuum runs (we hadn't modified from the default 30 days).

![](https://i.imgur.com/F0jS3Ed.png)

They all fail with this exception:

```
   body: 
   { kind: 'Status',
     apiVersion: 'v1',
     metadata: {},
     status: 'Failure',
     message: 'secrets "fetch-microservice-repos-01cckwwb0nz7bzy42tze6ra1z0" already exists',
     reason: 'AlreadyExists',
     details: 
      { name: 'fetch-microservice-repos-01cckwwb0nz7bzy42tze6ra1z0',
        kind: 'secrets' },
     code: 409 } } 
```

So our best guess on what's happening is that the new upgraded brigade components come up and read all the old secrets and then kick off the workers as if they were just triggered by the gateways.

Is this expected behavior?

---

cc @awwithro who was debugging this with me

technosophos:
This might be a duplicate of #462 

This is not the intended behavior. I tried this yesterday and didn't reproduced, but I think that's because I had already done some cleaning earlier in the week that would have prevented this from showing up.

technosophos:
Looks like @adamreese found the source of this, and is fixing it.

blimmer:
Great - thanks!

PR #482, opened 2018-05-25
fix(controller): add an accepted label to builds after sync

adamreese:
Resolves: #473

### DIFF
--- a/brigade-controller/cmd/brigade-controller/controller/handler.go
+++ b/brigade-controller/cmd/brigade-controller/controller/handler.go
@@ -21,44 +21,56 @@ var (
 	containerImageRegex = regexp.MustCompile("(.*):([^:]+)$")
 )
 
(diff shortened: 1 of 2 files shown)


brigade-controller/cmd/brigade-controller/controller/handler.go
```diff
@@ -21,44 +21,56 @@ var (
 	containerImageRegex = regexp.MustCompile("(.*):([^:]+)$")
 )
 
-func (c *Controller) syncSecret(secret *v1.Secret) error {
+func (c *Controller) syncSecret(build *v1.Secret) error {
+	// Ensure this secret has not yet been handled.
+	if status := build.Labels["status"]; status == "accepted" {
```
technosophos: Can this be simplified to just `if build.Labels["status"] == "accepted"`?

PR #487, opened 2018-05-30
Fix the check for data in prestart.js

radu-matei:
It appears to be a bug in `prestart.js` that only returns the script if the file is empty.


technosophos:
Yes, definitely a bug.

PR #492, opened 2018-05-31
Add log follows on worker and job Storage API

slok:
Hi!

This PR adds 2 new methods to the `Storage` interface API. `Storage.GetJobLogStreamFollow` and `storage.GetWorkerLogStreamFollow`. Just adds the `follow` flag to Kubernetes so we retrieve an `io.ReadCloser` that will follow the logs.

I created 2 new methods instead of adding a flag on the `Get(Worker|Job)LogStream` methods so the API continues being compatible with the current users and doesn't break current usage.

I don't think that requires testing this part, just uses the client-go (that it's very well tested :heart:), without domain logic at all.


slok:
Thank you! :)

PR #507, opened 2018-06-12
fix(storage): Improve storage layer for projects

technosophos:
Previously, storage did not need to know much about projects. This adds
better support for projects to the storage layer, and adds a
CreateProject function.

### DIFF
--- a/pkg/storage/mock/storage.go
+++ b/pkg/storage/mock/storage.go
@@ -2,6 +2,7 @@ package mock
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"time"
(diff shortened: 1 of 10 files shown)


pkg/storage/mock/storage.go
```diff
@@ -60,18 +61,18 @@ var (
...
+		Job:         StubJob,
+		LogData:     StubLogData,
 	}
 }
 
 // Store implements the storage.Storage interface, but returns mock data.
 type Store struct {
 	// Project is the project you want returned.
-	Project *brigade.Project
+	//Project *brigade.Project
```
adamreese: delete me

PR #510, opened 2018-06-14
docs(README) update main docs link

flynnduism:
Small follow up to #509 - to highlight the site at [https://azure.github.io/brigade](https://azure.github.io/brigade)

---

That site is  more navigable and user-friendly, so this updates the root README to use that as the
primary docs URL.


flynnduism:
@bacongobbler I didn't know it existed until today! thanks

PR #512, opened 2018-06-15
feat(brig): add brig project delete

technosophos:
This adds a command to delete a project.


technosophos:
This also contains a few minor refactors in anticipation of `brig project create`, which I will PR as soon as this one gets merged.

PR #513, opened 2018-06-15
docs(README) Point all docs links to azure.github.io/brigade

jarimatti:
Point all README.md documentation links to the site https://azure.github.io/brigade.

Also fix markup in the first link. This was initially the main point, but I went a bit overboard with this: please let me know if I did.

This is related to #510

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ for Kubernetes.
 - Project-based management
 - Configurable event hooks
 - Easy construction of pipelines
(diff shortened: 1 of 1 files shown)


flynnduism:
Ah, I just noticed the formatting of `docs]` and was about to do the same. 

Good to have the other links point to the same site too for continuity IMO

jarimatti:
Thanks! And I must say I like Brigade so far: it's been a pleasure working with it.

PR #514, opened 2018-06-18
Allow for customization of API and gateway paths

hharnisc:
This keeps the same defaults, but allows for path customization. Currently it's not possible to get Brigade working on GKE with the [gce ingress](https://github.com/kubernetes/ingress-gce) since it does paths a little differently.

gce wants `/*` for all paths rather than `/`

I've added the ability to customize the api and the gateway ingresses.


technosophos:
Thanks! I will definitely get this one merged once the defaults are set. I didn't know about the GKE thing, so thanks for fixing that!

Issue #517, opened 2018-06-19
Job Name Regex is inconsistent with Docs


slater-brian-john:
If I try to name a Job consistent with line 283 of scripting.md

"- a _name_: The name must be unique to the event handler and should be composed of letters, numbers, and dashes (`-`)."

And use name: sbx-Brian, I will get an error. I added an 'i; to the end of the regex check in /brigade-worker/job.ts and that seemed to clear things up. Was not able to push to this repo with my new branch so I decided to created an issue. 

Regardless, either docs are slightly off or the regex for job name evaluatation is a bit off.

technosophos:
Job names might have to be lowercase (because Kubernetes restricts to DNS-style RFC 1035 names). We should check this and update the docs.

slater-brian-john:
I will take a look then PR it, thank you

slater-brian-john:
link to PR, I opted to change the name at creation rather than allow user error. Should I update docs to explain this or leave it as is? 

https://github.com/Azure/brigade/pull/522

slater-brian-john:
2.3.3. Character Case

For all parts of the DNS that are part of the official protocol, all
comparisons between character strings (e.g., labels, domain names, etc.)
are done in a case-insensitive manner.  At present, this rule is in
force throughout the domain system without exception.  However, future
additions beyond current usage may need to use the full binary octet
capabilities in names, so attempts to store domain names in 7-bit ASCII
or use of special bytes to terminate labels, etc., should be avoided.

When data enters the domain system, its original case should be
preserved whenever possible.  In certain circumstances this cannot be
done.  For example, if two RRs are stored in a database, one at x.y and
one at X.Y, they are actually stored at the same place in the database,
and hence only one casing would be preserved.  The basic rule is that
case can be discarded only when data is used to define structure in a
database, and two names are identical when compared in a case
insensitive manner.

slater-brian-john:
Seems safe to do given notes from spec

technosophos:
I agree

technosophos:
I think this issue can be closed now that the PR is merged, right?

slater-brian-john:
Closing :)

PR #522, opened 2018-06-22
517: fix for job name error cause by mixed case job names

slater-brian-john:
This is a fix for issue #517. I opted to directly set the name to lowercase for the user rather than allow them to make a mistake


slater-brian-john:
Awesome, thank you. I had not thought of that, but is quite interesting

PR #531, opened 2018-07-03
implement `brig dashboard`

bacongobbler:
`brig proxy` is identical to `brig dashboard --open-dashboard=false`, so I decided to remove it in favour of `brig dashboard`. This also fixes an issue where the tunneled connection was not being closed on `brig proxy` (now `brig dashboard`).


bacongobbler:
hold up; bumping bacongobbler/browser to v1.0.1 for a security patch: https://github.com/bacongobbler/browser/commit/792d0443b3e34e0a99a68e254084079ad9356db5

Gopkg.toml
```diff
@@ -34,7 +34,3 @@
 [[constraint]]
   branch = "master"
   name = "github.com/Masterminds/kitt"
-
```
technosophos: github.com/bacongobbler/browser should be in here, since it's a direct dependency.

bacongobbler: good catch

bacongobbler: done

brig/cmd/brig/commands/dashboard.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+func init() {
+	Root.AddCommand(dashboard)
+
+	flags := dashboard.PersistentFlags()
+	flags.IntVar(&port, "port", 8081, "local port for the Kashti dashboard")
+	flags.IntVar(&apiPort, "api-port", 7745, "local port for the Brigade API server")
+	flags.StringVar(&kashtiNamespace, "kashti-namespace", "default", "namespace for Kashti")
+	flags.BoolVar(&openDashboard, "open-dashboard", true, "open the dashboard in the browser")
```
radu-matei: Should the default be `true` here?
For example this always fails on Bash on Windows (WSL) as `xdg-utils` is not installed by default. (an manually adding it with `apt-get` still won't work). This means the command always panics.

brig/cmd/brig/commands/dashboard.go
```diff
@@ -0,0 +1,99 @@
...
+		log.Println("Connected! When you are finished with this session, enter CTRL+C.")
+
+		if openDashboard {
+			if err := browser.Open(fmt.Sprintf("http://localhost:%d", port)); err != nil {
+				return err
+			}
+		}
+
+		// block until the user sends a CTRL+C
+		for {
```
radu-matei: The empty `for` loop here always blocks - at least when compiled with my Go versions (`go version go1.9.2 windows/amd64`, `go version go1.10.3 linux/amd64` and I think on `darwin` as well) - meaning it opens the browser, but the proxy doesn't correctly starts.

Adding instructions here (printing, waiting) seems to avoid the issue - will come back after investigating a bit.

brig/cmd/brig/commands/dashboard.go
bacongobbler: oh good point. I forgot about WSL. I'll see if there's a way we can detect that we're running under WSL. It's a bug in https://github.com/bacongobbler/browser

@technosophos any thoughts on flipping this flag off by default?

radu-matei: I've done searched about detecting WSL and it should not be that difficult. 

From WSL:

```
$ cat /proc/sys/kernel/osrelease
4.4.0-17134-Microsoft

$ cat /proc/version
Linux version 4.4.0-17134-Microsoft (Microsoft@Microsoft.com) (gcc version 5.4.0 (GCC) ) #137-Microsoft Thu Jun 14 18:46:00 PST 2018

$ uname -a
Linux <PC-name> 4.4.0-17134-Microsoft #137-Microsoft Thu Jun 14 18:46:00 PST 2018 x86_64 x86_64 x86_64 GNU/Linux
```

So checking either of those should tell us if running under WSL or not.

I'm not sure the default should be `false` though - if we detect it's WSL and output that we cannot open the browser, then everybody else (which I think is a bigger number) can still benefit from the default.

What do you think?

bacongobbler: Yeah I think we can special-case WSL. We might even be able to detect we're running under WSL and call `cmd /c start` from within WSL. I'll investigate

bacongobbler: https://github.com/bacongobbler/browser/issues/3

radu-matei: Nice! I think I like your solution more!
Should this case be treated by `bacongobbler/browser`?

Issue #504, opened 2018-06-11
META: Brigade 1.0.0 


technosophos:
This ticket is a placeholder for capturing the Brigade 1.0.0 requirements. We have three major areas to polish before Brigade 1.0.0: Installation, Core, and UX

## Installation

- [x] Create new repo with all charts for Brigade (Kashti, gateways, etc) _updated Oct 2018_ https://github.com/Azure/brigade/issues/718
  - [x] Improve the `brigade` chart to make it easier to configure *edit (vadice): sounds like including Kashti (as below) was the main task here; so marking as done*
  - [x] Kashti included in base Brigade chart (and documented) https://github.com/Azure/brigade-charts/issues/7
  - [x] Include other gateway charts here https://github.com/Azure/brigade/issues/718
- [x] Add `brig dashboard` command to open tunnel to Kashti and then connect to browser
- [x] `brig project create|delete|update` commands added
  - [x] And the deprecation of the `brigade-project` chart
  - [x] And the documentation of using `kubectl edit` to edit a brig project
  - [x] Remove the `brigade-project` chart 
  - [x] Bring `brigade project create` into parity with the current chart #745
- [x] Add generic webhook support to API server (and make it configurable to turn off) #602
- [x] GitHub moved to new gateway #746

## Core/Internals 🎉 

- [x] Make Kashti nginx proxy traffic to the API server #535
- [x] Fix sort order bugs on API server post-cache update (See #511)
- [x] Improve Git sidecar (cf. #668)
- [x] Extract `brigadier` into external library and then use it within brigade worker #584 

## User Experience

- [x] Improve documentation #747
- [x] ~~Examples and tutorials for common Brigade scenarios #748~~ -> moved to v1.1
- [x] [Kashti - project board](https://github.com/Azure/kashti/projects/4) - The Kashti project board can be tracked independently (vdice)

squillace:
+1 :-)

technosophos:
Made a few updates to this based on feedback from core devs and PMs. Changes are tagged *update*.

/cc @squillace @radu-matei

lukepatrick:
Is `brig project create` intended to replace `helm install brigade-project` or `helm inspect values brigade/brigade-project > myvalues.yaml` ?

technosophos:
Both (with an option to still edit a file by hand). Here's what my WIP looks like:

[...]

PR #534, opened 2018-07-06
feat(brig): Interactive prompt for creating projects 

technosophos:
This adds `brig project create` to create projects from interactive
prompts or from an existing secret.

See #504

### DIFF
--- a/brig/cmd/brig/commands/project_create.go
+++ b/brig/cmd/brig/commands/project_create.go
@@ -1,0 +1,521 @@
+package commands
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/AlecAivazis/survey"
+	"github.com/Masterminds/goutils"
+	"github.com/spf13/cobra"
+
+	"k8s.io/api/core/v1"
+
+	"github.com/Azure/brigade/pkg/brigade"
+	"github.com/Azure/brigade/pkg/storage/kube"
+)
+
+const projectCreateUsage = `Create a new project.
+
+Create a new project by answering questions or supplying a configuration file.
+
+Project names are typically in the form 'ORG/PROJECT' or 'USER/PROJECT". For
+example, Azure/brigade or deis/empty-testbed.
+
+A Brigade project provides a context in which a brigade.js file is executed.
+Projects are frequently tied to (Git) source code repositories, and contain
+configuration data such as secrets, Kubernetes-specific directives, and
+settings for gateways or workers.
+
+If a file is provided (-f), then the project defaults will be set from the file.
+
+The user will be prompted to answer questions to build the configuration. If -f
+is specified in conjunction with -x/--no-prompts, then the values in the file will
+be used without prompting the user for any changes.
+
+`
+
+var (
+	projectCreateConfig      = ""
+	projectCreateOut         = ""
+	projectCreateFromProject = ""
+	projectCreateDryRun      = false
+	projectCreateNoPrompts   = false
+	projectCreateReplace     = false
+)
+
+var defaultProject = &brigade.Project{
+	Name: "deis/empty-testbed",
+	Repo: brigade.Repo{
+		Name:     "github.com/deis/empty-testbed",
+		CloneURL: "https://github.com/deis/empty-testbed.git",
+		SSHKey:   "",
+	},
+	Secrets:      map[string]string{},
+	SharedSecret: "",
+	Github:       brigade.Github{},
+	Kubernetes:   brigade.Kubernetes{},
+	Worker: brigade.WorkerConfig{
+		PullPolicy: "IfNotPresent",
+	},
+}
+
+func init() {
+	project.AddCommand(projectCreate)
+	flags := projectCreate.Flags()
+	flags.StringVarP(&projectCreateConfig, "config", "f", "", "Path to JSON Kubernetes Secret.")
+	flags.StringVarP(&projectCreateFromProject, "from-project", "p", "", "Retrieve the given project from Kubernetes and use it to set the default values.")
+	flags.StringVarP(&projectCreateOut, "out", "o", "", "File where configuration should be saved. The configuration is stored as a JSON Kubernetes Secret")
+	flags.BoolVarP(&projectCreateDryRun, "dry-run", "D", false, "Do not send the config to Kubernetes")
+	flags.BoolVarP(&projectCreateNoPrompts, "no-prompts", "x", false, "Do not prompt the user for input. Use this with -f to skip prompts.")
+	flags.BoolVarP(&projectCreateReplace, "replace", "r", false, "Replace an existing project with this revised one. Use this to modify existing projects.")
+}
+
+var projectCreate = &cobra.Command{
+	Use:   "create",
+	Short: "create a new Brigade project",
+	Long:  projectCreateUsage,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if projectCreateFromProject != "" && projectCreateConfig != "" {
+			return errors.New("cannot specify both -f/--config and -p/--from-project.")
+		}
+		return createProject(cmd.OutOrStderr())
+	},
+}
+
+func createProject(out io.Writer) error {
+	c, err := kubeClient()
+	if err != nil {
+		return err
+	}
+
+	store := kube.New(c, globalNamespace)
+
+	proj := defaultProject
+	if projectCreateConfig != "" {
+		if proj, err = loadProjectConfig(projectCreateConfig, proj); err != nil {
+			return err
+		}
+	}
+
+	if projectCreateFromProject != "" {
+		var err error
+		proj, err = store.GetProject(projectCreateFromProject)
+		if err != nil {
+			return fmt.Errorf("could not load %s: %s", projectCreateFromProject, err)
+		}
+	}
+
+	if !projectCreateNoPrompts {
+		if err := projectCreatePrompts(proj); err != nil {
+			return err
+		}
+	}
+
+	secret, err := kube.SecretFromProject(proj)
+	if err != nil {
+		return err
+	}
+	data, err := json.MarshalIndent(secret, "", "  ")
+	if err != nil {
+		return err
+	}
+
+	if globalVerbose {
+		out.Write(data)
+		fmt.Println()
+	}
+
+	if projectCreateOut != "" {
+		if err := ioutil.WriteFile(projectCreateOut, data, 0755); err != nil {
+			return err
+		}
+	}
+
+	fmt.Fprintf(out, "Project ID: %s\n", proj.ID)
+	if projectCreateDryRun {
+		return nil
+	}
+
+	if !projectCreateReplace {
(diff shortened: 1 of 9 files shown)


brig/cmd/brig/commands/project_create.go
```diff
@@ -0,0 +1,537 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/AlecAivazis/survey"
```
bacongobbler: 👍

brig/cmd/brig/commands/project_create.go
```diff
@@ -0,0 +1,537 @@
...
+Projects are frequently tied to (Git) source code repositories, and contain
+configuration data such as secrets, Kubernetes-specific directives, and
+settings for gateways or workers.
+
+If a file is provided (-f), then the project defaults will be set from the file.
+
+The user will be prompted to answer questions to build the configuration. If -f
+is specified in conjunction with -x/--no-prompts, then the values in the file will
+be used without prompting the user for any changes.
+
```
bacongobbler: delete this

brig/cmd/brig/commands/project_create.go
```diff
@@ -0,0 +1,537 @@
...
+`
+
+var (
+	projectCreateConfig      = ""
```
bacongobbler: dont'e need to set these to their default values since `.StringVarP` and `.BoolVarP` already do that after `.ParseFlags()` is called. Should be just `projectCreateConfig string`

brig/cmd/brig/commands/project_create.go
```diff
@@ -0,0 +1,537 @@
...
+		return err
+	}
+	data, err := json.MarshalIndent(secret, "", "  ")
+	if err != nil {
+		return err
+	}
+
+	if globalVerbose {
+		out.Write(data)
+		fmt.Println()
```
bacongobbler: nit: this could be simplified to

```
fmt.Fprintf(out, "%v\n", data)
```

brig/cmd/brig/commands/project_create.go
```diff
@@ -0,0 +1,537 @@
...
+		},
+	}
+
+	abort := "project could not be saved: %s"
+	if err := survey.Ask(qs, &p.Repo); err != nil {
+		return fmt.Errorf(abort, err)
+	}
+
+	// Only prompt for key if clone URL requires it
+	if strings.HasPrefix(p.Repo.CloneURL, "ssh") || strings.HasPrefix(p.Repo.CloneURL, "git+ssh") {
```
bacongobbler: do we accept ssh-style clone URLs, such as `git@github.com:Azure/brigade`? If so should we add a check for that, or should we punt it and document that? Or, perhaps a yes/no prompt for whether we need to attach an SSH key?

brig/cmd/brig/commands/project_create.go
```diff
@@ -0,0 +1,537 @@
...
+		if script := loadFileStr(fname); script != "" {
+			p.DefaultScript = loadFileStr(fname)
+		}
+	}
+
+	return nil
+}
+
+// loadFileValidator validates that a file exists and can be read.
+func loadFileValidator(val interface{}) error {
```
bacongobbler: I see why we have to do this, but do you have any ideas on if there are any cleaner ways to this approach? it feels a little awkward to accept an `interface{}` then immediately cast it into a string.

brig/cmd/brig/commands/project_create.go
technosophos: What am I deleting here?

brig/cmd/brig/commands/project_create.go
technosophos: I believe that to satisfy the Validator interface, we have to do `func (interface{}) error` as the signature.

brig/cmd/brig/commands/project_create.go
technosophos: I'm going to invert the selection, and _not_ ask if the URL is http or https, since those are really the two forms that cannot take an SSH key

brig/cmd/brig/commands/project_create.go
bacongobbler: Yeah, I tried brainstorming if there were any Go hacks we could do to make this cleaner, but this looks as good as it gets. :)

technosophos:
@bacongobbler Okay, this is ready for a re-review

brig/cmd/brig/commands/project_create.go
radu-matei: @technosophos - what revision of this package did you use?
It's not in `Gopkg.toml`, and building off its latest revision currently fails

brig/cmd/brig/commands/project_create.go
```diff
@@ -0,0 +1,540 @@
...
+	},
+}
+
+func createProject(out io.Writer) error {
+	c, err := kubeClient()
+	if err != nil {
+		return err
+	}
+
+	store := kube.New(c, globalNamespace)
```
radu-matei: Configuring `globalNamespace` which is `default` means the namespace configured through the survey is not used - so all projects are created in the `default` namespace, regardless of the value passed.

brig/cmd/brig/commands/project_create.go
```diff
@@ -0,0 +1,540 @@
...
+	}, &configureK8s, nil); err != nil {
+		return fmt.Errorf(abort, err)
+	} else if configureK8s {
+		if err := survey.Ask([]*survey.Question{
+			{
+				Name: "namespace",
+				Prompt: &survey.Input{
+					Message: "Kubernetes Namespace",
+					Help:    "The namespace in which this project applies",
+					Default: p.Kubernetes.Namespace,
```
radu-matei: This is not used, see comment above.

technosophos: Oh, you're right... the value here should just be set to `globalNamespace`.

technosophos:
@radu-matei The namespace thing was fixed. Now `--namespace` is the only way to set _all_ the namespace-related stuff. This appears to be the best way to impose consistency.

docs/topics/projects.md
```diff
@@ -28,7 +28,36 @@ other Git services, and how to extend Brigade to work with other VCS systems.
...
+? Full repository name github.com/deis/empty-testbed
+? Clone URL (https://github.com/your/repo.git) https://github.com/deis/empty-testbed.git
+? Add secrets? No
+? Configure GitHub Enterprise or OAuth? No
+? Configure advanced Kubernetes options? No
+? Configure Brigade worker? No
+? Configure Advanced Brigade Settings? No
+```
+
+You can use `--dry-run --debug` to see the answers to the question without creating
```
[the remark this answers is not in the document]
technosophos: Oops... it was --verbose. Will fix that.

brig/cmd/brig/commands/project_create.go
```diff
@@ -0,0 +1,535 @@
...
+			return err
+		}
+	}
+
+	fmt.Fprintf(out, "Project ID: %s\n", proj.ID)
+	if projectCreateDryRun {
+		return nil
+	}
+
+	if !projectCreateReplace {
```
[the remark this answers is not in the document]
technosophos: I think I would opt for doing that as a separate PR. The `--replace` will walk you through the questions again, but with the old value s pre-populated. I figured this would be useful for people who needed help text and stuff like that.

brig/cmd/brig/commands/project_create.go
```diff
@@ -0,0 +1,535 @@
...
+				Prompt: &survey.Input{
+					Message: "Custom VCS Sidecar",
+					Help:    "The default sidecar uses Git to fetch your repository",
+					Default: p.Kubernetes.VCSSidecar,
+				},
+			},
+			{
+				Name: "buildStorageSize",
+				Prompt: &survey.Input{
+					Message: "Build storage size",
```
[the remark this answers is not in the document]
technosophos: Done

docs/topics/projects.md
```diff
@@ -28,7 +28,36 @@ other Git services, and how to extend Brigade to work with other VCS systems.
...
+a new release. For more, see `brig project create --help`.
+
+### Managing Your Projects
+
+With `brig project create`, your projects are stored in Kubernetes secrets. You
+can save a local copy of that secret using `brigade project create --out=myproject.json`.
+
+If you have already created the secret, you can fetch it from Kubernetes by running
+`brig project get my/project` where `my/project` is the project name you assigned.
```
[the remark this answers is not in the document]
technosophos: I have a branch that does this. I will open it as a PR.

docs/topics/projects.md
```diff
@@ -184,6 +214,9 @@ You can use SSH keys and a `git+ssh` URL to secure a private repository.
 In this case, your project's `cloneURL` should be of the form `git@github.com:Azure/brigade.git`
 and you will need to add the SSH _private key_ to the `values.yaml` file.
 
+When doing `brig project create`, if you choose a URL that looks like it supports
+an SSH key, you will be prompted to enter the path to the SSH private key file.
```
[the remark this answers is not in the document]
technosophos: Rewriting that to make 'use' unambiguous

brig/cmd/brig/commands/project_create.go
```diff
@@ -0,0 +1,535 @@
...
+					Help:    "By default, 50Mi of shared temp space is allocated per build. Larger values slow down build startup.",
+					Default: p.Kubernetes.BuildStorageSize,
+				},
+			},
+			{
+				Name: "buildStorageClass",
+				Prompt: &survey.Input{
+					Message: "Build Storage Class",
+					Help:    "Kubernetes provides named storage classes. If you want to use a custom storage class, set the class name here.",
+					Default: p.Kubernetes.BuildStorageClass,
```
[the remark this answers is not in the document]
technosophos: That is a good idea. I will file an issue to add this in a subsequent PR.

docs/topics/projects.md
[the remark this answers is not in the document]
technosophos: See #560

brig/cmd/brig/commands/project_create.go
```diff
@@ -0,0 +1,550 @@
...
+var (
+	projectCreateConfig      string
+	projectCreateOut         string
+	projectCreateFromProject string
+	projectCreateDryRun      bool
+	projectCreateNoPrompts   bool
+	projectCreateReplace     bool
+)
+
+var defaultProject = &brigade.Project{
```
adamreese: It might be safer to make this a non-pointer.  Force a copy of the object so the default stays default.

brig/cmd/brig/commands/project_create.go
```diff
@@ -0,0 +1,550 @@
...
+				},
+			},
+		}, &p.Kubernetes); err != nil {
+			return fmt.Errorf(abort, err)
+		}
+	}
+
+	configureWorker := false
+	if err := survey.AskOne(&survey.Confirm{
+		Message: "Configure Brigade worker?",
```
adamreese: Should this be part of the `Advanced Brigade Settings`?

technosophos:
🔥

PR #539, opened 2018-07-09
Add docs on brigade.json

### LINKED ISSUE #497: re-write docs around adding dependencies to a brigade worker
Now with #494 merged into master, it would be awesome to document this new alternative workflow, including the syntax around brigade.json, how it interacts with brigade.js etc.

radu-matei:
This PR adds documentation around using a `brigade.json` file to add custom dependencies to a Brigade worker.

closes #497

### DIFF
--- a/docs/topics/workers.md
+++ b/docs/topics/workers.md
@@ -1,17 +1,116 @@
-# Building Custom Workers
+# Adding custom libraries to a Brigade worker
 
 A worker is the component in Brigade that executes a `brigade.js`. Brigade ships
 with a general purpose worker focused on running jobs. This generic worker
 exposes a host of Node.js libraries to `brigade.js` files, as well as the
 `brigadier` Brigade library.
 
-Sometimes, though, it is desirable to add additional libraries -- perhaps even
-custom libraries -- to your workers. This guide walks through the process of
-adding custom libraries.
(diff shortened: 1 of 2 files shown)


docs/topics/workers.md
```diff
@@ -1,18 +1,117 @@
...
+[brigade:app] after: default event handler fired
+[brigade:app] beforeExit(2): destroying storage
+[brigade:k8s] Destroying PVC named brigade-worker-01cexrvrs08shcev26961cwd6n
+```
+
+Notes:
+
+- the dependencies should point the exact version (and not use the tilde ~ and caret ^ to indicate semver compatible versions)
+
+- note the already existing dependencies and dev dependencies for a Brigade worker:
```
bacongobbler: these dependencies will change over time as we update brigade-worker's package.json so this file may need to be constantly updated. Would it be better to link to https://github.com/Azure/brigade/blob/master/brigade-worker/package.json#L21-L38 instead?

radu-matei: Yeah, that's a great idea!
I will probably add some dependencies as examples in the doc though, but will link to the actual file from master.

bacongobbler: that sounds like a good compromise. 👍

Issue #540, opened 2018-07-09
Possible error in storage cache


technosophos:
I'm working on a feature to add `brig build list PROJECT_ID`, using `store.GetProjectBuilds()`. But I'm noticing that the first call to that API returns a different result than subsequent calls.

<img width="909" alt="screen shot 2018-07-09 at 16 35 36" src="https://user-images.githubusercontent.com/89193/42479497-5ee08280-8396-11e8-9a73-0598ec23b697.png">

I'm starting to hunt down the cause. But if anyone else has seen this, let me know. The relevant code is:

```go
		proj, err := store.GetProject(project_id)
		if err != nil {
			return err
		}

		fmt.Printf("getting builds for %s\n", proj.Name)

		bs, err = store.GetProjectBuilds(proj)
		if err != nil {
			return err
		}
```

technosophos:
Okay, it looks like the issue is that when Brig uses the storage, it does not give the cache time to sync. So adding a BlockUntilAPICacheSynced works.

technosophos:
Closed by #571

PR #541, opened 2018-07-09
 feat(brig): List builds by project (brig build list PROJECT)

technosophos:
This add `brig build list PROJECT_ID`.

It also modifies `brig project get` to return a Secret, which is the
default format we work with, rather than a one-off representation.

Depends on #534

### DIFF
--- a/brig/cmd/brig/commands/build_list.go
+++ b/brig/cmd/brig/commands/build_list.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"fmt"
 	"io"
+	"time"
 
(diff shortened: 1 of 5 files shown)


technosophos:
Also, this can be used to reproduce #540 reliably.

technosophos:
I added the fix for #540, which turns out to be a change to Brig.

technosophos:
Sorry, totally mis-titled this issue.

technosophos:
Needs rebase after #534 is merged

PR #544, opened 2018-07-10
Make service account creation conditional

imduffy15:
In some situations a user may want to create the service accounts outside
of the helm process.

Currently, this isn't possible with brigade as brigade will always attempt
to create the service accounts.

This update introduces new values to conditionally enable/disable the
creation of the service accounts for worker nodes and the vacuum.

Signed-off-by: Ian Duffy <ian@ianduffy.ie>

### DIFF
--- a/charts/brigade/Chart.yaml
+++ b/charts/brigade/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 description: Brigade provides event-driven scripting of Kubernetes pipelines.
 name: brigade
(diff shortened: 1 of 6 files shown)


imduffy15:
Thanks for reviewing @technosophos. What needs to be done to get it merged?

technosophos:
I am just checking on the backward compatibility impact. I think we're okay, though. I'll just have to document this in the release notes.

PR #545, opened 2018-07-10
feat(brig): Add AGE to brig build list

technosophos:
This adds the AGE header with fuzzy durations to the `brig build list`
command.


adamreese:
LGTM after rebase

PR #546, opened 2018-07-10
fix(brigade.js): add build ID for debugging broken builds.

technosophos:
This adds the build ID to the GitHub status message so that we can debug
builds more effectively.


technosophos:
For background, this is to help troubleshoot the brigade builds that are failing to complete on a flaky (and very old) cluster.

PR #552, opened 2018-07-16
use select{} instead of for{}

bacongobbler:
for{} causes the main loop to constantly claim clock cycles, whereas
select{} causes the goroutine to block and puts it to sleep, which is
the intention.


bacongobbler:
thanks @radu-matei for testing. :+1:

bacongobbler:
also manually tested and this works for myself as well. Thanks for identifying the bug!

PR #553, opened 2018-07-16
run go fmt on dashboard.go

bacongobbler:


brig/cmd/brig/commands/dashboard.go
```diff
@@ -67,8 +67,7 @@ var dashboard = &cobra.Command{
 		}
 
 		// block until the user sends a CTRL+C
-		select {
-		}
+		select {}
```
radu-matei: This will be introduced by #551 (rebased and ran the linter).

PR #568, opened 2018-07-23
Adding project defaultSecret to the build event secret

### LINKED ISSUE #567: project defaultScript is not getting added in the event secret
Data mounted from event secret is missing the `script` key.
As a result of this if `/vcs/brigade.js` and `/etc/brigade-default-script/brigade.js` is missing and the users are depending on the project's defaultScript things will break.

The fix for this is to add `Script` key in here
https://github.com/Azure/brigade/blob/master/pkg/webhook/github.go#L229

I will raise a PR for the same.

kumudt:
Fixes #567

### DIFF
--- a/pkg/webhook/github.go
+++ b/pkg/webhook/github.go
@@ -227,6 +227,7 @@ func (s *githubHook) build(eventType string, rev brigade.Revision, payload []byt
 		Provider:  "github",
 		Revision:  &rev,
 		Payload:   payload,
+		Script: []byte(proj.DefaultScript),
 	}
 
 	return s.store.CreateBuild(b)


kumudt:
@technosophos Can you please have a look at this?

technosophos:
I need to look at this carefully, but I think it should be the Controller that is loading the default script if no script (or an empty script) is passed in the event. In theory, the gateways should not have to do this work.

kumudt:
@technosophos You are right. Thinking about the right way to pass it to the worker. Can we do this?
Pass the default script via an environment variable (referencing project secret) to the worker and the worker loads it in prestart.js

mumoshu:
Can we modify the brigade-project chart to optionally create a configmap named per `defaultScriptName`(#467) with its content being the value of `defaultScript`? This way, perhaps, we don't need to touch any code in controller and worker.

technosophos:
At this point, we'd add that to `brig project create`, too. But, yeah, it would be pretty easy to add this.

radu-matei:
I think this is available in `brig project create`.

Feel free to open an issue if the behaviour is not the expected one.
Thanks!

PR #579, opened 2018-08-02
fix(vacuum): set default max builds to -1

### LINKED ISSUE #576: VACUUM_MAX_BUILDS variable not honoured
Latest brigade helm release
Docker image: deis/brigade-vacuum:v0.15.0

We noticed that VACUUM_MAX_BUILDS env variable is ignored by vacuum. Although VACUUM_AGE variable gets picked up as expected.

We tested both maxBuilds() and getAge() functions with both env variables against custom docker image and both return values.
(https://github.com/Azure/brigade/blob/master/brigade-vacuum/cmd/brigade-vacuum/commands/app.go)

We suspect the issue around deis/brigade-vacuum:v0.15.0 image.

Please let us know if you need further information.

Thanks

technosophos:
There was a discrepancy between whether the default value for max builds was -1 or 0. As a result, `VACUUM_MAX_BUILDS` was always overridden by the default `-m` value.

Closes #576

### DIFF
--- a/brigade-vacuum/cmd/brigade-vacuum/commands/app.go
+++ b/brigade-vacuum/cmd/brigade-vacuum/commands/app.go
@@ -52,14 +52,14 @@ var (
 	globalNamespace  = ""
 	globalAge        = ""
 	globalVerbose    = false
-	globalMaxBuilds  = -1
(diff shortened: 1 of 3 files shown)


technosophos:
Ugh... I forgot to update the test. I'll fix that and re-push.

PR #585, opened 2018-08-08
WIP ref(worker): switch to @azure/brigadier

technosophos:
This refactors the Brigade internal library to use @Azure/brigadier. This contains breaking changes in the pathing. (See https://www.npmjs.com/package/@azure/brigadier)

This also intercepts and overrides `require("@azure/brigadier")` for brigade.js files.


technosophos:
Still todo:

- [ ] Manual testing
- [ ] Publish Azure/brigadier GitHub project

brigade-worker/src/app.ts
```diff
@@ -1,11 +1,11 @@
 /**
  * Module app is the main application runner.
  */
-import * as events from "./events";
+import * as events from "@azure/brigadier/out/events";
```
radu-matei: nit: I am mildly annoyed by the fact that `/out/` should be included every time the package is used - I suspect there's a configuration property that can help us avoid this - I'll look into it tomorrow morning.

radu-matei: Ok, mildly was an understatement.

brigade-worker/src/k8s.ts
```diff
@@ -238,8 +238,12 @@ export class JobRunner implements jobs.JobRunner {
   pod: kubernetes.V1Pod;
   cancel: boolean;
   reconnect: boolean;
+  
+  constructor() {}
 
-  constructor(job: jobs.Job, e: BrigadeEvent, project: Project) {
+  public init<T extends jobs.Job>(job: T, e: BrigadeEvent, project: Project) {
+    // init takes a generic so we can run this against mocks as well as against the real
```
radu-matei: nit: can the function comment be before the function definition?

brigade-worker/src/app.ts
radu-matei: Update: found a solution for pretty import paths - not ideal, but will PR against Brigadier.

brigade-worker/src/k8s.ts
technosophos: fixed

Issue #485, opened 2018-05-29
`brig rerun` to reuse or renew the script in the target build


mumoshu:
Hi @technosophos and @adamreese,

Can you let me write my thoughts on `brig rerun` and help shaping supported use-cases of the command, suggest an additional option to it? :)

TL;DR; I'd like to add `brig rerun --renew-script`.

---

Ideally, I'd expect `brig rerun` to have two flavors: `brig rerun --reuse-script` or `brig rerun --renew-script`. 

The former is intended to retry a job that failed due to transient errors like network. We don't necessarily use the latest script at the time of the retry in this case. But using the latest on retry would also be acceptable.

The latter is to retry a job that failed permanently due to bugs in a script. I believe that you want to consume the same event(e.g. a github deployment event) with a new script w/ bug fixes in this case, hence `--renew-script` to use the latest script.

What do we actually have today? After #467, `brig rerun` seems to work like the former as long as the build contained a script. And it works like the latter if the build misses the script. So, it seems like missing the latter use-case. We have no way to retry a pipeline on a exact same event but with a latest script with fixes.

So, how about adding the `brig rerun --renew-script` one, to retry the failed build with emptying the script contained it? Then, brigade could use the latest script rather than the one contained in the failed build. To be extra clear, the latest script would be just populated from one of available sources according to #467.

technosophos:
Yeah, I agree with the intent of this. For consistency, would it be easier for users if we added a flag to explicitly override thescript, and called this flag `-f, --file` like it is in `brig run`? Or maybe we could just make `-f` the short flag for `--renew-script`?

I think that for the case where the script is fetched out of the repo, in most cases, the same exact version will be checked out if the SHA was set in the original event. But you are right that if only a ref was set, we currently cannot guarantee an identical rerun.

mumoshu:
> in most cases, the same exact version will be checked out if the SHA was set in the original event

Thanks for clarifying that. I somehow missed it in my original statement!

So, I now believe that my concern is mostly addressed by just adding `-f --file` to `brig rerun`. I'd prefer `--file` for the longer variant because it follows the same flag that exists in `brig run`.

mumoshu:
Resolved by #588.
I meant to tackle this myself, but thanksfully, @technosophos picked it up 🙏

PR #588, opened 2018-08-11
feat(brig): add -f flag to rerun command

technosophos:
This provides a way to rerun a payload, but replace the script file.
This ought to greatly improve the debugging experience when a script
fails.


mumoshu:
This resolves #485

technosophos:
Sorry that took me so long.

PR #591, opened 2018-08-13
fix(brig): set a default sidecar

technosophos:
In the course of debugging this, I noticed a case in the worker where the error data was getting lost.

Signed-off-by: Matt Butcher <matt.butcher@microsoft.com>


technosophos:
Hrm... I cannot reproduce that error. I'll have to see if it's a flake

technosophos:
Alright, it was an issue with needing to rebase. Seems to be fixed at this point.

PR #593, opened 2018-08-14
add --verbose flag

bacongobbler:
this feature flag enables more verbose logging of incoming http requests to the brigade API.


technosophos:
The tests passed. The error was a transient error with the pod log that caused the update notification to fail.

Issue #592, opened 2018-08-14
Suggest to make brigade-worker watch instead of polling for job status


cjhpaul:
Hi,

I did some stress test to see how my cluster (AKS) would handle when there are a large number of brigade events (or brigade workers).
Turned out that my cluster couldn't handle even when 10 brigade workers were instantiated and were polling for their jobs at the same time.
Btw, each brigade worker has 2 jobs that are not compute-intensive and usually take less than 10 seconds to complete.
The errors from the brigade workers were 504 from readNamespacedPod in k8s.ts.
Since it's one of transient faults, I could either increase capacities on the server side (possibly kube-apiserver) or make client (i.e., brigade-worker) more efficient.
Setting aside the server side trouble shooting (though I don't have full access to master node on AKS), I tried to make polling more efficient.
Essentially, when I made brigade worker use watch instead of polling (I also had to update @kubernetes/client-node to 0.5.2, or whichever the latest is), it significantly improved in terms of the number of brigade workers to run at the same time on my cluster.
So I just wanted to suggest to use watch instead for brigade worker.
In case interested, here is some code snippet for k8s.ts (note that I tried to minimize changing the existing code as much as possible. Certainly it's possible to make brigade-worker more efficient with watch I think):

```
  private startWatchingJob(): any {
    return kwatch.watch(`/api/v1/namespaces/${this.project.kubernetes.namespace}/pods`,
      { "labelSelector": `build=${this.event.buildID},jobname=${this.job.name}` },
      (_type, obj) => {
        this.pod = obj
      },
      (err) => {
        if (err) {
          this.logger.error("failed pod lookup");
          this.logger.error(err);
          this.cancel = true;
        }
      });
  }

  /** wait listens for the running job to complete.*/
  public wait(): Promise<jobs.Result> {
    // Should probably protect against the case where start() was not called
    let k = this.client;
    let timeout = this.job.timeout || 60000;
    let name = this.name;
    let ns = this.project.kubernetes.namespace;
    this.cancel = false;

    // This is a handle to clear the setTimeout when the promise is fulfilled.
    let waiter;

    this.logger.log(`Timeout set at ${timeout}`);
    let jobWatch = undefined;

    // At intervals, poll the Kubernetes server and get the pod phase. If the
    // phase is Succeeded or Failed, bail out. Otherwise, keep polling.
    //
    // The timeout sets an upper limit, and if that limit is reached, the
    // polling will be stopped.
    //
[...]

PR #596, opened 2018-08-16
use watch in worker for efficiency #592

cjhpaul:
regarding #592

### DIFF
--- a/brigade-worker/src/k8s.ts
+++ b/brigade-worker/src/k8s.ts
@@ -8,6 +8,8 @@ import * as kubernetes from "@kubernetes/client-node";
 import * as jobs from "./job";
 import { LogLevel, ContextLogger } from "./logger";
 import { BrigadeEvent, Project } from "./events";
+import * as fs from "fs";
+import * as path from "path";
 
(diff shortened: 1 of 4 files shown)


brigade-worker/src/k8s.ts
```diff
@@ -20,6 +22,18 @@ import { BrigadeEvent, Project } from "./events";
 const expiresInMSec = 1000 * 60 * 60 * 24 * 30;
 
 const defaultClient = kubernetes.Config.defaultClient();
+const getWatchClient = (): kubernetes.Watch => {
+  const kc = new kubernetes.KubeConfig();
+  const config = path.join(process.env.HOME, ".kube", "config");
```
technosophos: This was my biggest complaint about the watch API that got added. I don't like that we have to supply it with a config different from the one `defaultClient` loads. In my case, I sometimes override the config via `$KUBECONFIG`, so we might need to add handling for that. But that actually only impacts local dev (which I also don't do very often). So I am going to LGTM as-is.

cjhpaul:
It’d be appreciated if anyone could give an update on this.

PR #599, opened 2018-08-17
Re-use globalNamespace flag for brig dashboard

### LINKED ISSUE #598: brig dashboard not working
`brig dashboard` was working until I enabled rbac on brigade. Once that happened I see...

```
2018/08/16 16:35:34 Connecting to kashti at http://localhost:8081...
Error: cannot start port forward for kashti: could not find server
```

I dunno if this is an rbac issue. Brigade and Kashti both run in the same namespace.

But, If I run `k -n brigade port-forward svc/kashti-kashti 8080:80` it works just fine.

radu-matei:
This PR removes the `kashti-namespace` provided for `brig dashboard` and re-uses the global `--namespace` flag for `brig` - that is because `brig dashboard` only creates a tunnel to the Kashti pod and doesn't need to forward the port of the Brigade API.

While this removes the confusion brought by having two flags for the namespace, it re-uses the global namespace flag that in the rest of the CLI is used to indicate the namespace where Brigade is deployed, to indicate the namespace where Kashti is deployed.

Considering that we want to [include Kashti in the Brigade chart](https://github.com/Azure/brigade/issues/504#issue-331321180), having the two deployed in separate namespaces might disappear in the past.

cc @mattfarina @technosophos 

closes #598

### DIFF
--- a/brig/cmd/brig/commands/dashboard.go
+++ b/brig/cmd/brig/commands/dashboard.go
@@ -33,7 +33,6 @@ func init() {
 
 	flags := dashboard.PersistentFlags()
 	flags.IntVar(&port, "port", 8081, "local port for the Kashti dashboard")
-	flags.StringVar(&kashtiNamespace, "kashti-namespace", "default", "namespace for Kashti")
 	flags.BoolVar(&openDashboard, "open-dashboard", true, "open the dashboard in the browser")
 }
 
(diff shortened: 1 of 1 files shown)


radu-matei:
Kicking off another build (and solving a `gometalinter` issue in the meantime) - the failing tests do not appear to be related to the changes introduced by this PR though.

PR #603, opened 2018-08-21
vacuum skips deleting currently running builds if skipRunningBuilds is true

cjhpaul:
vacuum skips deleting currently running builds if skipRunningBuilds is true (false by default).
Previously, vacuum deletes all builds regardless of their phase even if running or pending.

### DIFF
--- a/brigade-vacuum/cmd/brigade-vacuum/commands/app.go
+++ b/brigade-vacuum/cmd/brigade-vacuum/commands/app.go
@@ -15,10 +15,11 @@ import (
 )
 
 const (
-	envKubeConfig = "KUBECONFIG"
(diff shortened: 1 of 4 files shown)


cjhpaul:
Just to add some context, this is regarding #604
In this commit, I added a helm config param for vacuum to skip deleting builds in pending/running phase.
To enable this feature, you would just add "skipRunningBuilds: true" under vacuum in values.yaml. Otherwise, vacuum would behave as it used to (i.e., deleting all builds) by default.
I updated vacuum_test.go as well to cover this use case when the param is set. All vacuum tests are passing.
It'd be appreciated if anyone could review this.

PR #606, opened 2018-08-22
feat(github-gateway): add endpoint for creating github webhook

adamreese:


radu-matei:
Ping @adamreese - what's the status of this PR? Since we plan on moving to the new Checks API app, is this change still actual?

Thanks!

PR #617, opened 2018-08-31
fix: add defaultScriptName to chart

technosophos:


technosophos:
#618 includes a fix for this, so I am closing.

PR #619, opened 2018-09-03
fix BuildStorage in brigade worker to prevent crash when project load fails

mansoorcheema:
Build storage class initializes its variables in BuildStorage.create() after loading project but when project does not exists or there is error in loading project, build storage is not created so variables are left uninitialized e.g (logger). Later when error is handled in app.ts , it tries to destroy build storage. In this case the storage wasn't initialized in the first place which terminates worker in the middle of exception. 

capture


technosophos:
Merged. Thank you for tracking that down and fixing it!

Issue #611, opened 2018-08-29
TypeError in brigade


mattfarina:
I got the following error..

```sh
± b build logs 01cp3at17bpwnkars7yng7dh74

==========[  brigade-worker-01cp3at17bpwnkars7yng7dh74  ]==========
prestart: no dependencies file found
prestart: no script override
TypeError: Cannot read property 'toString' of undefined
error Command failed with exit code 1.
```

_note, `b` is an alias to `brig` in the namespace I use_

Notice, this came before the message I usually see of:

```
[brigade] brigade-worker version: 0.17.0
```

This happened for the charts issue at https://github.com/helm/charts/pull/7424. Is this because `brigade.js` didn't exist on the PR branch? If so, it would be nice if there was an easier to follow message so I could easily pick this up.

mattfarina:
I found the source of the TypeError. 

https://github.com/Azure/brigade/blob/c5907e734358eaec08cf4cb83a9d059c1c622647/brigade-worker/prestart.js#L31-L34

It's the `e.toString()` call here. Happens when there is no brigade.js or default file.

bacongobbler:
fixed via #621

PR #621, opened 2018-09-06
fix(prestart): don't call .toString() on an error

bacongobbler:


michelleN:
resolves #611

bacongobbler:
any time :)

PR #630, opened 2018-09-13
fix(doc): broken code block in the topics/developers doc

mumoshu:
This should fix the broken code block in https://azure.github.io/brigade/topics/developers.html:

image


mumoshu:
Thanks for the review ☺️

PR #650, opened 2018-10-03
fix(charts): fix incorrect logic for default sidecar

adamreese:


adamreese:
good catch

elvisace:
What if I _don't_ want to use a vcs sidecar?

PR #655, opened 2018-10-09
fix(*): address linting errors, add test-style to brigade build task

vdice:


vdice:
Oddly, CI still hits a few errors that I'd already fixed here in this PR -- but perhaps via disparate `gometalinter` & co. tools... need to debug a bit...

vdice:
Ah ha! Linting errors seen/claimed previously in CI but not locally seem to be due to use of `golang:1.10` for the go build/test pod, whereas I have golang 1.11 on my box.  When I bump image, linting passes.  PTAL @technosophos

PR #656, opened 2018-10-10
WIP feat(*): switch to Debian Jessie-Slim

technosophos:
This switches the base image to Debian Jessie as an experiment to work around Alpine DNS issues.


technosophos:
DO NOT MERGE. This is largely experimental at this point, and it comes at the cost of adding 50M per image.

PR #658, opened 2018-10-11
Fix brig run file flag

kooba:
Fix `brig run` file flag. I believe `-e` flag is for providing event name


vdice:
Ah, indeed, thanks for the docs fix @kooba !

Issue #469, opened 2018-05-22
If a Job.env{} var is empty, it will cause a Kubernetes error


technosophos:
Consider this:

```javascript
j = new Job("foo", "alpine:3.7")
j.env = {
  foo: null
}
```

The above will create a Kubernetes pod with this in the env section:

```json
          {
            "name": "foo",
            "valueFrom": {}
          }
```

The above is invalid. It should be:

```json
          {
            "name": "foo",
            "value": ""
          }
```

Or something else that is structurally allowed.

The code that generates this is in the worker's `k8s.ts` file somewhere.

jdotjdot:
This one's my fault.  It's these lines:

https://github.com/Azure/brigade/blob/a2646de88d941a4921b0cf805d108fe4a2f6f8f1/brigade-worker/src/k8s.ts#L216-L243

This could be fixed by adding another part to the if-statement at line 234, something like:

```javascript 
} else if (val === null) {
   envVars.push({
     name: key,
     valueFrom: {
       secretKeyRef: {
         name: secName,
         key: ""
       }
     }
  } as kubernetes.V1EnvVar);
} else {
// ...
```

shubheksha:
I can pick this up!

technosophos:
@shubheksha excellent! Thanks!

shubheksha:
@biskit1, I'm on the move for a bit, so I won't be able to, you can pick it up!

biskit1:
Thanks @shubheksha! Will try to get this done over the weekend. Will keep you posted!

biskit1:
Just to update anyone who comes around: I'm trying to figure out how to set up the dev environment so I can run/implement the appropriate tests once I make the small change suggested above. 

I had to prioritize some other things given the time I estimated setting this up/figuring it out would take, so I just want you to know that I'm still on it and will complete it ASAP. 

If anyone (@jdotjdot , @technosophos) has some suggestions for setting up the dev environment just for this portion, please share :) for now, I'm getting started by following the steps listed in this [developers guide](https://azure.github.io/brigade/topics/developers.html) to get the appropriate tools installed and set up the environment.

dgkanatsios:
closing since #670 has been merged

PR #670, opened 2018-10-18
Kubernetes error caused by Job.env{} empty var

biskit1:
Fixes issue #469. 

Added a check for `null` which will set `value` in Kubernetes pod env section to an empty string, instead of the default behavior indicated in issue #469 which was invalid and would cause an error. 

Please note I had some problems getting the dev environment set up, and haven't been able to test this locally as of yet.


biskit1:
Thanks @vdice! And I appreciate you outlining the steps you took for testing.

PR #672, opened 2018-10-19
ref(*): update brigade image repos

vdice:
Use/supply MCR-hosted images as default


radu-matei:
@vdice you removed the awaiting review label, is this something we still want to merge?

PR #673, opened 2018-10-19
ref(brigade.js): add dockerhub publish job, split up acrBuild

vdice:
This refactors the `brigade.js` file to add publishing the "classic" `deis/brig*` images to Dockerhub for the time being while we transition to using Microsoft's Container Registry (see also https://github.com/Azure/brigade/pull/672/files).


vdice:
Refreshed/updated this branch and tested via rerunning a recent `master` build via `brig`, supplying the `brigade.js` file on this branch.  Looking good.  (Dockerhub and ACR/MCR `latest` images updated)

PR #683, opened 2018-10-27
Added the length check in brigadier for job name

### LINKED ISSUE #679: Request: validate job name length
Kubernetes imposes a 63 character limit on pod names. If I create a job who's name exceeds this limit k8s sends back a failure about the issue but brigade ignores it and repeatedly calls k8s to create the pod with the bad name. This just fills up the log with unnecessary k8s request/responses.

# Add code in brigadier to validate a job's name is less than 37 characters and throw a javascript error if it fails. (-26 for the hash brigade already takes up with the unique build id)

codegagan:
fixes #679

### DIFF
--- a/brigade-worker/src/brigadier.ts
+++ b/brigade-worker/src/brigadier.ts
@@ -52,6 +52,10 @@ export class Job extends jobImpl.Job {
   jr: JobRunner;
 
   run(): Promise<jobImpl.Result> {
+    if (this.name.length > 36) {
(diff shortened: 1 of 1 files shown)


radu-matei:
So if I understand correctly, can we close this and keep #684?

Thanks for your contribution!

codegagan:
@radu-matei yes, unless someone wants the fix this way.

PR #684, opened 2018-10-27
Added job name length check in job constructor

### LINKED ISSUE #679: Request: validate job name length

codegagan:
Job seems to be correct place to implement name check instead of
brigadier. And there are other checks on job name.

fixes #679

### DIFF
--- a/brigade-worker/src/job.ts
+++ b/brigade-worker/src/job.ts
@@ -243,7 +243,7 @@ export abstract class Job {
   ) {
     if (!jobNameIsValid(name)) {
       throw new Error(
-        "job name must be letters, numbers, and '-', and must not start or end with '-'"
+        "job name must be letters, numbers, and '-', and must not start or end with '-', having max length 36"
       );
     }
     this.name = name.toLocaleLowerCase();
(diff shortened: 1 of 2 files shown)


codegagan:
@vdice Good suggestion.

I think either of this PR or #683 should be taken and not both. I personally prefer this one. But It was specifically asked to put the check in brigadier in issue #679 so I implemented both ways.

Please let me know, so I will make changes (jobNameLength constant) in respective PR.

vdice:
@codegagan I agree with you; this PR seems the way to go (especially w/ convenient test addition!)  Unless other interested parties speak up, let's go this route.

codegagan:
Did the suggested changes, kindly have a look.

brigade-worker/src/job.ts
```diff
@@ -150,6 +150,8 @@ export class JobResourceRequest {
  * the Docker container to be run.
  * */
 export abstract class Job {
+  public static readonly MAX_JOB_NAME_LENGTH = 36;
```
radu-matei: Any particular reason this is not an exported constant?

Otherwise, this LGTM 👍

codegagan: As the class is getting exported and this value very much relates to Job (than being independent). The value can always be accessed outside the class (being public) like `Job.MAX_JOB_NAME_LENGTH`

PR #693, opened 2018-11-20
Add resource limits to job

slok:
Now we can add also resource limits to the jobs:

```js
job.resourceRequests.memory = "3Gi";
job.resourceRequests.cpu = "1";
job.resourceLimits.memory = "6Gi";
job.resourceLimits.cpu = "3";
```

Also fixes the issue of only setting request resources on the container pod specif you set CPU **and** memory. Now if any of those two exist it sets the ones described.

Updated the docs to show how to add resources to the jobs

### DIFF
--- a/docs/topics/javascript.md
+++ b/docs/topics/javascript.md
@@ -22,7 +22,6 @@ Brigade files respond to _events_. That is, Brigade scripts are typically compos
 more _event handlers_. When the Brigade environment triggers an event, the associated
 event handler will be called.
 
-
 ## The `brigadier` Library
 
 The main library for Brigade is called `brigadier`. The Brigade runtime grants access to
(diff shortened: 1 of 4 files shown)


docs/topics/javascript.md
```diff
@@ -197,7 +196,24 @@ Properties of `Job`
 - `storage: JobStorage`: Preferences for the way this job attaches to the build storage
 - `docker: JobDockerMount`: Preferences for mounting a Docker socket
 - `serviceAccount: string`: The name of the service account to use (if you need to override the default).
-- `annotations: {[key: string]:string}`: Name/value pairs of annotations to add to the job's pod
+- `resourceRequests: JobResourceRequest`: CPU and memory request resources for the job pod container.
+- `annotations: JobResourceLimit`: CPU and memory limit resources for the job pod container.
```
vdice: The key here should be `resourceLimits`, correct?

Also, it looks like the `annotations: {[key: string]:string}: Name/value pairs of annotations to add to the job's pod` line may have inadvertently been removed?

slok: Yes! My bad, when adding the changes with git. Sorry

vdice: No worries!  Thanks for the feature and documentation!

Issue #586, opened 2018-08-09
Make it easy to remove old builds manually


emmayang:
Hi, 

I have a question about if it is possible to remove old builds, since during the development phase, I created about 50+ builds, and all of them are shown in Kashti dashboard. I was thinking there might be maximum number of builds setting, and latest build will replace oldest ones if exceeds the maximum number. However I found the list keeps growing, and I cannot find a way to clean them. (I tried brig cli v0.16.0, and still 3 available commands under `brig build`)

And Due to Kashti limitation, it's not sorted correctly, and I have to scroll down a really long list to find my latest build. 

I even tried delete the project, and recreat it, but still found previous builds are attached to it again. 

Sorry if this is an obvious question and something I missed in the documentation.

technosophos:
You can run a `brigade-vacuum` run in your cluster manually by duplicating an older run of `brigade-vacuum` and modifying the env vars. I wonder if that would be a useful feature to add to `brig` itself.

emmayang:
@technosophos thanks, I think it would be very nice and handy to have in brig command line, or even better in Kashti UI.

technosophos:
Re-marking this as a feature request. Changed the title accordingly

vdice:
I could see this as a handy brig command.  Perhaps ‘brig build delete <build id> <project>’ and, perhaps more useful, ‘brig build delete-all <project>’ or some such, to delete all (probably non-running) builds for a project.  

Perhaps the third enhancement as mentioned above would be to delete all builds when a project is deleted, since, at least at the time this was written, it doesn’t appear to do so successfully.

dgkanatsios:
So,
1. `brig build delete [buildID]` should delete corresponding Secret and Pod 
2. `brig build delete-all [project]` should delete all Secrets and Pods for [project] that are Succeeded

On (1), I wouldn't include [project] since the relevant command `brig build get [buildID]` does not include [project]. Thoughts? I'd like to give this a shot. cc @radu-matei

radu-matei:
That would be fantastic, @dgkanatsios!

- I think `brig build delete [buildID]` is sufficient without the project ID as parameter, as it is consistent with  in `brig build get/list`
 
- `brig build delete-all [projectID]` would be fantastic.

Thanks for picking this up, let us know if we can help in any way!

vdice:
Closing as https://github.com/Azure/brigade/pull/695 as been merged 🎉 !

PR #695, opened 2018-11-27
brig build delete BuildID and brig build delete-all ProjectID

dgkanatsios:
Implements #586 

Notes:
- `brig build delete` will delete the BuildID regardless of its state
- `brig build delete-all` will delete only non-running Builds from Project
- there was some code in vacuum about Build deletion - moved that to `pkg/storage/kube` so it can be used from both vacuum and brig

many thanks to @radu-matei for guidance and encouragement!

### DIFF
--- a/brig/cmd/brig/commands/build_delete.go
+++ b/brig/cmd/brig/commands/build_delete.go
@@ -10,7 +10,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const buildDeleteUsage = `Deletes a build.`
+const buildDeleteUsage = `Deletes a build and its corresponding jobs.`
 
 var forceDeleteRunning bool
 
--- a/brig/cmd/brig/commands/build_delete.go
+++ b/brig/cmd/brig/commands/build_delete.go
@@ -1,0 +1,42 @@
+package commands
+
+import (
+	"errors"
+	"io"
+
+	"github.com/Azure/brigade/pkg/storage"
+	"github.com/Azure/brigade/pkg/storage/kube"
+
+	"github.com/spf13/cobra"
+)
+
+const buildDeleteUsage = `Deletes a build, regardless of its state.`
+
+func init() {
+	build.AddCommand(buildDelete)
+}
+
+var buildDelete = &cobra.Command{
+	Use:   "delete BUILD_ID",
+	Short: "deletes build",
+	Long:  buildDeleteUsage,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) == 0 {
+			return errors.New("build ID is a required argument")
+		}
+		return deleteBuild(cmd.OutOrStdout(), args[0])
+	},
+}
+
+func deleteBuild(out io.Writer, bid string) error {
+	c, err := kubeClient()
+	if err != nil {
+		return err
+	}
+
+	store := kube.New(c, globalNamespace)
+	err = store.DeleteBuild(bid, storage.DeleteBuildOptions{
+		SkipRunningBuilds: false,
+	})
+	return err
+}
--- a/brig/cmd/brig/commands/build_delete.go
+++ b/brig/cmd/brig/commands/build_delete.go
@@ -23,6 +23,7 @@ var buildDelete = &cobra.Command{
 	Use:   "delete BUILD_ID",
 	Short: "deletes build",
 	Long:  buildDeleteUsage,
+	Args:  cobra.ExactArgs(1)
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if len(args) == 0 {
 			return errors.New("build ID is a required argument")
--- a/brig/cmd/brig/commands/build_delete.go
+++ b/brig/cmd/brig/commands/build_delete.go
@@ -10,10 +10,13 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const buildDeleteUsage = `Deletes a build, regardless of its state.`
+const buildDeleteUsage = `Deletes a build.`
+
+var forceDeleteRunning bool
 
 func init() {
 	build.AddCommand(buildDelete)
+	buildDelete.Flags().BoolVar(&forceDeleteRunning, "force", false, "If set, will also delete running builds. Default: false")
 }
 
 var buildDelete = &cobra.Command{
@@ -35,8 +38,7 @@ func deleteBuild(out io.Writer, bid string) error {
 	}
 
 	store := kube.New(c, globalNamespace)
-	err = store.DeleteBuild(bid, storage.DeleteBuildOptions{
-		SkipRunningBuilds: false,
+	return store.DeleteBuild(bid, storage.DeleteBuildOptions{
+		SkipRunningBuilds: !forceDeleteRunning,
 	})
-	return err
 }
--- a/brig/cmd/brig/commands/build_delete.go
+++ b/brig/cmd/brig/commands/build_delete.go
@@ -23,7 +23,7 @@ var buildDelete = &cobra.Command{
 	Use:   "delete BUILD_ID",
 	Short: "deletes build",
 	Long:  buildDeleteUsage,
-	Args:  cobra.ExactArgs(1)
+	Args:  cobra.ExactArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if len(args) == 0 {
 			return errors.New("build ID is a required argument")
--- a/brig/cmd/brig/commands/build_delete_all.go
+++ b/brig/cmd/brig/commands/build_delete_all.go
@@ -10,15 +10,18 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const buildDeleteAllUsage = `Deletes all non-running builds for a project.`
+const buildDeleteAllUsage = `Deletes all builds for a project.`
+
+var forceDeleteRunningAll bool
 
 func init() {
 	build.AddCommand(buildDeleteAll)
+	buildDeleteAll.Flags().BoolVar(&forceDeleteRunningAll, "force", false, "If set, will also delete running builds. Default: false")
 }
 
 var buildDeleteAll = &cobra.Command{
 	Use:   "delete-all PROJECT_ID",
-	Short: "deletes all non-running builds for a project",
+	Short: "deletes all builds for a project",
 	Long:  buildDeleteAllUsage,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if len(args) == 0 {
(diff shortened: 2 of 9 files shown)


brig/cmd/brig/commands/build_delete.go
```diff
@@ -0,0 +1,42 @@
...
+func init() {
+	build.AddCommand(buildDelete)
+}
+
+var buildDelete = &cobra.Command{
+	Use:   "delete BUILD_ID",
+	Short: "deletes build",
+	Long:  buildDeleteUsage,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) == 0 {
```
radu-matei: For future reference, here you could also use the following for the `cobra.Command`:

```
		Args:  cobra.ExactArgs(1),
```

brig/cmd/brig/commands/build_delete.go
```diff
@@ -0,0 +1,42 @@
...
+
+func deleteBuild(out io.Writer, bid string) error {
+	c, err := kubeClient()
+	if err != nil {
+		return err
+	}
+
+	store := kube.New(c, globalNamespace)
+	err = store.DeleteBuild(bid, storage.DeleteBuildOptions{
+		SkipRunningBuilds: false,
```
radu-matei: An idea would be not to delete a running build by default, but provide a `--force` flag, so by default, both `delete` and `delete-all` have consistent behaviour when it comes to running builds.

brig/cmd/brig/commands/build_delete.go
```diff
@@ -0,0 +1,42 @@
...
+	})
+	return err
```
radu-matei: If `store.DeleteBuild` already returns an `error`, we could simply just `return store.DeleteBuild(...)`, if the result is not used again.

brig/cmd/brig/commands/build_delete_all.go
```diff
@@ -0,0 +1,63 @@
...
+
+	builds, err := store.GetProjectBuilds(proj)
+	if err != nil {
+		return err
+	}
+
+	var errDeletes error
+	for _, b := range builds {
+		err = store.DeleteBuild(b.ID, storage.DeleteBuildOptions{
+			SkipRunningBuilds: false,
```
radu-matei: The command description says that it `Deletes all non-running builds for a project` - however, it seems it would not skip running builds - should this be `true`?

pkg/storage/kube/build.go
```diff
@@ -34,7 +40,48 @@ func (s *store) GetBuild(id string) (*brigade.Build, error) {
...
+func (s *store) DeleteBuild(bid string, options storage.DeleteBuildOptions) error {
+	opts := meta.ListOptions{
+		LabelSelector: fmt.Sprintf(jobFilter, bid),
+	}
+	delOpts := meta.NewDeleteOptions(0)
+	pods, err := s.client.CoreV1().Pods(s.namespace).List(opts)
+	if err != nil {
+		return err
+	}
+	if options.SkipRunningBuilds {
```
vdice: I'm wondering if these lines 53-67 can be consolidated into one pod iteration loop?  Something like
```
for _, p := range pods.Items {
  if p.Labels["component"] == "build" {
    if options.SkipRunningBuilds && (p.Status.Phase == v1.PodRunning || p.Status.Phase == v1.PodPending) {
      return errors.New("skipping because build is still running")
    } else {
      log.Printf("Deleting pod %q", p.Name)
      if err := s.client.CoreV1().Pods(s.namespace).Delete(p.Name, delOpts); err != nil {
          log.Printf("failed to delete job pod %s (continuing): %s", p.Name, err)
      }
    }
  }
}
```

Or, if the above isn't quite right, just looking for some way to condense logic/iterations.  If it just doesn't work this way, no worries!

pkg/storage/kube/build.go
```diff
@@ -34,7 +40,48 @@ func (s *store) GetBuild(id string) (*brigade.Build, error) {
...
+				if p.Status.Phase == v1.PodRunning || p.Status.Phase == v1.PodPending {
+					return errors.New("skipping because build is still running")
+				}
+			}
+		}
+	}
+	for _, p := range pods.Items {
+		log.Printf("Deleting pod %q", p.Name)
+		if err := s.client.CoreV1().Pods(s.namespace).Delete(p.Name, delOpts); err != nil {
+			log.Printf("failed to delete job pod %s (continuing): %s", p.Name, err)
```
vdice: Should we return an error here?  And/or, I'm not clear if `(continuing)` implies we will be attempting deletion again if we hit this error?

pkg/storage/kube/build.go
```diff
@@ -34,7 +40,48 @@ func (s *store) GetBuild(id string) (*brigade.Build, error) {
...
+		}
+	}
+
+	secrets, err := s.client.CoreV1().Secrets(s.namespace).List(opts)
+	if err != nil {
+		return err
+	}
+	for _, sec := range secrets.Items {
```
vdice: If `skipRunningBuilds` is `true` or more generally, pods remain for a build after this is called, should we forestall build secret deletion?

brig/cmd/brig/commands/build_delete_all.go
```diff
@@ -0,0 +1,63 @@
...
+		})
+		if err != nil {
+			if errDeletes == nil {
+				errDeletes = errors.New("")
```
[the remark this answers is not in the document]
dgkanatsios: exactly!

brig/cmd/brig/commands/build_delete.go
dgkanatsios: that's a great idea, will do!

pkg/storage/kube/build.go
radu-matei: I think it continues ranging over `pods.Items`.

brig/cmd/brig/commands/build_delete.go
dgkanatsios: updated cobra and it works, thanks!

pkg/storage/kube/build.go
vdice: Created https://github.com/Azure/brigade/issues/696 to track

pkg/storage/kube/build.go
vdice: Created https://github.com/Azure/brigade/issues/696 to track

pkg/storage/kube/build.go
vdice: Created https://github.com/Azure/brigade/issues/696 to track

vdice:
@dgkanatsios getting around to testing this on some local projects.  Overall, looks great.  Though, noticing the following case.  I kicked off a run and then simultaneously invoked `build delete` to test the case of skipping running builds and I see:

```
 $ b build delete 01cxbjnw7t3tpkjv0mh36knrnq
Error: skipping because build is still running
Usage:
  brig build delete BUILD_ID [flags]

Flags:
      --force   If set, will also delete running builds. Default: false
  -h, --help    help for delete

Global Flags:
      --kube-context string   The name of the kubeconfig context to use.
      --kubeconfig string     The path to a KUBECONFIG file, overrides $KUBECONFIG.
  -n, --namespace string      The Kubernetes namespace for Brigade (default "default")
  -v, --verbose               Turn on verbose output

skipping because build is still running
```

So, it does properly error out, but it provides a few issues that I think we should fix up and/or discuss (fixes/amendments would probably apply in the `delete-all` case, too.)

1. Fix: prevent printing command usage if the error is due to skipping running builds
2. Thought:  Maybe this shouldn't be an error in the first case?  Perhaps just informational/warning and then exit zero/success, perhaps also mentioning possibility of supplying `--force` to force deletion.  (And then, if actual attempted deletion failed, we'd throw an error and/or exit non-zero.)
3. Thought: perhaps include build ID with each skipped build log/error output?

dgkanatsios:
I've been looking at the [code for the deleteBuild](https://github.com/dgkanatsios/brigade/blob/brig-delete/pkg/storage/kube/build.go#L44) method, maybe we can fix its behavior and also discuss #696 as well.
So, what deleteBuild does/should do is 
- if skipRunningBuilds == false, it should delete all build pods as well as the corresponding job pods
- if skipRunningBuilds == true then it should check if the build to be deleted is running (or pending)
  - if it is, then it should skip deleting  the build but *not return an error* as it does now. Maybe just log that it's running and return nil
  - if it is not running or pending, then it should delete the build pod as well as the corresponding job pods (as it does now)

Probably both loops are needed. In the first loop, we're searching for a single Pod. We could get this Pod by another label-filtered call to K8s API Server, but I think it's better to do an in-memory search. In the second, we do try to delete the corresponding job pods as well as the build pod. Another thing we could do on the Delete calls though, is maybe use a Retry logic [sample](https://github.com/dgkanatsios/azuregameserversscalingkubernetes/blob/f7c0fee4c39d4827375d831cdd30cf6ff62380a6/e2e/cmd/main.go#L284).

Anyway, just did a push and changed the first loop to this:

```golang
	if options.SkipRunningBuilds {
		for _, p := range pods.Items {
			if p.Labels["component"] == "build" {
				if p.Status.Phase == v1.PodRunning || p.Status.Phase == v1.PodPending {
					log.Printf("skipping Build %s because it is on Status %s", p.Labels["build"], p.Status.Phase)
					return nil
				}
			}
		}
	}
```

In short, if a Build is running, don't delete it, log it and don't return an error. 

Did a small test and seems OK, no errors:
```bash
dgkanatsios@digkanat:brig$ go run *.go build delete-all brigade-370f96797fcc5bebdfe76f919cf8c6e8b3c254b671889867ed08db
2018/11/28 11:21:31 skipping Build 01cxcsyxj7rp3gs1npdd9msnz7 because it is on Status Running
```

We also need to check vacuum, since it's depending on that method. There are 2 points in vacuum code that deleteBuild method is called. [point 1](https://github.com/dgkanatsios/brigade/blob/brig-delete/brigade-vacuum/cmd/brigade-vacuum/vacuum/vacuum.go#L72) and [point 2](https://github.com/dgkanatsios/brigade/blob/brig-delete/brigade-vacuum/cmd/brigade-vacuum/vacuum/vacuum.go#L105). Both of these snippets will *continue* in case of error, so the proposed change will not affect them. However I'd like a second pair of eyes on that.

vdice:
@dgkanatsios Thanks for digging in!  I think this PR is in a great state now.  The functionality feels right and I'm fine putting off more impactful deletion/vacuum refactors for a later date.  

Tested w/ all Brigade server images built from this branch (including vacuum) and it looks good.

cc @radu-matei

pkg/storage/kube/build.go
```diff
@@ -34,7 +39,49 @@ func (s *store) GetBuild(id string) (*brigade.Build, error) {
...
+	delOpts := meta.NewDeleteOptions(0)
+	pods, err := s.client.CoreV1().Pods(s.namespace).List(opts)
+	if err != nil {
+		return err
+	}
+	if options.SkipRunningBuilds {
+		for _, p := range pods.Items {
+			if p.Labels["component"] == "build" {
+				if p.Status.Phase == v1.PodRunning || p.Status.Phase == v1.PodPending {
+					log.Printf("skipping Build %s because it is on Status %s", p.Labels["build"], p.Status.Phase)
```
vdice: minor nit on wording/case here... perhaps `skipping build <build id> because its status is <status>`?

dgkanatsios: just added that, thanks!

Issue #698, opened 2018-11-29
brigade-vacuum/skipRunningBuilds unit test failing


vdice:
@dgkanatsios oops, missed this unit test failure before merging the delete build work into master.  Let me know if you have a moment to look; otherwise I can check it out a bit later:

```
2018/11/29 11:23:54 Pruning records older than 2018-11-29 11:23:54.695447 -0700 MST m=+0.008611744
2018/11/29 11:23:54 Deleting pod "queequeg"
2018/11/29 11:23:54 Deleting pod "tashtego"
2018/11/29 11:23:54 Deleting secret "queequeg"
2018/11/29 11:23:54 Deleting secret "tashtego"
2018/11/29 11:23:54 skipping Build 234567 because its Status is Running
--- FAIL: TestRun_SkipRunningBuilds (0.00s)
    /Users/vdice/go/src/github.com/Azure/brigade/brigade-vacuum/cmd/brigade-vacuum/vacuum/vacuum_test.go:97: expected 1 deletion, got 2
FAIL
FAIL	github.com/Azure/brigade/brigade-vacuum/cmd/brigade-vacuum/vacuum	0.029s
Error: Tests failed.
```

dgkanatsios:
Error/inconsistency is probably in the Run method of Vacuum, checking now.

dgkanatsios:
So, the problem lies probably here on Run method

```golang
if v.age.After(ts) {
	if err := v.deleteBuild(bid); err != nil {
		log.Printf("Failed to delete build %s: %s (age)\n", bid, err)
		continue
	}
	deleted++
}
```

Before the PR, if deleteBuild returned an error, this would mean that it did not delete the Build. Now, it just returns nil if it will skip the build. So there is no easy way to determine whether deleteBuild has actually deleted the build. 
Proposed solutions:
1. modify deleteBuild so that it also returns a bool that says whether it delete the Build
2. modify Run so it does not return a deleted integer but instead, on the test method, use the go-client to check how many Pods/Secrets are now in the cluster (and determine how many were deleted)

I'd go for (2), but I'm cool with both options. Thoughts?

vdice:
@dgkanatsios ah, that explains it.  Thanks for taking a look.  I could go either way but also leaning towards (2).

PR #699, opened 2018-11-30
Modify vacuum.Run to not return the number of deleted builds

dgkanatsios:
As per discussion on #698

- modified vacuum.Run to not return the number of deleted builds. #695  modified deleteBuild behavior and this led to test inconsistencies
- refactored tests to explicitly check for deleted/existent pods
- the only downside is that vacuum execution now does not report the number of deleted builds. However, relevant deletion messages are logged through deleteBuild, so I guess there's not much harm in that


vdice:
Closes https://github.com/Azure/brigade/issues/698

PR #704, opened 2018-12-04
Add resources to worker and jobs

### LINKED ISSUE #701: Resources in worker and job pods
As far as I know, there are no way of setting the container resources for both worker and jobs pods.

This issue is intended to discuss the way of implement this to create a PR with an agreed schema.

**Actual status**

| pod | container | status |
|---|---|---|
| worker | init-container | no resources |
| worker | worker container | no resources |
| job | init-container | no resources |
| job | job container | set by job configuration |

**Desired status**: ability to set resources in every container

**Proposal**

| pod | container | status |
|---|---|---|
| worker | init-container | set by project definition |
| worker | worker container | set by flags in the controller |
| job | init-container | set by project definition |
| job | job container | set by job configuration |

I'd create four flags for the worker configuration:
- `worker-limits-cpu`
- `worker-limits-memory`
- `worker-requests-cpu`
- `worker-requests-memory`

This would set the resources on the main container of the worker.

For the `vcs-sidecar` init-containers, I'd set the resources in the project configuration:
- `vcsSidecar.limits.cpu`
- `vcsSidecar.limits.memory`
- `vcsSidecar.requests.cpu`
- `vcsSidecar.requests.memory`

What do you think about this?

jchanam:
Implements #701 

Adds flags for the controller to set the worker container resources.

Adds project configuration to set the init-container vcs-sidecar resources.

### DIFF
--- a/brigade-controller/cmd/brigade-controller/main.go
+++ b/brigade-controller/cmd/brigade-controller/main.go
@@ -31,6 +31,10 @@ func main() {
 	flag.StringVar(&ctrConfig.WorkerImage, "worker-image", defaultWorkerImage(), "kubernetes worker image")
 	flag.StringVar(&ctrConfig.WorkerPullPolicy, "worker-pull-policy", defaultWorkerPullPolicy(), "kubernetes worker image pull policy")
 	flag.StringVar(&ctrConfig.WorkerServiceAccount, "worker-service-account", defaultWorkerServiceAccount(), "kubernetes worker service account name")
+	flag.StringVar(&ctrConfig.WorkerRequestsCPU, "worker-requests-cpu", "", "kubernetes worker cpu requests")
+	flag.StringVar(&ctrConfig.WorkerRequestsMemory, "worker-requests-memory", "", "kubernetes worker memory requests")
+	flag.StringVar(&ctrConfig.WorkerLimitsCPU, "worker-limits-cpu", "", "kubernetes worker cpu limits")
+	flag.StringVar(&ctrConfig.WorkerLimitsMemory, "worker-limits-memory", "", "kubernetes worker memory limits")
 	flag.Parse()
 
 	// creates the connection
--- a/charts/brigade-project/Chart.yaml
+++ b/charts/brigade-project/Chart.yaml
@@ -2,3 +2,4 @@ apiVersion: v1
 description: Create a Brigade project
 name: brigade-project
 version: 0.18.1
+appVersion: v0.18.0
--- a/charts/brigade-project/Chart.yaml
+++ b/charts/brigade-project/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: Create a Brigade project
 name: brigade-project
-version: 0.18.0
+version: 0.18.1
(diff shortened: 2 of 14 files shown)


charts/brigade-project/templates/secret.yaml
```diff
@@ -28,6 +28,10 @@ stringData:
   {{- else }}
   vcsSidecar: {{ printf "deis/git-sidecar:v%s" .Chart.Version }}
   {{- end }}
+  vcsSidecarResources.limits.cpu: {{ default "" .Values.vcsSidecarResources.limits.cpu | quote }}
```
vdice: I believe this section needs to be wrapped in a `  {{ if .Values.vcsSidecarResources }}...{{- end}}` block to satisfy the chart linter (`make test-chart` currently throws an error)

brigade-controller/cmd/brigade-controller/main.go
```diff
@@ -31,6 +31,10 @@ func main() {
 	flag.StringVar(&ctrConfig.WorkerImage, "worker-image", defaultWorkerImage(), "kubernetes worker image")
 	flag.StringVar(&ctrConfig.WorkerPullPolicy, "worker-pull-policy", defaultWorkerPullPolicy(), "kubernetes worker image pull policy")
 	flag.StringVar(&ctrConfig.WorkerServiceAccount, "worker-service-account", defaultWorkerServiceAccount(), "kubernetes worker service account name")
+	flag.StringVar(&ctrConfig.WorkerRequestsCPU, "worker-requests-cpu", "", "kubernetes worker cpu requests")
```
vdice: Shall we add/expose this configuration in the brigade chart such that these values can be set in the chart values.yaml?

jchanam:
@vdice Thanks for the comments! I've done what you suggested

charts/brigade-project/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: Create a Brigade project
 name: brigade-project
-version: 0.18.0
+version: 0.18.1
```
vdice: It looks like this value is used for sidecar image tags, which don't exist, naturally
```
  Normal   BackOff    13s   kubelet, aks-nodepool1-26912603-0  Back-off pulling image "deis/git-sidecar:v0.18.1"
  Warning  Failed     13s   kubelet, aks-nodepool1-26912603-0  Error: ImagePullBackOff
```
I say we bump this back to `0.18.0` here and in the `brigade/Chart.yaml` and we'll just bump to the next tagged version once this changeset is included in a release.

charts/brigade/values.yaml
```diff
@@ -45,6 +45,11 @@ controller:
   # The default tag is the value of .Chart.AppVersion
   #tag:
   #pullPolicy: IfNotPresent
+  #workerResources:
+  #  limitsCPU:
```
vdice: nit: do you think the structure of these values could be aligned with those used for the vcsSidecarResources in the brigade-project values.yaml?

(Meaning, `workerResources.limits.cpu`, `workerResources.limits.memory`, etc.)

charts/brigade-project/Chart.yaml
jchanam: I've added the `AppVersion` to be used for the image tag, in order to allow the chart to evolve.

This is the way it's done on the brigade chart, also It's what I understand is the best way to do it, after reading the helm documentation: https://docs.helm.sh/developing_charts/#charts-and-versioning and https://docs.helm.sh/developing_charts/#the-appversion-field

vdice:
This looks great; thanks again @jchanam !

PR #712, opened 2018-12-09
Add RBAC true value to quickstart

dgkanatsios:


radu-matei:
I'm actually unsure about whether to have it this way, or directly modify the value in `values.yaml`, as most out-of-the-box clusters now come with RBAC enabled by default.

dgkanatsios:
Agreed, closing

PR #715, opened 2018-12-10
fix(charts/brigade-project): allow vcs-sidecar NONE

vdice:
Fixes https://github.com/Azure/brigade/issues/710


radu-matei:
This is great, thanks for looking into this! 

While this works for pre-0.15.0, we should probably also have a fix for when creating projects using the `brig` CLI.

vdice:
@radu-matei added ability to disable the vcs sidecar container via `brig create`; doesn't look like this functionality existed previously...

dgkanatsios:
2 things and it'd be LGTM, I think
- I'd add a "Use NONE for no sidecar" message [here](https://github.com/Azure/brigade/blob/master/brig/cmd/brig/commands/project_create.go#L335)
- What happens when you add "NONE" is that the worker Pod has no initContainer (correct behavior), however the job Pods do get one. I think there is an extra check needed [here](https://github.com/Azure/brigade/blob/master/brigade-worker/src/k8s.ts#L344) so that nothing is assigned to the initContainers field when `project.kubernetes.vcsSidecar` is an empty string

vdice:
@dgkanatsios great calls!  Especially on the latter, it provoked further changes after digging into how the worker and job pods are created.  To that end, I pushed a recent commit wherein no sidecar resources (container, volume, volume mount) are added to worker or job pods if the sidecar is disabled (in other words, an empty string in the project secret).  Also added some test coverage.

Will proceed to test locally...  PTAL when you get a moment.  Thanks!

dgkanatsios:
Great work and tests, LGTM!

vdice:
ping @radu-matei on this PR.  Hoping to get this in/merged to enable transitioning the `brigade` chart to `Azure/brigade-charts` (per https://github.com/Azure/brigade/issues/718)

vdice:
Chart fix created in the separate brigade-charts repo now that they have been removed from this one: https://github.com/Azure/brigade-charts/pull/12

brigade-worker/src/k8s.ts
```diff
@@ -325,15 +325,14 @@ export class JobRunner implements jobs.JobRunner {
...
     ];
     this.runner.spec.containers[0].volumeMounts = [
-      { name: secName, mountPath: "/hook" } as kubernetes.V1VolumeMount,
-      { name: "vcs-sidecar", mountPath: mountPath } as kubernetes.V1VolumeMount
+      { name: secName, mountPath: "/hook" } as kubernetes.V1VolumeMount
     ];
 
-    if (job.useSource && project.repo.cloneURL) {
+    this.runner.spec.initContainers = [];
+    if (job.useSource && project.repo.cloneURL && project.kubernetes.vcsSidecar) {
```
radu-matei: nit: this is making me slightly nervous as we're checking a boolean and a string in the same way here. Could we check for value and length? Something along the lines of [this](https://codereview.stackexchange.com/a/5710)?

While this is by no means reason to bump this PR again, I'd love to get some linters and checkers in place to prevent this kinds of things in the future, and would like to fix this in #739 .

PR #716, opened 2018-12-10
ref(Makefile): switch to golangci for linting

vdice:
We've had good success with [golangci-lint](https://github.com/golangci/golangci-lint) in other [projects](https://github.com/deislabs/duffle), so proposing the switch here as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,7 +113,7 @@ test-js: bootstrap-js
 
 .PHONY: test-style
 test-style:
-	gometalinter --config ./gometalinter.json ./...
+	golangci-lint run --config ./golangci.yml
 
 .PHONY: test-chart
 test-chart:
@@ -130,7 +130,7 @@ format-go:
 format-js:
 	cd brigade-worker && yarn format
 
-HAS_GOMETALINTER := $(shell command -v gometalinter;)
+HAS_GOLANGCI     := $(shell command -v golangci-lint;)
 HAS_DEP          := $(shell command -v dep;)
 HAS_GIT          := $(shell command -v git;)
 HAS_YARN         := $(shell command -v yarn;)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -145,9 +145,8 @@ endif
 ifndef HAS_DEP
 	go get -u github.com/golang/dep/cmd/dep
 endif
-ifndef HAS_GOMETALINTER
-	go get -u github.com/alecthomas/gometalinter
-	gometalinter --install
+ifndef HAS_GOLANGCI
+	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(GOPATH)/bin
```
radu-matei: This breaks any Windows machines from doing a `make bootstrap`.

dgkanatsios: The alternative to wget would be to use Powershell, but I'm no expert there. Instead, maybe we can recommend [WSL](https://docs.microsoft.com/en-us/windows/wsl/install-win10), I'm using that for Brigade (and Golang in general) development and works perfectly fine.

radu-matei: The easiest way is to make two assumptions for local dependencies: `git` and `go` - both of which are reasonable if you are contributing to Brigade.

Then, we can clone the linter at the desired version, then go install it - or directly `go get` it, although not recommended, as it always pulls `master`.

radu-matei: This way, it works across any operating system, assuming you're trying to contribute to the project, and have `git` and `go`.

vdice: @radu-matei good call... how does the updated download procedure look?  (Pulled from https://github.com/golangci/golangci-lint#local-installation)

radu-matei:
We get colors in the output with this change 🎉 

Thanks a lot for this, @vdice !

radu-matei:
Now I want colors for all the other tests as well... Can we do that? Pretty please?

PR #717, opened 2018-12-11
ref(brigade.js): use GitHub App/Checks API

vdice:
Also
 - refactor jobs needing to loop over images (use canonical list in Makefile)
 - split up golang and javascript build/test jobs, add helm test job


brigade.js
```diff
@@ -48,73 +37,138 @@ function build(e, project) {
...
-    return ghNotify("failure", `failed build ${ e.buildID }`, e, project).run()
-  });
+    return j.run();
+  }
 }
 
-function releaseBrig(p, tag) {
+// Helper to wrap a job execution between two notifications.
+async function notificationWrap(job, note, conclusion) {
+  if (conclusion == null) {
```
dgkanatsios: conclusion is never set in any `notificationWrap` call, will always be undefined

brigade.js
```diff
@@ -48,73 +37,138 @@ function build(e, project) {
...
+    conclusion = "success"
+  }
+  await note.run();
+  try {
+    let res = await job.run()
+    const logs = await job.logs();
+
+    note.conclusion = conclusion;
```
dgkanatsios: maybe this could be set to 'success' directly

brigade.js
vdice: good call; fixed up

brigade.js
vdice: 👍 updated

vdice:
(Pushed amended commit sans any changes to trigger final Checks run as it required a config change from the usual webhook setup; then will merge.)

Issue #667, opened 2018-10-17
brig project create `--replace` flag does not work


blimmer:
I noticed that `brig project create` is documented to have a `--replace` flag. 

```
Usage:
  brig project create [flags]

Flags:
  -f, --config string         Path to JSON Kubernetes Secret.
  -D, --dry-run               Do not send the config to Kubernetes
  -p, --from-project string   Retrieve the given project from Kubernetes and use it to set the default values.
  -h, --help                  help for create
  -x, --no-prompts            Do not prompt the user for input. Use this with -f to skip prompts.
  -o, --out string            File where configuration should be saved. The configuration is stored as a JSON Kubernetes Secret
  -r, --replace               Replace an existing project with this revised one. Use this to modify existing projects.
```

However, I'm finding that it does not work as I expected. When I try to run this to replace an existing project, I get this error:

```
   -- command failed --
   status : 1
   output (20 lines):
     Project ID: brigade-6e35bd4c52f9e015cfe3308c2b19eba59990e1019beb499eadee67
     Error: secrets "brigade-6e35bd4c52f9e015cfe3308c2b19eba59990e1019beb499eadee67" already exists
     Usage:
       brig project create [flags]
```

I also wrote [a test](https://github.com/blimmer/brigade-project-integration-test/pull/2) to reproduce this error. Side note: this testing strategy would be a great addition to provide automated tests to this project.

dgkanatsios:
Your test is failing because ‘brig —replace’ will have Kube API Secrets.Create invoked, which fails because the Secret already exists (since the project exists). I think that this can easily be solved by calling API Secrets.Update called, when the —replace is used (and, of course, Secret exists).
Thoughts?

blimmer:
That sounds right to me!

dgkanatsios:
closing as #720 has been merged

PR #720, opened 2018-12-12
fixed brig project create --replace

dgkanatsios:
Proposed fix for #667 

Need a second opinion [here](https://github.com/dgkanatsios/brigade/blob/fix/project-create-replace/brig/cmd/brig/commands/project_create.go#L157) and [here](https://github.com/dgkanatsios/brigade/blob/fix/project-create-replace/brig/cmd/brig/commands/project_create.go#L165)
If I return the errors themselves, cobra will display usage info, which should not happen. So, I'm returning nil error and displaying a message instead. Is this sufficient?

### DIFF
--- a/brig/cmd/brig/commands/project_create.go
+++ b/brig/cmd/brig/commands/project_create.go
@@ -153,16 +153,14 @@ func createProject(out io.Writer) error {
 	// brig project create --replace
 	if projectCreateReplace {
 		if _, err = store.GetProject(proj.ID); err != nil {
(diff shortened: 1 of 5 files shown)


vdice:
> Btw, any hints on why CI is not running (noticed that in recent PRs)?
😬 Oops, I updated the project config for this repo per https://github.com/Azure/brigade/pull/717 to get Checks going, but forgot to switch back to default webhook while that change is still in PR form.

I've just reset the config -- when convenient, could you push an amended commit to re-trigger?  Sorry for the hassle!

vdice:
PS: no need to push an amended commit; found the webhook events and resent for this and #721  👍

Issue #602, opened 2018-08-21
Generic Webhook Handler


technosophos:
Per the 1.0.0 feature list:

- Add a generic webhook gateway to the API server (`POST /webook/:projectID/:secret`)
- Make this webhook callback configurable: It should be possible to disable this from config
- Add support in the charts/brigade chart (to turn of webhook gateway)

The gateway should create an event named `webhook`, and the payload should be whatever data was posted.

For `:secret`, we need to generate a per-project secret that can be embedded in the URL and validated per request. This prevents someone from finding all the projects and just spamming these projects with requests.

technosophos:
It should be disabled by default, and the main documentation should include enabling it:

```
$ helm install brigade/brigade --set generic-gateway=true
```

dgkanatsios:
Started a PoC for this on my local fork [here](https://github.com/dgkanatsios/brigade/commit/afb4c3466d50d13e23352231c5d5e0c8367128dd), looking for review

Points:
- what would the Generic Webhook payload? Currently I'm using 
```json
{
	"ref":"master",
	"commit": "lolo"
}
```
Is it enough?
- missing tests
- missing chart configuration (need to deep dive into charts' creation, pretty n00b)
- feeling a bit weird about [this](https://github.com/dgkanatsios/brigade/commit/afb4c3466d50d13e23352231c5d5e0c8367128dd#diff-eb4368b2ea4c9fce4a0dbedd95b4fbba). Explain: all source code files in the 'webhook' folder use Gin, whereas I used go-restful, since all the other methods on API server use that. Thoughts?
- added a new [field](https://github.com/dgkanatsios/brigade/commit/afb4c3466d50d13e23352231c5d5e0c8367128dd#diff-31af9081d3895845fef9878bdabf43dd) on Project struct with the GenericWebhook secret which can be [auto-generated](https://github.com/dgkanatsios/brigade/commit/afb4c3466d50d13e23352231c5d5e0c8367128dd#diff-f6bea21b11d01f1d30a6d06a42dfbd70)
- not sure about [this](https://github.com/dgkanatsios/brigade/commit/afb4c3466d50d13e23352231c5d5e0c8367128dd#diff-e85de98a3a4e38132ded5611895551c3R180) configuration

Thank you!

vdice:
@dgkanatsios the PoC looks great to me.  Let me know when I can help test and perhaps later next week @technosophos and others can take a look (hectic conference schedule currently!)

[...]

PR #721, opened 2018-12-12
Generic Webhook (#602)

dgkanatsios:
An attempt to implement #602, currently WIP

Includes
- Generic Webhook API Server
- Will start depending on `generic-gateway` command line flag
- Helm chart implementation (I'd appreciate a double/triple check here, first time editing a Chart)
- Tests

Does not include
- Documentation

Points
- feeling a bit weird about [this](https://github.com/Azure/brigade/compare/master...dgkanatsios:generic-webhook#diff-eb4368b2ea4c9fce4a0dbedd95b4fbba). Explain: all source code files in the 'webhook' folder use Gin, whereas I used go-restful, since all the other methods on API server use that. Thoughts?
- added a new [field](https://github.com/Azure/brigade/compare/master...dgkanatsios:generic-webhook#diff-31af9081d3895845fef9878bdabf43dd) on Project struct with the GenericWebhook secret which can be [auto-generated](https://github.com/Azure/brigade/compare/master...dgkanatsios:generic-webhook#diff-f6bea21b11d01f1d30a6d06a42dfbd70)
- not sure about [this](https://github.com/Azure/brigade/compare/master...dgkanatsios:generic-webhook#diff-e85de98a3a4e38132ded5611895551c3R180) configuration
- Generic Webhook handler will refuse to retrieve Projects that have empty secret [reference](https://github.com/Azure/brigade/compare/master...dgkanatsios:generic-webhook#diff-eb4368b2ea4c9fce4a0dbedd95b4fbbaR66)

Looking forward to your comments and reviews, is this what you had in mind?


charts/brigade/templates/api-deployment.yaml
```diff
@@ -23,6 +23,10 @@ spec:
       - name: {{ .Chart.Name }}
         image: "{{ .Values.api.registry }}/{{ .Values.api.name }}:{{ default .Chart.AppVersion .Values.api.tag }}"
         imagePullPolicy: {{ default "IfNotPresent" .Values.api.pullPolicy }}
+{{ if .Values.generic-gateway }}
```
vdice: Alas, `-` chars aren't allowed, so we'd need to go camelCase here...

While on the topic, I wonder if we do without the template conditional and just pass the value in, however it is set in the `values.yaml`:
```
        args:
          ["--generic-gateway", "{{ .Values.genericGateway }}"]
```
Also, I wonder if we should follow the precedent of appending `enabled` on the value for clarity, e.g. `genericGateway.enabled`, as is done for `rbac` ?  (Perhaps also the natural fit if we expect further properties to be added under the `genericGateway` config section.)

P.S., I tested the chart install via `helm install charts/brigade --dry-run --debug` to catch the aforementioned error, which also demonstrated a lack of test coverage in this repo, so I added a ticket to add, though in the future repo home of Brigade charts; see https://github.com/Azure/brigade-charts/issues/4.

charts/brigade/values.yaml
```diff
@@ -34,6 +34,8 @@
 rbac:
   enabled: true
 
+generic-gateway: false
```
vdice: (related to note prior re: needing to go camelCase/avoid `-`'s)

dgkanatsios:
thanks for the review, agree with all proposed changes 👍

brigade-api/cmd/brigade-api/main.go
```diff
@@ -32,6 +33,7 @@ func init() {
 	flag.StringVar(&namespace, "namespace", defaultNamespace(), "kubernetes namespace")
 	flag.StringVar(&apiPort, "api-port", defaultAPIPort(), "TCP port to use for brigade-api")
 	flag.BoolVar(&verbose, "verbose", false, "enables detailed logging of http request matching and filter invocation")
+	flag.BoolVar(&genericGateway, "genericGateway", false, "enables the generic webhook server")
```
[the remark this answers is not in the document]
radu-matei: Reading Vaughn's comments about the casing, haha! I think the CLI and deployment template should be consistent with the naming of flags and params - even if it means having two names for that...

radu-matei:
This looks great! I'll test the new gateway early next week, but at first glance it looks really great. 

Thanks!

brigade-api/cmd/brigade-api/main.go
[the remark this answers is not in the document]
dgkanatsios: thanks for the review. While it seems weird to have it `generic-gateway` on cmd and `genericGateway` on the chart, your reasoning makes perfect sense. Will fix and commit, thanks!

dgkanatsios:
Created PR on brigade-charts [here](https://github.com/Azure/brigade-charts/pull/16)
In any case, linter bit me again :( I suppose we can merge #728, then merge master and it will be fixed. Thoughts? cc @radu-matei @vdice

radu-matei:
One thing I don't see here (and I haven't actually gotten to testing just yet) is documentation - do we want to have it when merging? (It's a clear yes from me).

dgkanatsios:
definitely, was planning to work on that as soon as we all conclude that this PR contains all the features we want the Generic Webhook to have.

pkg/webhook/generic.go
```diff
@@ -0,0 +1,114 @@
...
+
+	if err != nil {
+		log.Printf("Failed to read GenericWebHookData: %s", err)
+		response.WriteErrorString(http.StatusBadRequest, "{\"status\": \"Malformed genericWebhookData\"}")
+		return
+	}
+
+	log.Printf("gwData: %#v\n", gwData)
+
+	body, err := ioutil.ReadAll(request.Request.Body)
```
technosophos: It might be good to move this part until after you load the project and compare the secret. Otherwise, you could end up with a denial of service-style attack where lots of anonymous clients send very large payloads to this endpoint.

docs/topics/genericwebhook.md
```diff
@@ -0,0 +1,45 @@
...
+## Intro to Generic Webhook 
+
+Brigade API Server can optionally be activated to accept `POST` webhook requests at `/webhook/:projectID/:secret` path. When this endpoint is called, Brigade will respond by creating a Build with a `webhook` event. This provides Brigade developers with the ability to trigger scripts based on messages received from any platform that can send a POST HTTP request.
+
+## Configuring Brigade API Server for Generic Webhook
+
+Generic Webhook support is disabled by default, but can easily be turned on during installation or upgrade of Brigade:
+
+```
+$ helm install -n brigade brigade/brigade --set generic-gateway.enabled=true
```
radu-matei: According to the chart, this should be `--set genericGataway.enabled=true` - https://github.com/Azure/brigade-charts/blob/f03f5a12a8df3eba12fdb44c4ac75282dbcbf7a5/charts/brigade/values.yaml#L49

radu-matei:
This looks great!
Could you please rebase and squash the commits with the small fixes? This helps keeping the commit history relevant, and reverting much easier.

Thanks!

dgkanatsios:
squashed a bunch of commits, let me know if it's better now, thanks!

dgkanatsios:
Rebased as well, let me know if anything else is needed, thanks!

radu-matei:
As discussed with @dgkanatsios, the commit history and changes here were getting pretty difficult to manage - so chery-picked, cleaned up, and opened #762 (This keeps the original author of the commits.)

Closing, and moving the discusison over in #762.
Thanks!

PR #722, opened 2018-12-12
ref(*): remove charts

vdice:
- remove `charts/`
 - remove `docs/*.tgz`
 - update `docs/index.yaml`
 - update docs

Depends on https://github.com/Azure/brigade-charts/pull/1 (completed)
Depends on https://github.com/Azure/brigade-charts/pull/6 (completed)
Ref https://github.com/Azure/brigade/issues/718


vdice:
Ok, this PR has been updated.  As mentioned in the updated description, this PR does the following:

 - remove `charts/`
 - remove `docs/*.tgz`
 - update `docs/index.yaml`
 - update docs

Since the updated `docs/index.yaml` references a patch version (adding Helm repo URL migration notes) for the brigade and brigade-project charts in https://github.com/Azure/brigade-charts/pull/6 , this is the only dependency prior to merge.

To put it in clearer terms, the strategy presented by this PR to ease transitioning around the chart URL change is: *only* serve the updated `index.yaml` (all chart urls prefixed with new host/path of `https://azure.github.io/brigade-charts/docs`) and remove everything else chart-related.  Therefore, users who haven't updated the corresponding chart repo URL can still fetch (though aforementioned notes and updated docs instruct to do so).  All subsequent chart releases will be in https://github.com/Azure/brigade-charts and the `index.yaml` in this repo will eventually be removed altogether.

Reviews appreciated @adamreese @michelleN @dgkanatsios et al.

PR #725, opened 2018-12-14
fix(brigade.js): add some fixes around publish/release

vdice:
- pass/consume correct params for goDockerBuild
 - return job from release func, to be ran with run() as appropriate
 - add noop() function

### DIFF
--- a/brigade.js
+++ b/brigade.js
@@ -22,6 +22,8 @@ const addMake =
   "apk upgrade 1>/dev/null && \
   apk add --update --no-cache make 1>/dev/null";
 
+const noop = {run: () => {return Promise.resolve()}};
+
 function goTest(e, project) {
   // Create a new job to run Go tests
   var goTest = new Job("go-test", goImg);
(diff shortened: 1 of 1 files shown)


vdice:
Tested on own fork/project; successful test release in https://github.com/vdice/brigade/releases/tag/vdice-test2

brigade.js
```diff
@@ -195,19 +197,19 @@ function release(p, tag) {
...
-function goDockerBuild(project, tag) {
+function goDockerBuild(e, p) {
   // We build in a separate pod b/c AKS's Docker is too old to do multi-stage builds.
   const builder = new Job(`${projectName}-docker-build`, goImg);
 
   builder.storage.enabled = true;
   builder.env = goEnv;
   builder.tasks = [
-    `cd /src && git checkout ${tag}`,
+    `cd /src`,
```
radu-matei: Is this because the sidecar already checked out the correct revision?

vdice: @radu-matei indeed 👍

PR #727, opened 2018-12-17
allow configuring the service account in the helm chart

rmb938:
In my environment RBAC and service accounts cannot be created via helm so I have added the ability to not create the service accounts as well as set their name.

### DIFF
--- a/charts/brigade/templates/api-deployment.yaml
+++ b/charts/brigade/templates/api-deployment.yaml
@@ -1,4 +1,6 @@
 {{ if .Values.api.enabled }}
+{{ $fullname := include "brigade.api.fullname" . }}
+{{ $serviceAccount := default $fullname .Values.api.serviceAccount.name }}
 apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
@@ -18,7 +20,7 @@ spec:
         app: {{ template "brigade.fullname" . }}-api
         role: api
     spec:
-      serviceAccountName: {{ template "brigade.api.fullname" . }}
+      serviceAccountName: {{ $serviceAccount }}
       containers:
       - name: {{ .Chart.Name }}
         image: "{{ .Values.api.registry }}/{{ .Values.api.name }}:{{ default .Chart.AppVersion .Values.api.tag }}"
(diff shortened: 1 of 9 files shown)


charts/brigade/values.yaml
```diff
@@ -50,6 +50,9 @@ controller:
   #  requests:
   #    cpu:
   #    memory:
+  serviceAccount:
+    create: true
+    name:
```
vdice: I wonder if it would be clearer to comment this line out, perhaps with an associated comment mentioning it only needs to be set if a non-default name is desired?

charts/brigade/templates/api-role.yaml
```diff
@@ -1,15 +1,18 @@
 {{ if .Values.api.enabled }}
 {{ $fname := include "brigade.api.fullname" . }}
+{{ $serviceAccount := default $fname .Values.api.serviceAccount.name }}
+{{ if .Values.api.serviceAccount.create }}
```
vdice: What do you think about a global `createServiceAccounts` bool (perhaps under the `rbac` section) instead of adding for each potential service account?  This assumes the use case is either all (can create service accounts) or none (cannot create any)...

rmb938: I did go that route at first, however looking at other helm charts in the main helm repo that have multiple service accounts, it seems that they typically have rbac and service account flags for each component.

charts/brigade/values.yaml
rmb938: That sounds reasonable.

charts/brigade/templates/api-deployment.yaml
```diff
@@ -1,4 +1,6 @@
 {{ if .Values.api.enabled }}
+{{ $fullname := include "brigade.api.fullname" . }}
+{{ $serviceAccount := default $fname .Values.api.serviceAccount.name }}
```
vdice: Looks like `$fname` should be `$fullname` here.

charts/brigade/templates/api-role.yaml
vdice: Ah, I see.  Indeed, I'd favor going with convention as well.

charts/brigade/values.yaml
vdice: I am noting this is aligned with the convention in other charts around `serviceAccount.name`, so I'm good with keeping as is -- defer to the author.

vdice:
Thanks @rmb938 !

Issue #643, opened 2018-10-01
Show Job pod logs during `brig run`


blimmer:
Today, the `brig run` command only shows the logs from the main worker. It'd be nice if you could also see logging from the Job pods while they were running. For instance:

```
brig run -f brigade.js my_project
Event created. Waiting for worker pod named "brigade-worker-01crrfzqmvndx8fr2zhnhakfqj".
Build: 01crrfzqmvndx8fr2zhnhakfqj, Worker: brigade-worker-01crrfzqmvndx8fr2zhnhakfqj
prestart: no dependencies file found
prestart: loading script from /etc/brigade/script
[brigade] brigade-worker version: 0.17.0
[brigade:k8s] Creating PVC named brigade-worker-01crrfzqmvndx8fr2zhnhakfqj
[brigade:k8s] Creating secret something-01crrfzqmvndx8fr2zhnhakfqj
[brigade:k8s] Creating pod something-01crrfzqmvndx8fr2zhnhakfqj
[brigade:k8s] Timeout set at 900000
[brigade:k8s] default/something-01crrfzqmvndx8fr2zhnhakfqj phase Pending
[brigade:k8s] default/something-01crrfzqmvndx8fr2zhnhakfqj phase Running
[brigade:k8s:something-01crrfzqmvndx8fr2zhnhakfqj] Hello from the Job logs!
```

Alternatively, maybe it could be configured in the Javascript API? e.g.

```javascript
var job = new Job("something", "node:8", ["my_task"]);
job.displayLogs = true;
```

Then, the `JobRunner` could fetch the logs at some interval and display them using `console.log`.

For now, you can work around this by fetching the logs after the run and outputting them to the console.

```javascript
  var job = new Job("something", "node:8", ["echo hi"]);
  try {
    await job.run();
  } finally {
    const logs = await job.logs();
    console.log(logs);
  }
```

OrthoDex:
Hi! Can I take this up?

blimmer:
That'd be great!

dgkanatsios:
Tried this today and seemed pretty trivial, did a PR on #728 
@OrthoDex have you started working on that? If yes, let me know and I'll close the PR :)

OrthoDex:
Hey @dgkanatsios, unfortunately, I haven't had much time to work on this :disappointed: 
You can continue with your PR, thanks for asking!

radu-matei:
Merged in #728

PR #728, opened 2018-12-17
show job logs (#643)

dgkanatsios:
Implementation for #643 
Specifically, added `streamLogs` boolean variable to Job object on brigade.js. This variable triggers the streaming/follow of the Job Pod(s) Logs, similar to `kubectl logs PODNAME -f` functionality.


vdice:
@dgkanatsios code LGTM and tested with a local project.

Note: While logs indeed show up at the end of a job run, I'm not seeing the logs stream in realtime.  If streaming was one of the criteria/intended enhancements, perhaps some more work is needed -- otherwise, looks great!

dgkanatsios:
thanks @vdice. As usual, you are correct, streaming (or in the K8s terminology, "following") the logs makes much more sense. Just pushed an implementation for that, let me know

dgkanatsios:
Both done, thanks. The "done" output is a bit tricky to remove, since it's not being sent from the worker but from the [delegated runner](https://github.com/Azure/brigade/blob/master/pkg/script/delegated_runner.go#L217)

brigade-worker/src/job.ts
```diff
@@ -245,6 +245,9 @@ export abstract class Job {
     return this._podName;
   }
 
+  /** displayLogs controls whether logs from the job Pod will be sent to output */
+  public displayLogs : boolean = false;
```
radu-matei: Is there a different variable name for following / tailing the logs?

I think one of those might be a little more intuitive if this actually follows logs (but this is just my opinion, though)

dgkanatsios: Hm, in the ‘brig run’ context, logs are streamed/followed/sent to stdout as ‘brig run’ is executed. There is no other behavior (mainly because ‘brig run’ gets executed with a ‘follow’ pattern, worker logs are also sent directly to stdout). In any case, I’m totally open to rename ‘displayLogs’ to ‘followLogs’ or ‘streamLogs’, thoughts? Maybe ‘grabLogs’ or ‘captureLogs’? Any particular choice?

mumoshu: Not a strong opinion, but `streamLogs` sounds good to me.

My mental model is that, `brig run` *follows* logs from the brigade worker. On the other hand the worker *streams* relevant logs to the worker. This change adds job pod logs to what's streamed by the worker. So, using `stream` works to me.

@dgkanatsios @radu-matei WDYT?

Btw, thanks a lot for your PR :) This had been a long-awaited feature to me!

dgkanatsios: agreed, `streamLogs` is a great choice, pushed a commit with this change. Thanks for the comment, hope this feature proves to be useful to you!

dgkanatsios:
@vdice and @radu-matei issue with the linter is that it requires import alias for short or dash-containing package names like:
- `"k8s.io/api/core/v1"` should become `v1 "k8s.io/api/core/v1"`
- `"github.com/emicklei/go-restful"` should become `restful "github.com/emicklei/go-restful"`
- `"gopkg.in/yaml.v2"` should become `yaml "gopkg.in/yaml.v2"`
- `"gopkg.in/gin-gonic/gin.v1"` should become `gin "gopkg.in/gin-gonic/gin.v1"`

I can fix that and re-commit here (well, VS Code 'save-file' functionality fixes that), just wanna make sure this is the desired behaviour.

vdice:
@dgkanatsios interesting; I wonder if this was a recent linting update/change?  SGTM.

dgkanatsios:
thank you, just pushed, it seems OK now. As to how this was introduced, no clue, really. Maybe it comes with golangci? Have to admit that it 'feels' better, though

vdice:
@radu-matei could you take a look when convenient?

dgkanatsios:
thanks! about the commit history, I was scratching my head as well, not exactly sure why this happens. Branch on my fork ([link](https://github.com/dgkanatsios/brigade/tree/job-pod-logs)) mentions 'This branch is 6 commits ahead of Azure:master.' which is accurate...

brigade-worker/src/job.ts
radu-matei: I know, too late, but `streamLogs` is a great name 😄

radu-matei:
One question about the `phase Running/Succeeded` output - why can't we remove it when following the logs (or always?).

cc @mumoshu, who I think worked on the delegated runner.

brigade-worker/src/k8s.ts
```diff
@@ -46,7 +46,7 @@ const wrapClient = fns => {
   // wrap client methods with retry logic
   for (let fn of fns) {
     let originalFn = defaultClient[fn.name];
-    defaultClient[fn.name] = function() {
+    defaultClient[fn.name] = function () {
```
radu-matei: I don't think there's an actual change here, right?
As it doesn't add anything, and we should probably add some linters for TypeScript and fix them everywhere (see #739), would it make sense to revert this type of changes in this file?

brigade-worker/src/k8s.ts
```diff
@@ -439,7 +439,7 @@ export class JobRunner implements jobs.JobRunner {
     // appended to job name.
     return `${this.project.name.replace(/[.\/]/g, "-")}-${
       this.job.name
-    }`.toLowerCase();
+      }`.toLowerCase();
```
radu-matei: Here as well.

brigade-worker/src/k8s.ts
```diff
@@ -461,7 +461,7 @@ export class JobRunner implements jobs.JobRunner {
    */
   public run(): Promise<jobs.Result> {
     return this.start()
-      .then(r => r.wait())
+      .then(r => this.wait())
```
radu-matei: Is this a linter recommendation?

```
Promise<JobRunner>.then<jobs.Result, never>(onfulfilled?: (value: jobs.JobRunner) => jobs.Result>
```

Technically, the `r` parameter passed to the promise is a `JobRunner`, so we can safely do `r.wait()` - I'm curious why the change, and if I'm missing something.

brigade-worker/src/k8s.ts
```diff
@@ -626,11 +630,16 @@ export class JobRunner implements jobs.JobRunner {
           this.logger.log("Pod not yet scheduled");
           return;
         }
+
         let phase = this.pod.status.phase;
         if (phase == "Succeeded") {
           clearTimers();
           let result = new K8sResult(phase);
           resolve(result);
+        } else if (phase == "Running") { // make sure Pod is running before we start following its logs
```
radu-matei: nit: can the comment be on the previous line?
(This way, we're consistent with the single line comments - at least in this file).

brigade-worker/src/k8s.ts
```diff
@@ -583,6 +583,8 @@ export class JobRunner implements jobs.JobRunner {
     return req;
   }
 
+
+
```
radu-matei: nit: One extra newline here.

brigade-worker/src/k8s.ts
dgkanatsios: sure, fixing

brigade-worker/src/k8s.ts
dgkanatsios: fixing

brigade-worker/src/k8s.ts
dgkanatsios: nah, this was Dimitris-visual-linter :)
probably should have discussed this first, but it seemed to me that both Promises should have either _r.Function_ or _this.Function_. In any case, reverting this as well

brigade-worker/src/k8s.ts
dgkanatsios: this was most certainly caused by VS Code auto-formatter on cmd-s while I was editing on Mac. I'm now editing on Windows and, weirdly, ctrl-s does not format it, so I'll revert it (and I'm pretty sure that I haven't done any explicit JS configuration on VS Code settings, so who knows...)

brigade-worker/src/k8s.ts
dgkanatsios: same as above

dgkanatsios:
thanks @radu-matei for the review, definitely +1 for the usage of TS linters

PR #752, opened 2019-01-05
Allow setting the job service account

### LINKED ISSUE #749: Job Service Account
Currently the service account jobs use is set to the `BRIGADE_SERVICE_ACCOUNT` environment variable set on the worker which is set via `BRIGADE_WORKER_SERVICE_ACCOUNT` on the controller.

It seems like this should be split into two different variables i.e

`BRIGADE_WORKER_SERVICE_ACCOUNT` - the service account used by worker pods
`BRIGADE_JOB_SERVICE_ACCOUNT` - the service account used by job pods

This will allow jobs to be more secure as currently they are able to do everything the worker is allowed to do (create, update, delete, ect.. pods, secrets, pvcs and configmaps).

I do not want my jobs to be able to do those things as they could pose a security risk.

rmb938:

The ability to set the job service account. Fixes #749 

**If applicable**:
- [X] this PR has been tested for backwards compatibility


brigade-controller/cmd/brigade-controller/main.go
```diff
@@ -31,6 +32,7 @@ func main() {
 	flag.StringVar(&ctrConfig.WorkerImage, "worker-image", defaultWorkerImage(), "kubernetes worker image")
 	flag.StringVar(&ctrConfig.WorkerPullPolicy, "worker-pull-policy", defaultWorkerPullPolicy(), "kubernetes worker image pull policy")
 	flag.StringVar(&ctrConfig.WorkerServiceAccount, "worker-service-account", defaultWorkerServiceAccount(), "kubernetes worker service account name")
+	flag.StringVar(&ctrConfig.JobServiceAccount, "job-service-account", defaultJobServiceAccount(), "kubernetes job service account name")
```
radu-matei: This means the service account used by Brigade jobs is a global setting per Brigade deployment - meaning _all_ projects ran by this Brigade instance must use the same service account. 

Would it make sense for this to be a per-project setting? I might want to grant one project the ability to create resources, but not others - for example a CD pipeline that needs to deploy a new version of a deployment - but not others. In the version proposed here, I'd need to either grant all of them this permission, or have two Brigade instances (which is the case right now).

What do you think?

Also, nit: `"kubernetes job service account name"` - since we're not creating actual Kubernetes jobs, this should probably be `"brigade job service account name"`.

rmb938: I actually was going to add that ability in another pr. That way there can be a global default and jobs can override it if needed.

And yes I can change the help text.

brigade-controller/cmd/brigade-controller/controller/handler.go
```diff
@@ -234,7 +234,7 @@ func workerEnv(project, build *v1.Secret, config *Config) []v1.EnvVar {
 		{Name: "BRIGADE_REMOTE_URL", Value: string(project.Data["cloneURL"])},
 		{Name: "BRIGADE_WORKSPACE", Value: "/vcs"},
 		{Name: "BRIGADE_PROJECT_NAMESPACE", Value: build.Namespace},
-		{Name: "BRIGADE_SERVICE_ACCOUNT", Value: config.WorkerServiceAccount},
+		{Name: "BRIGADE_SERVICE_ACCOUNT", Value: config.JobServiceAccount},
```
radu-matei: If we consider per-project service accounts, the value from the project secret would take precedence over the default one set by the controller, right?

pkg/storage/kube/project.go
```diff
@@ -174,6 +175,7 @@ func NewProjectFromSecret(secret *v1.Secret, namespace string) (*brigade.Project
 	proj.Kubernetes.BuildStorageSize = def(sv.String("buildStorageSize"), "50Mi")
 	proj.Kubernetes.BuildStorageClass = sv.String("kubernetes.buildStorageClass")
 	proj.Kubernetes.CacheStorageClass = sv.String("kubernetes.cacheStorageClass")
+	proj.Kubernetes.ServiceAccount = sv.String("serviceAccount")
```
rmb938: This is going to break backwards compatibility, not sure if we want to prevent issues by adding some sort of `StringOrDefault` to `SecretValues`. Thoughts?

rmb938: Oh actually nevermind.

rmb938:
@radu-matei What do you think about adding a flag to only allow service accounts that start with a certain prefix or a regex.

Not really sure how we would validate that during project creation though since it doesn't go through the brigade api

radu-matei:
Well, we're in almost the same situation now, aren't we?
The user creating a Brigade project wants to use a specific service account for jobs. That information is unused until a build is actually created for that project - in which case, if the SA exists everything is OK - if not, it fails because the missing SA.

The same would be true if we validate with a regex - you try to use a SA, but until a build is scheduled, you don't know if a) the SA even exists and b) if the you are allowed to use it.

I like this, and the more I think about it, the more it makes sense - however, we should have the a sensible default here.

cc @technosophos

rmb938:
Yea that is true it will error during the build process either way.

I didn't want to break backwards compatibility so I left the default service account as `brigade-worker` but on my own deployments I will just be setting it to `default`.

If we don't really care I can change the default to just `default` as that service account typically has no permissions to do anything.

rmb938:
Also for the regex checking, should that be done in the controller or in the worker? If it's done in the worker we could print out some logs with error messages about it. I am not too sure how to implement that in the javascript though so I will need some help with it.

radu-matei:
Before changing the default service account we should also keep in mind lots of people use Brigade for CD inside the cluster, for example with a Helm step. (While we're still pre-v1.0, and technically we could break backwards compatibility, this would be pretty significant - on the other hand, if we want to do it, we should do it pre-v1.0).

There are implications to where to check as well - if the check is done in the controller, the worker wouldn't get started, so there'd be no logs for the build. Meaning that someone who doesn't have access to the controller logs would have no idea what happened.

radu-matei:
Oh, great suggestion, @vdice!
I think some tests (pretty much the same structure as the ones in your previous PR) would be great.

Thanks!

rmb938:
Will do, I'll get some tests added!

rmb938:
Sorry for the delay, been a bit busy over the past week but I plan on getting the tests added this weekend.

radu-matei:
Ping @rmb938.

rmb938:
@radu-matei I've added tests for the controller and rebased. I am not too sure how to go about adding tests for the regex validation since it's an environment variable (can't seem to change the environment during a test) and the check exists the process.

radu-matei:
@rmb938 - awesome, thanks!
I'll have a look early tomorrow at this.

vdice:
Ping @radu-matei for a final approval.

PR #753, opened 2019-01-05
add the ability to toggle usage of secretKeyRef in jobs

### LINKED ISSUE #750: Allowing secretKeyRef could be a security risk
Allowing `secretKeyRef` in jobs can be considered a security risk and I would like the ability to disable this functionality and/or the ability to whitelist secret names.

Since service account tokens are kubernetes secrets a job can use any service account's token in the namespace. 

This is very bad as there may be service accounts in the namespace with elevated permissions that jobs should not have. (related issue #749)

rmb938:
**What this PR does / why we need it**: Adds the ability to toggle usage of `secretKeyRef` in jobs. Fixes #750 

I am not too familiar with javascript so I am not sure if this is the correct way to do it.

### DIFF
--- a/brigade-controller/cmd/brigade-controller/controller/controller_test.go
+++ b/brigade-controller/cmd/brigade-controller/controller/controller_test.go
@@ -13,6 +13,8 @@ import (
 	core "k8s.io/client-go/testing"
 )
 
+const expectedEnvironmentLength = 16
+
 func TestController(t *testing.T) {
 	createdPod := false
 	client := fake.NewSimpleClientset()
@@ -94,8 +96,8 @@ func TestController(t *testing.T) {
 	if c.Name != "brigade-runner" {
 		t.Error("Container.Name is not correct")
 	}
-	if envlen := len(c.Env); envlen != 15 {
-		t.Errorf("expected 15 items in Container.Env, got %d", envlen)
+	if envlen := len(c.Env); envlen != expectedEnvironmentLength {
+		t.Errorf("expected %d items in Container.Env, got %d", expectedEnvironmentLength, envlen)
 	}
 	if c.Image != config.WorkerImage {
 		t.Error("Container.Image is not correct")
@@ -127,8 +129,8 @@ func TestController(t *testing.T) {
 		t.Fatalf("Expected 1 init container, got %d", l)
 	}
 	ic := pod.Spec.InitContainers[0]
-	if envlen := len(ic.Env); envlen != 15 {
-		t.Errorf("expected 15 env vars, got %d", envlen)
+	if envlen := len(ic.Env); envlen != expectedEnvironmentLength {
+		t.Errorf("expected %d env vars, got %d", expectedEnvironmentLength, envlen)
 	}
 
 	if ic.Image != sidecarImage {
@@ -233,8 +235,8 @@ func TestController_WithScript(t *testing.T) {
 	if c.Name != "brigade-runner" {
 		t.Error("Container.Name is not correct")
 	}
-	if envlen := len(c.Env); envlen != 15 {
-		t.Errorf("expected 15 items in Container.Env, got %d", envlen)
+	if envlen := len(c.Env); envlen != expectedEnvironmentLength {
+		t.Errorf("expected %d items in Container.Env, got %d", expectedEnvironmentLength, envlen)
 	}
 	if c.Image != config.WorkerImage {
 		t.Error("Container.Image is not correct")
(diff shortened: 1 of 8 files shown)


rmb938:
This doesn't prevent someone from extending the Job class and adding the references manually of course. So maybe this is pointless :man_shrugging:

brigade-controller/cmd/brigade-controller/controller/controller_test.go
```diff
@@ -94,7 +94,7 @@ func TestController(t *testing.T) {
 	if c.Name != "brigade-runner" {
 		t.Error("Container.Name is not correct")
 	}
-	if envlen := len(c.Env); envlen != 15 {
+	if envlen := len(c.Env); envlen != 16 {
 		t.Errorf("expected 15 items in Container.Env, got %d", envlen)
```
dgkanatsios: "expected 15" should now be "expected 16"

brigade-controller/cmd/brigade-controller/controller/controller_test.go
```diff
@@ -127,7 +127,7 @@ func TestController(t *testing.T) {
 		t.Fatalf("Expected 1 init container, got %d", l)
 	}
 	ic := pod.Spec.InitContainers[0]
-	if envlen := len(ic.Env); envlen != 15 {
+	if envlen := len(ic.Env); envlen != 16 {
 		t.Errorf("expected 15 env vars, got %d", envlen)
```
dgkanatsios: "expected 15" should now be "expected 16"

brigade-controller/cmd/brigade-controller/controller/controller_test.go
```diff
@@ -233,7 +233,7 @@ func TestController_WithScript(t *testing.T) {
 	if c.Name != "brigade-runner" {
 		t.Error("Container.Name is not correct")
 	}
-	if envlen := len(c.Env); envlen != 15 {
+	if envlen := len(c.Env); envlen != 16 {
 		t.Errorf("expected 15 items in Container.Env, got %d", envlen)
```
dgkanatsios: "expected 15" should now be "expected 16"

brigade-controller/cmd/brigade-controller/controller/controller_test.go
```diff
@@ -309,7 +309,7 @@ func TestController_NoSidecar(t *testing.T) {
 	}
 
 	c := pod.Spec.Containers[0]
-	if envlen := len(c.Env); envlen != 15 {
+	if envlen := len(c.Env); envlen != 16 {
 		t.Errorf("expected 15 items in Container.Env, got %d", envlen)
```
dgkanatsios: "expected 15" should now be "expected 16"
maybe the 15 should be a constant somewhere in this file?)

brigade-controller/cmd/brigade-controller/controller/handler.go
```diff
@@ -243,6 +246,7 @@ func workerEnv(project, build *v1.Secret, config *Config) []v1.EnvVar {
 			ValueFrom: secretRef("github.token", project),
 		},
 	}
+
```
radu-matei: nit: extra new line here.

brigade-controller/cmd/brigade-controller/controller/controller_test.go
rmb938: Yup adding a constant would be a good idea. I'll get that added later today.

brigade-controller/cmd/brigade-controller/main.go
```diff
@@ -35,6 +35,7 @@ func main() {
 	flag.StringVar(&ctrConfig.WorkerRequestsMemory, "worker-requests-memory", "", "kubernetes worker memory requests")
 	flag.StringVar(&ctrConfig.WorkerLimitsCPU, "worker-limits-cpu", "", "kubernetes worker cpu limits")
 	flag.StringVar(&ctrConfig.WorkerLimitsMemory, "worker-limits-memory", "", "kubernetes worker memory limits")
+	flag.BoolVar(&ctrConfig.AllowSecretKeyRef, "allow-secret-key-ref", true, "allow jobs to use secretKeyRef in their environment")
```
radu-matei: Again, I think we want this to be a per-project setting, not a global one. What do you think?

rmb938: Similar to the service account we could have a global default with the ability for projects to override?

radu-matei: Yes, I think that would be the right approach.
Thanks!

rmb938: Actually while adding this it probably isn't needed since golang bools will default to false so projects will always have it turned off unless set to true.

brigade-worker/src/k8s.ts
```diff
@@ -319,6 +320,11 @@ export class JobRunner implements jobs.JobRunner {
         // For environmental variables that are directly references,
         // add the reference to the env var list.
 
+        if (val.secretKeyRef != null && !allowSecretKeyRef) {
+         // allowSecretKeyRef is not to true so disallow setting secrets in the environment
```
vdice: Thoughts on logging this message to user, on the off-chance `allowSecretKeyRef` is currently `false` but not known to current user expecting secretKeyRefs in env?

rmb938: Yup logging to the user makes sense, I'll get that done tomorrow.

rmb938: dome

PR #766, opened 2019-01-10
added option to specify brigade.js path on project creation

### LINKED ISSUE #765: Custom per-project brigade.js location
We need a way to specify a custom location for a `brigade.js` file within a repository.

We already have a [list of default locations](https://github.com/Azure/brigade/blob/master/brigade-worker/prestart.js#L5-L21) for the `brigade.js` file where the Brigade worker searches for it:

```
// Script locations in order of precedence.
const scripts = [
  // manual override for debugging
  process.env.BRIGADE_SCRIPT,

  // data mounted from event secret (e.g. brig run)
  "/etc/brigade/script",

  // checked out in repo
  "/vcs/brigade.js",

  // data mounted from project.DefaultScript
  "/etc/brigade-project/defaultScript",

  // mounted configmap named in brigade.sh/project.DefaultScriptName
  "/etc/brigade-default-script/brigade.js"
];
```

We need to add a prompt to `brig project create` for the custom script location within the repo (as advanced configuration for project), then read that value in the controller and set it as `$BRIGADE_SCRIPT` environment variable on the worker (we already do the same with build ID, commit ID, event provider, type and others). 

Then, when the worker starts, the first place it looks is already a custom location - however, we need to be careful when parsing it, as when creating the project, the script location should be relative to the repo root, while in the worker it has to be relative to `/vcs` (directory in the worker pod where Brigade clones the repo).

dgkanatsios:

Provides an option during `brig project create` to specify a custom Brigade.js location

Closes #765

### DIFF
--- a/brig/cmd/brig/commands/project_create.go
+++ b/brig/cmd/brig/commands/project_create.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/Masterminds/goutils"
@@ -410,7 +411,7 @@ func projectAdvancedPrompts(p *brigade.Project) error {
 			Name: "tag",
 			Prompt: &survey.Input{
 				Message: "Custom worker image tag",
-				Help:    "The woerker image tag to pull, e.g. 1.2.3 or latest",
+				Help:    "The worker image tag to pull, e.g. 1.2.3 or latest",
 				Default: p.Worker.Tag,
 			},
 		},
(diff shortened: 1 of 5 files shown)


brig/cmd/brig/commands/project_create.go
```diff
@@ -471,6 +472,21 @@ func projectAdvancedPrompts(p *brigade.Project) error {
 				Default: p.DefaultScriptName,
 			},
 		},
+		{
+			Name: "brigadejsPath",
+			Prompt: &survey.Input{
+				Message: "Brigade.js relative path",
```
radu-matei: I would word this as "brigade.js file path relative to the repository root".

dgkanatsios: fixed, thanks!

brigade-controller/cmd/brigade-controller/controller/handler.go
```diff
@@ -243,6 +245,12 @@ func workerEnv(project, build *v1.Secret, config *Config) []v1.EnvVar {
 			ValueFrom: secretRef("github.token", project),
 		},
 	}
+
+	if psv.String("brigadejsPath") != "" {
```
radu-matei: Would it make sense to sanitize the path and check if it actually is a relative path?

radu-matei:
This looks great, thanks for picking it up!

I would also like to see some tests on the Brigade worker and a note in the docs. But otherwise, this looks great, thanks!

brigade-controller/cmd/brigade-controller/controller/handler.go
dgkanatsios: added!

dgkanatsios:
thanks for the review! Not sure what would be the ideal page in the documentation to add this, any hints?

brigade-controller/cmd/brigade-controller/controller/handler.go
dgkanatsios: done!

PR #776, opened 2019-01-18
[WIP] declarative pipelines

FrodeHus:
@radu-matei 
## Use case

Defining components that can be re-used and composed into pipelines - all in YAML.
The pipelines themselves can be composed of other pipelines in addition to components.
This allows for the creation of a library of Brigade components and pipelines ready for use without the need to write brigade scripts.

For example, someone can create a security scan pipeline and you can then refer to that pipeline as a step in your own pipeline without having to create and maintain it yourself.

These declarative pipelines should also make it easy to create nice visualizations of the build process.
The end result of the pipelines is the generation of a brigade.js that is attached to the worker just as default script is today.


FrodeHus:
Build issues due to merging with master

PR #777, opened 2019-01-18
[WIP] Declarative pipelines

FrodeHus:
## Use case
Defining components that can be re-used and composed into pipelines - all in YAML.
The pipelines themselves can be composed of other pipelines in addition to components.
This allows for the creation of a library of Brigade components and pipelines ready for use without the need to write brigade scripts.

For example, someone can create a security scan pipeline and you can then refer to that pipeline as a step in your own pipeline without having to create and maintain it yourself.

These declarative pipelines should also make it easy to create nice visualizations of the build process.
The end result of the pipelines is the generation of a brigade.js that is attached to the worker just as default script is today.

cc @radu-matei


brandonkal:
Why is there a desire to turn YAML into a scripting language? You can easily do something like:
```ts
import * as actions from 'actions'
actions.securityScan(params)
```
If you really must use YAML, handling that in a custom worker makes more sense.

krancour:
First, let me apologize, because I don't know how this PR escaped our attention for so long, but I agree with @brandonkal that support for declarative pipelines would be better achieved through a custom worker image, and in fact, there's been some work done on that front that was demoed at KubeCon last year. This is not _official_ in any capacity yet, but it works...

https://github.com/lovethedrake/brigdrake

Officially, Brigade 2.0 ([read the proposal here](https://github.com/brigadecore/brigade/pull/1101)) will be drastically improving support for custom workers. After that is accomplished, you can reasonable expect support for declarative pipelines to become official in some subsequent 2.x release. And even better, with custom workers becoming easier to implement, the community is encouraged to create and contribute alternative workers, so please do consider adapting the code in this PR into something that can be natively executed by a new worker, rather than using code generation to turn YAML into JS.

Issue #450, opened 2018-05-10
Document loading JS files into `brigade.js`


technosophos:
It is fairly easy to store additional JS files in your VCS repo, and then load it into the `brigade.js` using `require()`. But this is not documented.

vdice:
I believe @radu-matei may have some more extensive examples, but my rudimentary exercise for including local js yielded the following:

`brigade.js` has:
```javascript
const { events, Job}  = require("brigadier");
const { Helper } = require("/vcs/helpers");

events.on("exec", (e, project) => {
  try {
    var helper = new Helper(e, project, Job);
    helper.slackNotify("===> exec event received!");
  } catch (e) {
    console.log(e);
  }
})
```

(where I wrapped the main functionality in a `try/catch` to capture errors that didn't seem to properly bubble up otherwise... need to investigate why this was needed.)

`helpers.js` has:
```javascript
class Helper {
  constructor(event, project, Job) {
    this.event = event;
    this.project = project;
    this.Job = Job;
  }

  slackNotify(message) {
    var slack = new this.Job("slack-notify", "technosophos/slack-notify:latest");
  
    slack.env = {
      SLACK_WEBHOOK: this.project.secrets.SLACK_WEBHOOK,
      SLACK_USERNAME: "BrigadeBot",
      SLACK_TITLE: "Build " + this.event.buildID,
      SLACK_MESSAGE: message,
      SLACK_COLOR: "#00ff00"
    };
    slack.tasks = ["/slack-notify"];
  
    if (this.event.cause && this.event.cause.trigger != "success") {
      slack.env.SLACK_COLOR = "#ff0000"
    }
  
    slack.run();
  }
}

exports.Helper = Helper;
```

A few notes from my exercise:

1. I seem to only be able to fetch local JS files when supplying the full path of `/vcs/path/to/file`.  I believe this is because Brigade currently doesn't have logic to copy non-`brigade.js` files into `/home/src/dist` in the worker pod (hence, `require("./helpers")` doesn't work.)  If this reasoning is correct, I wonder if we should consider an enhancement such that relative pathing can work?

2. Since my `helpers.js` file creates a `Job`, I needed to pass `Job` in.  It didn't appear possible to add a `const { Job } = require("brigadier");` as usual at the top of these non-`brigade.js` scripts.  Again, it would be nice to avoid needing to pass in brigadier classes via constructor or function invocation.  Ideas on how we might do so and/or what I might've been missing here?

radu-matei:
[...]

PR #787, opened 2019-01-29
 Enable pretty import paths for local imports (from repo) in brigade.js

radu-matei:
This PR enabled pretty import paths for local JavaScript modules from the same repo as the Brigade project, without the user having to know about the destination in the Brigade worker pod of the source code.

Specifically, it intercepts all local path imports in `brigade.js` and overrides them to the absolute path in the worker pod (`/vcs/<rest-of-relative-path>`).

For more context, see https://github.com/Azure/brigade/issues/450#issuecomment-458365387.

This PR does _not_ contain the docs for this feature, as #450 takes the larger task of documenting that in detail.

### DIFF
--- a/brigade-worker/src/require.ts
+++ b/brigade-worker/src/require.ts
@@ -18,16 +18,20 @@ class Pkg {
 }
 
 export function overridingRequire(pkg: string): any {
+  return require(getOverriddenPackage(pkg));
+}
+
+export function getOverriddenPackage(pkg: string): string {
   for (let p of pkgOverrides) {
     if (p.name == pkg) {
-      return require(p.override);
+      return p.override;
     }
 
     // we want to intercept the loading of relative modules in the repo
-    // and and override with the absolute path inside the worker pod
+    // and and override them with the absolute path inside the worker pod
     if ((!path.isAbsolute(pkg)) && (pkg.includes("./"))) {
-      return require(pkg.substr(0, pkg.indexOf("./")) + "/vcs" + pkg.substr(pkg.indexOf("./") + 1))
+      return pkg.substr(0, pkg.indexOf("./")) + "/vcs" + pkg.substr(pkg.indexOf("./") + 1);
     }
   }
-  return require(pkg);
+  return pkg;
 }
--- a/brigade-worker/src/require.ts
+++ b/brigade-worker/src/require.ts
@@ -1,3 +1,5 @@
+const path = require("path");
+
 const pkgOverrides: Pkg[] = [
   { name: "brigade", override: "./brigadier" },
   { name: "brigadier", override: "./brigadier" },
@@ -19,6 +21,12 @@ export function overridingRequire(pkg: string): any {
   for (let p of pkgOverrides) {
     if (p.name == pkg) {
       return require(p.override);
+    }
+
+    // we want to intercept the loading of relative modules in the repo
+    // and and override with the absolute path inside the worker pod
+    if ((!path.isAbsolute(pkg)) && (pkg.includes("./"))) {
+      return require(pkg.substr(0, pkg.indexOf("./")) + "/vcs" + pkg.substr(pkg.indexOf("./") + 1))
     }
   }
   return require(pkg);
--- a/docs/topics/dependencies.md
+++ b/docs/topics/dependencies.md
@@ -1,0 +1,123 @@
+# Import dependencies in your `brigade.js` file
+
+A Brigade worker is responsible for executing your `brigade.js` file. By default, Brigade comes with a general purpose worker which does not have any external dependency that is not critical to controlling the flow of your pipeline.
+
+If you want to have other dependencies available in your worker execution environment (and available in `brigade.js`), there are multiple approaches:
+
+- by creating a custom worker container image, which has your dependencies. This approach is [described in detail in this document](workers.md). In a nutshell, use this approach if you have the same dependency for multiple projects, or if your dependencies take a long time to pull.
+
+- without creating a custom container image:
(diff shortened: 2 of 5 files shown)


dgkanatsios:
Manually tested with your repo [here](https://github.com/radu-matei/brigade-javascript-deps), works like a charm. Only one code question, else LGTM.

brigade-worker/src/require.ts
```diff
@@ -16,10 +18,20 @@ class Pkg {
...
+export function getOverriddenPackage(pkg: string): string {
   for (let p of pkgOverrides) {
     if (p.name == pkg) {
-      return require(p.override);
+      return p.override;
+    }
+
+    // we want to intercept the loading of relative modules in the repo
+    // and and override them with the absolute path inside the worker pod
+    if ((!path.isAbsolute(pkg)) && (pkg.includes("./"))) {
```
dgkanatsios: shouldn't it be `path.startsWith`?

radu-matei: Thanks for catching that, updating.

radu-matei: Although on second thought, here's a scenario I haven't tested yet - 
In the context of #766, with `brigade.js` is in a custom location, say in a directory called `brigade`, and I want to load a module from `../js/my-module`, both our assumptions fail:

- the path does not start with `./`;
- the result of that path manipulation is not correct - it becomes `"./vcs/js/my-module"`, which is not the correct absolute path.

dgkanatsios: hm.... trying to think of something else as well. brigade.js references fileA.js, which references fileB.js. Do we need to do any changes in fileA or fileB?

radu-matei:
I honestly don't have a preference for that.
I was thinking, because I'd like the doc to cover importing both local and NPM packages, to have it as a separate PR, in a more tutorial style.

WDYT?

vdice:
Separate PR makes sense, especially w/ the extended examples you've mentioned.  👍

radu-matei:
@vdice, in its current state, this PR does not allow importing local modules when `brigade.js` is in a relative directory  (context of #766) - mainly because it involves resolving two levels of relative paths:
- the path of `brigade.js` relative to the repo root
- the path of every imported local module relative to _both_ the repo root and `brigade.js`

Given that having `brigade.js` in a custom location is a corner case, do we want to merge this PR in its current form, mark it as experimental, and clearly state the described scenario is not supported?

vdice:
@radu-matei Yes, I think that sounds fair for now.  Perhaps the disclaimer can mention that only local modules existing on same directory level as the `brigade.js`file are currently supported -- then, we'd be supporting the custom brigade.js location as well (or could, with an updated bit of logic to pull in custom path prefix if set).

docs/topics/dependencies.md
```diff
@@ -0,0 +1,123 @@
...
+[brigade:k8s] Creating PVC named brigade-worker-01cexrvrs08shcev26961cwd6n
+is 13 thirteen? true
+[brigade:app] after: default event handler fired
+[brigade:app] beforeExit(2): destroying storage
+[brigade:k8s] Destroying PVC named brigade-worker-01cexrvrs08shcev26961cwd6n
+```
+
+Notes:
+
+- the dependencies should point the exact version (and not use the tilde ~ and caret ^ to indicate semver compatible versions)
```
vdice: Is `should` the right word here -- or is `must` the right one?  (Basically, will Brigade's dep fetching break down if an exact version is _not_ supplied?)

vdice: typo: missing `to` in  `...point the exact version...`

docs/topics/dependencies.md
```diff
@@ -0,0 +1,123 @@
...
+
+- when adding a custom dependency using `brigade.json`, `yarn` will add it side-by-side with [the worker's 
+dependencies](../../brigade-worker/package.json) - this means that the process will fail if a dependency that conflicts with one of the 
+worker's dependency is added. However, already existing dependencies (such as `@kubernetes/client-node`, `ulid` or `chai`)
```
vdice: typo: pluralization needed, should read `... worker's dependencies...`

vdice: Curious -- what does the failure look like to the user?  Will it be obvious?

docs/topics/dependencies.md
```diff
@@ -0,0 +1,123 @@
...
+- as the Brigade worker is capable of modifying the state of the cluster, be mindful 
+of any external dependencies added through `brigade.json`
+
+- for now, the only part of `brigade.json` used is the `dependencies` object - it means the rest of the file 
+can be used to pass additional information to `brigade.js`, such as environment data - however, as the `brigade.json` 
+file is passed in the source control repository, information passed there will be available to anyone with access to the repository.
+
+- all dependencies added in `brigade.json` are dynamically installed on every Brigade build - this means if the dependencies added 
+are large, and the build frequency is high for a particular project, it might make sense to make a pre-built Docker image that 
+already contains the dependencies, as described in the sections below.
```
vdice: update needed here -- no longer `described in the sections below`, but back in the `workers.md` doc that we'd linked to above...

docs/topics/dependencies.md
radu-matei: It's a _must_, thanks for catching this!

docs/topics/dependencies.md
radu-matei: Hm, that's a very good point - let me open a new issue for that. Thanks!

radu-matei:
Thanks a lot for the reviews!

PR #791, opened 2019-01-31
docs(topics): add testing doc

vdice:

Adds Brigade testing examples per https://github.com/Azure/brigade/issues/451


vdice:
Hoping to get @blimmer 's thoughts on these initial docs as well, since his handy [Brigade Project Integration Test](https://github.com/blimmer/brigade-project-integration-test) is featured :)

PR #797, opened 2019-02-05
ref(github-gw): remove github-gateway in favor of brigade-github-app

vdice:

Per https://github.com/Azure/brigade/issues/746, removes github gateway code in favor of using [Azure/brigade-github-app](https://github.com/Azure/brigade-github-app) by way of an opt-in (disabled by default) sub-chart (See corresponding brigade-charts PR in https://github.com/Azure/brigade-charts/pull/22).

Functionality-wise, this PR should be ready; ~~just keeping in WIP while reviewers contribute feedback (plus TODOs below).~~

This was tested in combination with https://github.com/Azure/brigade-charts/pull/22 and running the following (on minikube cluster):
```console
$ make docker-buid
$ make helm-install
$ make test-functional
```

TODO ~~(leaving until this approach gains consensus)~~:
 - [x] update documentation
 - [x] address `TODO`s (in `Makefile`)
 - [x] test to see if backwards compatible (though not blocking if it is not, as we are pre-1.0; might just need disclaimer(s))

- Although use of the [Azure/brigade-github-app](https://github.com/Azure/brigade-github-app) is fairly proven, consider the case of an existing Brigade setup with multiple (OAuth-based) projects and its migration strategy -- if any are needed -- with these changes.  Again, since the time is now to make breaking changes pre-1.0,  perhaps this isn't crucial.
- Assuming this PR gains approval, think about ticket(s) we can add to [Azure/brigade-github-app](https://github.com/Azure/brigade-github-app) for improvments -- thinking improving the flow of GH App creation, functional tests, etc.

**If applicable**:


Makefile
```diff
@@ -100,6 +103,8 @@ test-unit: vendor
 	go test -v ./...
 
 # Functional tests assume access to github.com
+# and `brigade-github-app.enabled=true` set in the Brigade chart
```
radu-matei: Should this be `--set brigade-github-app.enabled=true` ?

radu-matei:
Just wondering if it made sense to move a copy of the old gateway over in the `brigadecore` org, so if there is anyone that still needs it they can use it.

It would move the gateway directory, together with the associated `pkg` files.

What do you think?

Makefile
vdice: Yes, that would be more explicit/helpful; I'll update.

vdice:
> Just wondering if it made sense to move a copy of the old gateway over in the `brigadecore` org, so if there is anyone that still needs it they can use it.

We could do this.  Alternatively, as we'll continue to host the last most recent artifacts for the classic gateway via the respective `v0.20.0`-tagged images (https://cloud.docker.com/u/deis/repository/docker/deis/brigade-github-gateway) and Brigade chart that consumes them (`helm install -n brigade brigade/brigade --version 0.20.0`), I wonder if we just add documentation on how to run with the older gateway (using these versions of the chart/image).

Or, compromising with both thoughts, we could continue to keep the classic gh gateway resources in the brigade chart, but toggled via a flag (with deployment pinned to last most recent image version, v0.20.0, as mentioned above).

radu-matei:
Currently moving all the related source code into a separate repo, under the new org.
This way, anyone who wants to continue to use / support / extend this gateway could. I'll update this comment with the PR on that repo once it's up.

vdice:
Re: testing and backwards compatibility, I'm hoping the approach mentioned in https://github.com/Azure/brigade-charts/pull/22#issuecomment-461673875 will be a good solution for supporting legacy Brigade installations while encouraging the switch to the newer-better-faster-stronger Brigade GitHub App for GitHub hooks when users are ready (documentation already has been updated with this approach the default).

How does this look to others?

vdice:
Just finished another round of sanity-check testing, including:

1. Upgrading from current brigade-0.20.0 chart to the brigade-0.21.0 chart in https://github.com/Azure/brigade-charts/pull/22 with the default/classic GitHub gateway (`gw`) enabled.  Confirmed persistence of this gateway through upgrade and continued webhook delivery.

2. Upgrading from brigade-0.20.0 chart to the brigade-0.21.0 without any custom values set.  Confirm that the default/classic GitHub gateway resources do not persist through upgrade as the default starting in 0.21.0 is _disabled_ by default.

3. Upgrading from brigade-0.20.0 chart to the brigade-0.21.0 with default/classic GitHub gateway disabled but the GitHub App gateway enabled.  Confirmed GitHub App webhook functionality continues to work though upgrade.

I'm thinking we're good to move this through.

PR #803, opened 2019-02-08
[WIP] Add ARM64 support

### LINKED ISSUE #773: Provide ARM images
I would like to run brigade on my cluster running on arm64 nodes. Looking at the makefile, it seems that the project is made to support building for multiple architectures, would it be possible to release images for arm64 as well?

Alxandr:
**What this PR does / why we need it**: It adds ARM64 support to brigade, allowing it to run on ARM (and hybrid ARM64/AMD64) clusters.

**If applicable**:

No code changes have been performed, so it should not need unit tests, nor break backwards compatibility. It might warrant an update to docs just to say it supports ARM though?

Resolves #773


Alxandr:
@radu-matei I added building of multi-arch manifest. 

image

https://cloud.docker.com/repository/docker/alxandr/brigade-generic-gateway/tags

radu-matei:
Question: I'm getting some warnings while building the worker:

```
Step 5/6 : RUN yarn install && yarn build
 ---> Running in 3301a56a3fa4
Unknown host QEMU_IFLA type: 40
Unknown host QEMU_IFLA type: 41
Unknown host QEMU_IFLA type: 40
Unknown host QEMU_IFLA type: 41
Unknown QEMU_IFLA_INFO_KIND ipip
Unknown host QEMU_IFLA type: 40
Unknown host QEMU_IFLA type: 41
Unknown QEMU_IFLA_INFO_KIND ip6tnl
Unknown host QEMU_IFLA type: 40
Unknown host QEMU_IFLA type: 41
yarn install v1.12.3
Unknown host QEMU_IFLA type: 40
Unknown host QEMU_IFLA type: 41
Unknown host QEMU_IFLA type: 40
Unknown host QEMU_IFLA type: 41
Unknown QEMU_IFLA_INFO_KIND ipip
Unknown host QEMU_IFLA type: 40
Unknown host QEMU_IFLA type: 41
Unknown QEMU_IFLA_INFO_KIND ip6tnl
Unknown host QEMU_IFLA type: 40
Unknown host QEMU_IFLA type: 41
```

However, they don't seem to directly affect building the image. Do we at least know what is causing them?

Alxandr:
I have the same warnings, and no. I'm not quite sure. As you said, they don't seem to be affecting the image. Google has been most unhelpful. What operating system are you running on? I was hoping that building on linux might not give this error (or a different one).

That being said, thinking about it (while typing); I would guess the issue is yarn downloading native (pre-built) dependencies of some sort?

radu-matei:
I'm currently testing on Windows, but in a couple of hours I can also test on Linux and macOS.
Yeah, it appears it has to do with pulling the Node packages - as long as the resulting worker image is functional, (and we open an issue to remember to actually track this at some point), I'm fine with merging.

Alxandr:
I need to actually test that the resulting arm images work first. I'm looking at the helm chart to see what values I need to modify to deploy to my cluster.

radu-matei:
I tested the build, build images, push images and push manifest bits of this PR.
Let us know how actually deploying Brigade on an ARM cluster goes, and I'd also like to ping @vdice and/or @adamreese since this PR pokes at the Makefile.

Thanks!

Alxandr:
Also on the TODO: the brigade.js file might need changing. I've not really looked at it though, but I assume it also builds the project?

radu-matei:
Yeah, and all releases are actually done through Brigade, so that will also need updating.

Alxandr:
I'm getting the following error from vacuum:

```
/usr/bin/brigade-vacuum: line 1: syntax error: unterminated quoted string
```

Any idea what could be causing this?

Alxandr:
Figured it out. I was putting the amd64 binary in the arm image -.-

I can now run jobs successfully:
https://asciinema.org/a/Y8rreMLgk7Voo48CAt2pkfBG9

radu-matei:
@vdice - could you have a look at this when you get a couple of minutes?
Also, the question about the changes in `brigade.js` seem like right up your alley :joy: 

Thanks!

Makefile
```diff
@@ -49,14 +50,26 @@ docker-build: build-docker-bins
...
-	docker push $(DOCKER_REGISTRY)/$*:$(IMAGE_TAG)
+	docker push $(DOCKER_REGISTRY)/$*:amd64-$(IMAGE_TAG)
+	docker push $(DOCKER_REGISTRY)/$*:arm64v8-$(IMAGE_TAG)
+
+	# Manifest creation
+	@if [ -d $(HOME)/.docker/manifests/docker.io_$(DOCKER_REGISTRY)_$*-$(IMAGE_VERSION) ]; then \
+		rm -rf $(HOME)/.docker/manifests/docker.io_$(DOCKER_REGISTRY)_$*-$(IMAGE_VERSION); \
+	fi
+
+	docker manifest create $(DOCKER_REGISTRY)/$*:$(IMAGE_TAG) $(DOCKER_REGISTRY)/$*:amd64-$(IMAGE_TAG) $(DOCKER_REGISTRY)/$*:arm64v8-$(IMAGE_TAG)
```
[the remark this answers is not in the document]
vdice: This manifest feature is really cool -- thanks for sparking my learning on it.

Although the manifest creation/annotation/push works for my Dockerhub and ACR registries, I see the following when attempting to push to my `quay.io` registry:
```
$ make docker-push
...
# Manifest creation
docker manifest create quay.io/vdice/brigade-api:v0.20.0-15-gc1af200 quay.io/vdice/brigade-api:amd64-v0.20.0-15-gc1af200 quay.io/vdice/brigade-api:arm64v8-v0.20.0-15-gc1af200
unsupported manifest media type and no default available: application/vnd.docker.distribution.manifest.v1+prettyjws
make: *** [brigade-api-push] Error 1
```

Wondering if there might be other registries where manifest is not yet supported?  (I'll dig into Quay.io's status here.)

If so, I'm wondering if we can come up with an approach where `docker-push` will still work for users pushing to registries without manifest support?

Makefile
```diff
@@ -40,7 +40,8 @@ $(BINS): vendor
 build-docker-bins: $(addsuffix -docker-bin,$(BINS))
```
vdice: Do we want to build `linux-arm64` `brig` binaries as well?  If so, perhaps we add a section in the `%-cross-compile` target along the lines of:
```
if [ $$os = 'linux' ]; then \
  GOOS=$$os GOARCH=arm64 CGO_ENABLED=0 go build -ldflags '$(LDFLAGS)' -o ./bin/$*-$$os-arm64 ./$*/cmd/$*; \
fi; \
```

to build this arch along with the default/amd64?

Makefile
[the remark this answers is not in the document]
Alxandr: > To invoke `docker manifest` commands, I had to add `"experimental": "enabled"` to my cli config at `~/.docker/config.json` -- we might want to add a comment/note here or somehow inform the user if this target errors out due to the lack of this flag in the user's cli config.

This is fairly easily detected and automatically ajusted. It would even be possible to set experimental: true, run the manifest commands, then set it back to what it used to be using `jq` for instance. I build some other multi-arch images on travis/azure pipelines using this technique for instance.

> Although the manifest creation/annotation/push works for my Dockerhub and ACR registries, I see the following when attempting to push to my `quay.io` registry:

I thought brigade was just using docker hub, so I didn't think the manifest usage would be a problem. However, achieving support for quay etc. should be pretty straight forward, just use a toggle `USE_MANIFEST` or whatever, and modify what push does. I can definitely add this to the makefile. Wrt quay specifically, they are apparently working on schema v2 support: https://github.com/moby/buildkit/issues/409#issuecomment-394757219, but no updates (at least in that thread) for months.

For the simplest solution, simply not creating/pushing the manifest lists would enable quay support. It would however force people to modify helm chart values before installing, but I guess that would be the case regardless, given that they are using their own registry.

Makefile
Alxandr: Do you mean in the `$(BINS): vendor` target?

vdice: Oh good clarification... I was thinking the `%-cross-compile` target.  It's not super clear in the Makefile, but it's only currently used for building the brig cli binaries (which we upload with the GitHub release in CI on a tag push).  It's called by:
```
.PHONY: build-release
build-release: brig-cross-compile
```

vdice: But only really necessary if we think the new `arm64`-variant `brig` binaries might be useful and should join the pre-existing amd64 linux, windows and darwin binaries.

Makefile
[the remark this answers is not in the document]
vdice: > For the simplest solution, simply not creating/pushing the manifest lists would enable quay support. It would however force people to modify helm chart values before installing, but I guess that would be the case regardless, given that they are using their own registry.

Good point -- perhaps, if the user needs to update tag values anyways, supporting quay should be as straightforward as configuring the manifest commands to run optionally.  (Thanks for that related issue link.)

I'd be fine with either the `USE_MANIFEST` approach for conditionalizing the push command or with `%-docker-push` only issuing `docker push` commands and maybe a separate `%-docker-manifest` command to wrap the manifest commands.  Then, CI (or anyone) can run `make docker-push docker-manifest` to push and create/update manifests for all images.  Will leave it up to you.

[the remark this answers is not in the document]
vdice: > This is fairly easily detected and automatically ajusted. It would even be possible to set experimental: true, run the manifest commands, then set it back to what it used to be using `jq` for instance. I build some other multi-arch images on travis/azure pipelines using this technique for instance.

I'd be down to add this logic as a special target, for use in CI.  Not sure I'd want it to run automagically by default on user's machines, though.

Alxandr:
@vdice I did a major overhaul of the makefile. Targets that are based on files (like how make is supposed to work) as far as is possible etc. I've not tested the manifest creation yet, but building single or multiple arches works. The dockerfiles now also have arguments, such that they will not go out of sync. However, as a downside, the makefile is now harder to understand (though I guess that's subjective). Please take a look at it and see if you like it, if not I can go back to scratch and see if it can be done in another way. Though considering how you're using make (all of the targets are `.PHONY`), switching to another, easier to handle task runner might be a suggestion?

Makefile
```diff
@@ -16,75 +15,141 @@ BRIGADE_GITHUB_GW_EXTERNAL_PORT := 7744
...
+arm64_DOCKER_ARCH = arm64v8
+arm32_DOCKER_ARCH = arm32v6
+
+amd64_DOCKER_ANNOTATIONS = --os linux --arch amd64
+arm64_DOCKER_ANNOTATIONS = --os linux --arch arm64 --variant v8
+arm32_DOCKER_ANNOTATIONS = --os linux --arch arm
+
+.PHONY: build build-images push-images dockerfiles
+
+define GO_CMD_TARGETS
```
vdice: I'm thinking some further comments around each of the `define` sections would help new users -- and even us after time goes by! -- more quickly grok how it all ties together.

For example, maybe we add, godoc style :
```
# GO_CMD_TARGETS generates go build targets using the provided 
# Brigade component ($1) and architecture ($2), e.g., brig/rootfs/amd64/rootfs
#
# See the `foreach` loop that invokes this function - elsewhere in this file - for how these targets are generated
define GO_CMD_TARGETS
...
```
Definitely open as to wording, but at a high-level being sure to mention what each arg (`$1`, `$2`, etc.) corresponds to and a literal example.

Makefile
```diff
@@ -16,75 +15,141 @@ BRIGADE_GITHUB_GW_EXTERNAL_PORT := 7744
...
+bin/$3-$1-$2: vendor $(shell find $3 -type f -name '*.go')
+	GOOS=$1 GOARCH=$$($2_GOARCH) CGO_ENABLED=0 go build -ldflags '$(LDFLAGS)' -o ./$$@ ./$3/cmd/$3
+endif
+endef
+
+define BRIG_STATIC_TARGETS
+$(foreach cmd,$(BINS),$(eval $(call BRIG_STATIC_ARCH_TARGETS,$2,$1,$(cmd))))
+endef
+
+define GO_ARCH_TARGETS
```
vdice: I'm wondering if we move lines 128-140 further up at the top, before the definitions of the functions that are invoked in each `foreach` statement (`BRIG_STATIC_TARGETS`, etc.).

Then the flow to understand would be: seeing these meta targets and foreach loops, then digging in to each looped function for how each parameterized target is built...

Makefile
```diff
@@ -16,75 +15,141 @@ BRIGADE_GITHUB_GW_EXTERNAL_PORT := 7744
...
+push-$2-images push-$1-images: push-$1-$2-image
+endef
+
+define DOCKER_MANIFEST_TARGETS
+.PHONY: create-$1-manifest $1
+
+$1: $1-images
+
+create-$1-manifest: push-$1-images
+	# Remove old manifest if present
```
vdice: Are there any drawbacks in removing the need for this cleanup logic and instead adding `--amend` to the `docker manifest create`?

I explored this option as the cleanup logic broke down for me when my registry was a non-docker.io registry (ACR) -- so the manifest file was named differently (`$(DOCKER_REGISTRY)_$1-$(IMAGE_VERSION)`).  Instead of updating the cleanup logic, I thought it'd be great if we could get rid of it altogether if `docker manifest create --amend ...` would work for us.

Makefile
```diff
@@ -16,75 +15,141 @@ BRIGADE_GITHUB_GW_EXTERNAL_PORT := 7744
...
+
+amd64_DOCKER_ARCH = amd64
+arm64_DOCKER_ARCH = arm64v8
+arm32_DOCKER_ARCH = arm32v6
+
+amd64_DOCKER_ANNOTATIONS = --os linux --arch amd64
+arm64_DOCKER_ANNOTATIONS = --os linux --arch arm64 --variant v8
+arm32_DOCKER_ANNOTATIONS = --os linux --arch arm
+
+.PHONY: build build-images push-images dockerfiles
```
vdice: Should we add a top-level `manifests` target as well?  (Or, `update-manifests`, etc., to follow pattern of `<verb>-<noun>`)

Then, we could add the following in the `DOCKER_MANIFEST_TARGETS` function:
```
manifests: create-$1-manifest
```

Flow for CI would then be: `make build-images push-images manifests` for building/pushing/updating manifests.

git-sidecar/Dockerfile.arm64v8
```diff
@@ -0,0 +1,13 @@
+FROM arm64v8/alpine:3.8
```
vdice: Was there a reason to keep this arch-specific Dockerfile around?  Or can it be removed now that they are dynamically generated?

Makefile
```diff
@@ -16,75 +15,141 @@ BRIGADE_GITHUB_GW_EXTERNAL_PORT := 7744
...
+$1-$2-image: $1/Dockerfile.$2
+	docker build $(DOCKER_BUILD_FLAGS) -t $(DOCKER_REGISTRY)/$1:$2-$(IMAGE_TAG) -f $1/Dockerfile.$2 $1
+
+# Push image for a single arch
+push-$1-$2-image: $1-$2-image
+	docker push $(DOCKER_REGISTRY)/$1:$2-$(IMAGE_TAG)
+
+# Create arch dockerfile
+$1/Dockerfile.$2: $1/Dockerfile
+	@cat $1/Dockerfile | ARCH="$2" QEMU_IMPORT="$$($2_QEMU_IMPORT)" DOCKER_ARCH="$$($2_DOCKER_ARCH)" envsubst '$$$${ARCH} $$$${QEMU_IMPORT} $$$${DOCKER_ARCH}' > $1/Dockerfile.$2
```
vdice: We'll need to check for the existence of `envsubst` and either install in a platform-agnostic manner if it doesn't exist -- or exit out and inform the user that it is a required dependency to run the targets needing it.

As a Mac user, I didn't have it by default, although I did have the package that contained it (`gettext`).  For me, the key was the latter command from: `brew install gettext && brew link --force gettext`

Makefile
```diff
@@ -16,75 +15,141 @@ BRIGADE_GITHUB_GW_EXTERNAL_PORT := 7744
...
+$(foreach img,$(IMAGES),$(eval $(call DOCKER_MANIFEST_TARGETS,$(img))))
 
 .PHONY: helm-install
 helm-install: helm-upgrade
 
 .PHONY: helm-upgrade
 helm-upgrade:
 	helm upgrade --install $(BRIGADE_RELEASE) brigade/brigade --namespace $(BRIGADE_NAMESPACE) \
-		--set brigade-github-app.enabled=true \
+		--set gw.service.type=ClusterIP \
```
vdice: I'd like to go back to how master looks for this target (`gw.*` sets removed in place of `--set brigade-github-app.enabled=true`), as we hope to deprecate the classic GH gateway (`gw`) in favor of the `brigade-github-app` gateway.

Makefile
```diff
@@ -16,75 +15,141 @@ BRIGADE_GITHUB_GW_EXTERNAL_PORT := 7744
...
+
+amd64_DOCKER_ANNOTATIONS = --os linux --arch amd64
+arm64_DOCKER_ANNOTATIONS = --os linux --arch arm64 --variant v8
+arm32_DOCKER_ANNOTATIONS = --os linux --arch arm
+
+.PHONY: build build-images push-images dockerfiles
+
+define GO_CMD_TARGETS
+# Build command
+$1/rootfs/$2/$1: vendor $(shell find $1 -type f -name '*.go')
```
vdice: I must admit I'm still trying to figure out _where_ the change is needed :) but I'm seeing these (docker binary) image targets being called when I run `make build`.

I think I prefer maintaining the pattern of `make build` only building go binaries (and placing into `/bin`) corresponding to the os+arch of the host machine.

Then, these docker binary build targets (putting the linux+<arch> binaries in <component>/rootfs) would only be run, as needed, by `make build-images` (and/or, if we had the need to run separately, e.g. via `make build-docker-bins`)

Makefile
Alxandr: Honestly, I don't know. The docker manifest commands are so badly documented, that I'm not even sure what they really do. I've just found *a* way to use them that works 😛. My main issue with doing amend would be that you would possibly just elongate the manifest list in repeated builds. This *shouldn't* be a problem as long as the manifest tag and the image tags are the same, but I would have to test it. Also, when doing `manifest push`, we can add  `--purge` (with the option to disable it for debugging purposes), which will purge after pushing.

git-sidecar/Dockerfile.arm64v8
Alxandr: Oversight. Will remove it.

Makefile
Alxandr: Yeah, the names should probably be walked over (put in a list, drawn in a tree, and then painted on a wall a few times, to get a good feel for them 😛). I spent quite a while on this, re-discovering make while doing it (been a while since I did a make template), so my state of mind probably changed quite a bit while writing it, leading to inconsistencies. Doing a dependency-graph in comments at the top of the file would probably be a good idea.

Makefile
Alxandr: I did the same command :). Alternatively, using something like go templates might be a better idea?

Alxandr: Just from half a minute of googeling, `go get github.com/hairyhenderson/gomplate/cmd/gomplate` maybe?

Makefile
Alxandr: This was probably just a merge conflict. Pretty sure I didn't intentionally touch the helm-upgrade target.

Makefile
Alxandr: Fair enough. I'll try drawing a dependency graph with the names. Then we can discuss how it should work after it's more visible. Makes it easier to make naming changes like this.

Makefile
Alxandr: I'm an F#-er, so I'm used to reading files with the exports at the bottom, and the implementation details at the top, hence the ordering (up to a point where I confused myself I think). Will try to move targets that are more likely to be called higher up.

Makefile
vdice: Nice -- if that utility works, a go get-able option would be a good choice.

vdice:
Hi @Alxandr , an update from today's (2/19/19) Brigade Dev Meeting.  Although we/I will still continue to test/review the functionality presented in this pull request, the main concern at this time is to reduce the potential for speed bumps/issues in releasing Brigade 1.0.0.  As the Makefile has been fully overhauled and the core maintainers have yet to fully understand its new design/mechanisms, the expectation is that the earliest this work would be merged would be after this milestone.

Alxandr:
Fair enough. I have things running on my cluster, which was my main concern. I have one issue though, which I would probably label as a "bug" (but would not matter if this branch got merged): I need to modify the worker of every new project I make, even though I set the worker in the values for the helm template. Other than that, things are working just as they should on my end.

odidev:
@vdice 

I am also interested in Linux/ARM64 Docker image release for ‘brigade’.

According to the previous comments, it seems that the Linux/ARM64 docker image addition was in your backlog, but due to v1.0.0 milestone, the activity was on-hold.

May I know why this activity is still on-hold? 
If you are still interested in this PR, can you please test and review the changes?

Kindly let me know if there is anything I can do, I will be happy to help and contribute.

krancour:
@odidev I cannot speak directly to why this PR stalled out two years ago. I _can_ tell you that, currently, all our available resources are being directed at Brigade 2.0.

We would _love_ to add ARM64 support and It's certain to happen in some version 2.x > 2.0.

I'm also certain the state of the art in building multi-architecture Docker images has advanced in the two years since this PR was touched.

Assistance is welcome with multi-arch builds for 2.0. See the v2 branch.

radu-matei:
One issue was that we didn't have the proper infrastructure for CI - and I'm still not sure if we have now.

I believe @squillace may have some more info around ARM CI?

Alxandr:
I've (personally) moved from ARM back to X86, so I don't really have a stake in this anymore, but I did keep up to date on multi-arch images etc, so I can say that there are several new(ish) things there that make things way easier than when I started on that journey (I don't remember where in the spectrum this image falls). For anyone who wants to look at this, I recommend looking at "docker buildx bake". I will close this PR as I have no idea what's in it anymore :P.

odidev:
@radu-matei  @krancour  @Alxandr 

Thank you for the clarifications.

I tried building the package (v2 branch) locally on Linux x86_64 and ARM64 machines using ‘make’ command. The build was successful on the x86_64 machine, and amd64 binaries were generated in the ‘bin’ folder. 
But the build failed on a Linux/ARM64 machine as qemu was unable to detect the host platform (linux/arm64/v8), as shown below:
~~~
qemu: uncaught target signal 11 (Segmentation fault) - core dumped
~~~

May I know if you are planning to use Github actions to include the linux jobs to the CI in the development branch ‘v2’? If yes, then it will be feasible to add the Linux/ARM64 docker image build and release process via GA CI.

Also, are you planning to add Linux/ARM64 support in the 2.0 release? If yes, then I will be happy to contribute for the ARM64 docker image release process.

Kindly share your opinion on this.

krancour:
> May I know if you are planning to use Github actions to include the linux jobs to the CI in the development branch ‘v2’?

We have used GitHub Actions for only one thing-- building the Windows log agent image. And the reason we're doing that is because building a Window-based Docker image _inside_ another Windows-based Docker container does not seem to be possible or practical yet. With this one exception, we're avoiding GitHub Actions for the sake of "dogfooding" Brigade. Our CI and CD are, and will remain, based on Brigade. (Full transparency, we're currently using Brigade 1.4.something to build the `v2` branch, but that will change as other things, like the GitHub gateway, in the Brigade v2 "orbit" start to mature.)

> Also, are you planning to add Linux/ARM64 support in the 2.0 release?

Reiterating what I already said:

> We would love to add ARM64 support and It's certain to happen in some version 2.x > 2.0.

Also, @radu-matei has raised an excellent point that we may still be lacking the infrastructure to accomplish this. i.e. I'm not sure we have access to ARM64 VMs right now.

odidev:
> Also, @radu-matei has raised an excellent point that we may still be lacking the infrastructure to accomplish this. i.e. I'm not sure we have access to ARM64 VMs right now.

> One issue was that we didn't have the proper infrastructure for CI - and I'm still not sure if we have now.
> 
> I believe @squillace may have some more info around ARM CI?

@krancour @radu-matei , Can't we use [Travis-CI](https://docs.travis-ci.com/user/multi-cpu-architectures/) for the same as it has support for multi-architecture builds.

krancour:
@odidev While Brigade isn't exclusively for CI, CI is a popular use case for Brigade. To the greatest extent we're able, we prefer to implement our CI by "dogfooding" our own product. I think I've previously indicated this preference for dogfooding. For the time being, using Travis or Circle or similar is off the table.

I can see you're keen to get ARM64 support. You're welcome to explore whatever options may exist that don't violate our preference for dogfooding.

squillace:
Per https://github.com/brigadecore/brigade/pull/803#issuecomment-779559533, @radu-matei, I _will_ insist on ARM outputs at some point, and I will arrange to pay for them as well. Lemme look into this more deeply. (I'm on Team Pinebook over here.)

As to anything but Brigade in the Brigade automation, _I'm completely on Team Dogfood._ FWIW and if anyone cares.

odidev:
@krancour , I've built the package on my local Arm64 system. It is getting built successfully there but to build this we need the Docker images of `"brigadecore/kaniko"`, `"brigadecore/helm-tools"` & `"brigadecore/go-tools"` packages for Arm64 platform at Dockerhub. I've raised issues regarding the  same in respective packages.  Can you please look into those issues?

Issue #804, opened 2019-02-12
Generic Gateway always try to connect to api.github.com


floriankoch:
Hi,

brigade version 0.20.

I try to use the GenericGatweay with an internal bitbucket repository.
I can Connect to the GenericGateway via the Connection test
```
[GIN] 2019/02/12 - 06:22:53 | 400 |    3.032987ms |     10.63.12.98 |  POST     /webhook/brigade-2fd9fa361b1333a8005735ac181a630b0a6bb96fc4f88b9bac93f8/abcdef                                                            
Failed to convert POST data into JSON: EOF
```

With a real commit, i got this error:
```
[GIN] 2019/02/12 - 06:22:26 | 200 |    5.449499ms |     10.63.12.98 |  POST     /webhook/brigade-2fd9fa361b1333a8005735ac181a630b0a6bb96fc4f88b9bac93f8/abcdef                                                            
Error getting file: GET https://api.github.com/repos/k8s/brigade-test/contents/: 404 Not Found []
failed genericWebhook event: Error getting file: GET https://api.github.com/repos/k8s/brigade-test/contents/: 404 Not Found []   
```

Does GenericGateway only supports github?

dgkanatsios:
thanks for reporting that, we'll take a look :)

dgkanatsios:
so, yes, Generic Gateway currently supports GitHub only.
The way we're using it is to get the contents of the "brigade.js" file in order to trigger a new Build.

I've been checking on how to implement BitBucket as well (and possibly other providers)
- currently we're using https://github.com/google/go-github which supports only GitHub
- https://github.com/hashicorp/go-getter could be an option but it requires client to have Git installed (we could do that at the container level, not sure if it's desirable)
- there's also [this](https://github.com/ktrysmt/go-bitbucket) library for BitBucket (but if we go down that path we would need to check if "GitHub"/"BitBucket"/"Gitlab") 
- there's also [this](https://github.com/src-d/go-git) Git library that we could use. I haven't tested it and I'm not sure how it would work with various authentication schemes, though.

Thoughts?

radu-matei:
I'm having trouble understanding why we have GitHub specific libraries for this - it should simply get the request from the webhook and create the Brigade build - and it should be possible to use it directly without even a repo.

PR #805, opened 2019-02-12
[docs] mention that Generic Gateway supports only GitHub 

dgkanatsios:
As per #804, we should mention that the only VCS currently supported by Generic Gateway is GitHub, for the time being.


dgkanatsios:
Closing as we'll follow an alternate route

PR #809, opened 2019-02-13
fix regex matching on the service account

### LINKED ISSUE #808: Default service account error
Testing against master (https://github.com/Azure/brigade/commit/5e2fbe537713e9196273d757eb3c3cbf46803979) in prep for cutting a beta `v1.0.0` release and hit the following.

I'm seeing pre-existing (default/unconfigured when it comes to service account) Brigade project builds error out with:
```
[brigade] Service Account brigade-worker does not match regex /brigade-worker/
```
It looks like the default service account is correct but that the regex matching needs fixing?

cc @rmb938

rmb938:
fixes #808


rmb938:
@vdice @radu-matei

Issue #798, opened 2019-02-06
[Design] CloudEvents support for Generic Gateway


dgkanatsios:
This is the design proposal for CloudEvents support for Generic Gateway, as discussed in #764 

Currently, Generic Gateway listens to "/webhook" for a simple JSON payload and will raise a "webhook" event that should be handled by brigade.js.

Generic Gateway could support latest version of CloudEvents [v0.2](https://github.com/cloudevents/spec/blob/v0.2/spec.md). As per proposed implementation, Generic Gateway could listen to "/cloudevent" for such an [event](https://github.com/cloudevents/sdk-go/blob/master/v02/event.go) and handle it via a "cloudevent" event handler on brigade.js. 

Some points:

1. Relevant Brigade data (e.g. `ref` and `commit`) could be stored on Event's [Data](https://github.com/cloudevents/sdk-go/blob/master/v02/event.go#L37) field. Brigade.js consumer should be able to see the entire payload, of course.
2 CloudEvents could support the existing security mechanism we already have for Generic Gateway (with a custom `secret` per Project)
3. It should be made clear in the documentation that Generic Gateway supports a Generic webhook endpoint as well as a CloudEvents endpoint. In this context, maybe rename the `Generic webhook` endpoint to `simple webhook`?
4. Brigade should have a dependency to `github.com/cloudevents/sdk-go/v02`

Thoughts?

radu-matei:
I think the design described here would be a great first iteration!

vdice:
My only thoughts are at a higher-level around the future of gateways and Brigade.  I'm curious if it would be desirable that any additional gateways besides "legacy" (container registry, generic) should exist in separate repos and be pluggable by way of sub-charts to the main Brigade chart?  

We're exploring this approach for the Github gateway as we transition to using a pluggable but separate brigade-github-app for this gateway functionality (and thus, pulling out the classic GH gateway code; see https://github.com/Azure/brigade/pull/797)

Each additional gateway being built as a standalone project might be preferable _if_ Brigade is in a position to support this (or could be with a small amount of work).  I suppose one drawback of standalone-but-composable gateways might be separate service IPs for each and the overhead of covering each w/ ingress/TLS?

However, as @dgkanatsios has the most recent experience with gateway work, I'd definitely defer to his experience and opinions around this idea.

radu-matei:
[...]

Issue #810, opened 2019-02-13
brig build list mentions wrong time for failed Builds


dgkanatsios:
Output of `brig version`: 0.20.0

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.2", GitCommit:"cff46ab41ff0bb44d8584413b598ad8360ec1def", GitTreeState:"clean", BuildDate:"2019-01-10T23:35:51Z", GoVersion:"go1.11.4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.4", GitCommit:"f49fa022dbe63faafd0da106ef7e05a29721d3f1", GitTreeState:"clean", BuildDate:"2018-12-14T06:59:37Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): AKS

brig build list command mentions wrong AGE for failed builds. This number should be something more relevant.

```
dgkanatsios@digkanat:~$ brig build list
ID                              TYPE            PROVIDER        PROJECT                                                         STATUS          AGE
01d3k1x8k82z44ctgfeg54vscc      simpleevent     GenericWebhook  brigade-31ac53f7aa2f4d06a6715d83a4b78db950c79d448bfb76e7eb8083  Succeeded       11m
01d3k1x2n1vp2fe30jax3e0t66      simpleevent     GenericWebhook  brigade-31ac53f7aa2f4d06a6715d83a4b78db950c79d448bfb76e7eb8083  Failed          292y
```

dgkanatsios:
Hm, `Age` mentions [PodEndTime](https://github.com/Azure/brigade/blob/master/brig/cmd/brig/commands/build_list.go#L112), I think it should be StartTime. I'll proceed with a PR.

Issue #564, opened 2018-07-19
Component that uses the Brigade kube store automatically needs access to pods


radu-matei:
A Brigade gateway (or any component that vendors `github.com/Azure/brigade/pkg/storage/kube`) that is deployed on a cluster with RBAC enabled needs access to pods, even if the said component might not need it.

That is because of the API cache (`github.com/Azure/brigade/pkg/storage/kube/apicache/cache.go`), which contains a pod store:

```
	return &apiCache{
		hasSyncedInitially: merged,
		client:             client,
		secretStore:        secretStoreFactory{}.new(client, namespace, resyncPeriod, secretsSynced),
		podStore:           podStoreFactory{}.new(client, namespace, resyncPeriod, podsSynced),
	}
```

If the role used for the deployment will not have access to pods (which might be the case for gateways, which only need to get and create secrets), this results in the logs being flooded with error logs about not having access to pods:

```
reflector.go:205] github.com/radu-matei/brigade-vsts-gateway/vendor/github.com/Azure/brigade/pkg/storage/kube/apicache/liststore.go:36: Failed to list *v1.Pod: pods is forbidden: User "system:serviceaccount:brigade:brigade-vsts-gateway" cannot list pods in the namespace "brigade"
```

Note that this does't crash the component, but in order to have usable logs, you need to grant the deployment access to pods as well.

Should there be a separation at the cache level for secrets and pods, and different components create the required cache store?

PR #811, opened 2019-02-13
Generic Gateway should not try and find a brigade.js file

### LINKED ISSUE #804: Generic Gateway always try to connect to api.github.com

dgkanatsios:
Closes #804 

At its current incarnation, Generic Gateway tries to pull brigade.js file from GitHub. As discussed in #804 and in last community call, it should not be the job of a gateway to add/edit the brigade.js file and should delegate this job to to the worker.
This PR fixes this.

Moreover, it was discussed in #798 that, since we'll also introduce CloudEvents support for the existing Generic Gateway, we should do a rename of the existing Generic Gateway endpoint. So, you will see that the endpoint for the Generic Gateway is now called `simpleevent` (was called `webhook`). The same name change was applied to the event triggered by the Generic Gateway. Necessary name refactors were introduced in the code well, but they should be transparent to the end user.

Last but not least, more accurate tests were added.

#804 was created because of an error with a private BitBucker repo, if any of the reviewers do have an account there, I'd appreciate a check.

**If applicable**:
- [X] this PR contains unit tests
- as noted, this PR **breaks** backwards compatibility with the previous endpoint/event of the Generic Gateway


floriankoch:
would be cool when this goes into 1.0.0

docs/topics/genericgateway.md
```diff
@@ -35,22 +36,24 @@ When you create a new Brigade Project via Brig CLI, you can optionally create su
...
-When calling the Generic Gateway webhook endpoint, you can include a custom JSON payload such as:
+When calling the Generic Gateway `simpleevent` endpoint, you must include a custom JSON payload such as:
 
 ```json
 {
 	"ref": "refs/heads/changes",
-	"commit": "b60ad9543b2ddbbe73430dd6898b75883306cecc"
+	"commit": "b60ad9543b2ddbbe73430dd6898b75883306cecc",
+  "key1": "value1",
+  "key2": "value2"
```
radu-matei: Badly formatted JSON here.

docs/topics/genericgateway.md
```diff
@@ -35,22 +36,24 @@ When you create a new Brigade Project via Brig CLI, you can optionally create su
...
 
-`Ref` and `commit` values would be used to configure the specific revision that Brigade will pull from your repository.
+`Ref` and `commit` values are used to configure the specific revision that Brigade will pull from your repository, if you're using one. If both values are missing, then your Build's `ref` will be set to `master`. If you do not wish to provide any payload, just include an empty JSON object (like `{}`).
 
 Last but not least, here is a sample Brigade.js file that could be used as a base for your own scripts that respond to Generic Gateway's `webhook` event. This script will echo the name of your project and 'webhook'.
 
 ```javascript
-events.on("webhook", (e, p) => {
+events.on("simpleevent", (e, p) => {
```
radu-matei: Formatting is a bit off here.

pkg/storage/kube/build.go
```diff
@@ -115,7 +115,7 @@ func (s *store) CreateBuild(build *brigade.Build) error {
 			"log_level":      build.LogLevel,
 		},
 	}
-
+	fmt.Printf("lala %s\n", string(secret.Data["payload"]))
```
radu-matei: Leftover from debugging?

pkg/webhook/genericsimpleevent.go
```diff
@@ -55,57 +50,46 @@ func (g *genericWebhook) Handle(c *gin.Context) {
...
-		c.JSON(http.StatusBadRequest, gin.H{"status": "Malformed body"})
+		log.Printf("Failed to convert POST data into JSON: %s", err)
+		c.JSON(http.StatusBadRequest, gin.H{"status": "Malformed POST data - Invalid JSON"})
 		return
 	}
-	defer c.Request.Body.Close()
 
-	go g.notifyGenericWebhookEvent(proj, body, gwData)
-	c.JSON(200, gin.H{"status": "Success"})
+	go g.notifyGenericWebhookSimpleEvent(proj, payload, revision)
```
radu-matei: Why is a separate Goroutine used here?

pkg/webhook/genericsimpleevent.go
```diff
@@ -55,57 +50,46 @@ func (g *genericWebhook) Handle(c *gin.Context) {
...
+	c.JSON(200, gin.H{"status": "Success. Build created"})
 }
 
-func (g *genericWebhook) notifyGenericWebhookEvent(proj *brigade.Project, payload []byte, gwData *genericWebhookData) {
-	if err := g.genericWebhookEvent(proj, payload, gwData); err != nil {
-		log.Printf("failed genericWebhook event: %s", err)
+func (g *genericWebhookSimpleEvent) notifyGenericWebhookSimpleEvent(proj *brigade.Project, payload []byte, revision *brigade.Revision) {
+	if err := g.genericWebhookSimpleEvent(proj, payload, revision); err != nil {
```
radu-matei: Why the indirection here?

pkg/webhook/genericsimpleevent.go
```diff
@@ -55,57 +50,46 @@ func (g *genericWebhook) Handle(c *gin.Context) {
...
 		Payload:   payload,
 		Revision:  revision,
-		Script:    script,
 	}
-	if proj.DefaultScript != "" {
-		b.Script = []byte(proj.DefaultScript)
+
+	// set a default Revision if user has not provided any information about commit or ref
+	if b.Revision == nil || (b.Revision.Commit == "" && b.Revision.Ref == "") {
+		b.Revision = &brigade.Revision{Ref: "master"}
```
radu-matei: Is this actually needed?

docs/topics/genericgateway.md
dgkanatsios: removing

pkg/webhook/genericsimpleevent.go
dgkanatsios: it's based on [this](https://github.com/Azure/brigade/blob/master/pkg/webhook/dockerhub.go#L63). I suppose the original intention is that we should return a 200 immediately to the caller, since kube API might delay the Secret creation. However, I don't see any harm in making it a synchronous call. WDYT?

pkg/webhook/genericsimpleevent.go
dgkanatsios: This follows the pattern [here](https://github.com/Azure/brigade/blob/master/pkg/webhook/dockerhub.go#L68). Indeed, if we decide to make the call synchronous, there's no need for that.

pkg/webhook/genericsimpleevent.go
dgkanatsios: I think so. If no Commit or Ref was specified, I had an issue on a GitHub repo I used (and this was the reason I encountered #810, hehe). Let me give it one more shot **without** including set to 'master' to make sure

dgkanatsios: Indeed, just tested with these lines commented out, running a build against `deis/empty-testbed`. Sidecar fails to initialize, with the following error
```
+ :
+ :
+ : /vcs
+ refspec=
+ cut -f2
+ git ls-remote --exit-code https://github.com/deis/empty-testbed.git
+ full_ref=
+ git init -q /vcs
+ cd /vcs
+ retry git fetch -q --force --update-head-ok https://github.com/deis/empty-testbed.git
+ local 'n=1'
+ local 'max=5'
+ local 'delay=5'
+ true
+ git fetch -q --force --update-head-ok https://github.com/deis/empty-testbed.git
+ break
+ retry git checkout -q --force
+ local 'n=1'
+ local 'max=5'
+ local 'delay=5'
+ true
+ git checkout -q --force
fatal: empty string is not a valid pathspec. please use . instead if you meant to match all paths
Command failed. Attempt 1/5. Waiting for 5 seconds before retrying.
+ test 1 -lt 5
+ echo 'Command failed. Attempt 1/5. Waiting for 5 seconds before retrying.'

...
sidecar retries but ultimately fails :(
```

Setting the `master` to the Ref did the trick. Can you also check, maybe fault is something in my configuration?

radu-matei: No, you are right - I think we should let it like this.

pkg/webhook/genericsimpleevent.go
radu-matei: Oh, yes, that makes sense - not entirely sure if we should directly return a 200, but it makes sense, and if we are to change the behaviour, we should do it at once, across gateways.

Marking this as resolved, thanks!

dgkanatsios:
a couple of comments before merging 
- Container Registry gateway might use goroutine to create the Build, but Github Gateway does not ([source](https://github.com/Azure/brigade-github-app/blob/master/pkg/webhook/github.go#L320)). Maybe we should open an issue to have a common approach?
- added a comment with the sidecar error when we set Ref as "master", for future reference

thanks!

radu-matei:
Yes, I think we should have another look at whether to wait before returning 200 when creating the build - but it shouldn't concern this PR.

Thanks!

radu-matei:
Got around to testing this one more time:

```
E0222 01:35:31.658531       1 reflector.go:205] github.com/Azure/brigade/pkg/storage/kube/apicache/liststore.go:36: Failed to list *v1.Pod: pods is forbidden: User "system:serviceaccount:brigade:brigade-brigade-generic-gateway" cannot list resource "pods" in API group "" in the namespace "brigade"
```

Essentially, we are getting hit by #564 - workaround is to give this component access to pods.

PR #817, opened 2019-02-22
Fix service account regex test

radu-matei:
This PR updates the service account regex check to use `RegExp.test` instead of `string.match`, and updates Brigadier to the latest version, v0.3.

Now the questions I've been going around for the last couple of days for this PR:

- whether the job object should actually expose the service account regex - I don't think it should ever be updated by `brigade.js`.
- would it be used elsewhere in the worker?

cc @vdice, @rmb938

### DIFF
--- a/brigade-worker/package.json
+++ b/brigade-worker/package.json
@@ -33,7 +33,7 @@
   },
   "dependencies": {
     "@kubernetes/client-node": "^0.7.2",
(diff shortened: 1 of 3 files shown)


radu-matei:
Decided to remove the job service account regex from the user-facing Job object and stick to the initial implementation model from @rmb938 -- only updating the regex test to use the JavaScript regex library -- as opposed to `string.match`.

Issue #764, opened 2019-01-09
CloudEvents support for generic gateway


radu-matei:
As #602 and #762 are getting worked on, one feature request we've been getting is support for [CloudEvents](https://cloudevents.io/).

Most of the work for this is already done in [the Brigade EventGrid gateway, which already has CloudEvents support](https://github.com/radu-matei/brigade-eventgrid-gateway/tree/master/pkg/cloudevents), but it would be nice for this to be supported out of the box by Brigade.

Are there any downsides to having support out of the  box?

cc @squillace

squillace:
None but resources. As the code is a good way there, this could be completed, but not as a part of 1.0 blocking list.

dgkanatsios:
Should we add another endpoint/URL on Generic Gateway that would accept CloudEvents?

radu-matei:
Yeah, I think that should be the way to go here.
But it shouldn't block #762 from getting merged in its current state.

dgkanatsios:
cool, I'm tempted to take a look :) Is there a go package that we can go get or just simple copy [this](https://github.com/radu-matei/brigade-eventgrid-gateway/blob/master/pkg/cloudevents/envelope.go)?

dgkanatsios:
Cool, I'm tempted to take a look as soon as we merge #762 so we have a baseline

radu-matei:
_As soon as we merge_ is now :joy: 
That sounds great, let me know if there is anything I can help with.

Thanks!

dgkanatsios:
thank you, I'll take a look during the next days and probably come up with a rough design here, so we can discuss and then start implement

radu-matei:
Closing as the design proposal for this moved to #798.

PR #818, opened 2019-02-22
CloudEvents 0.2 support

### LINKED ISSUE #798: [Design] CloudEvents support for Generic Gateway

dgkanatsios:
Closes #798 
This PR adds support for [CloudEvents 0.2](https://github.com/cloudevents/spec/blob/v0.2/spec.md) on the GenericGateway. It is using [official Go SDK](https://github.com/cloudevents/sdk-go) and is based on the discussion in #764 and #798. Tests included, documentation for Generic Gateway enhanced. It also fixes a minor logging issue on the SimpleEvent tests.

- [X] this PR contains unit tests
- [X] this PR has been tested for backwards compatibility

Currently the Generic Gateway starts both the 'simpleevent' and the 'cloudevent' endpoints/listeners. I'm not sure about is whether this is bad, though, considering from a security perspective that both endpoints are protected by the same authentication mechanism. Thoughts are more than welcome.


Gopkg.lock
```diff
@@ -62,6 +62,17 @@
   revision = "35431d848dcddf136bfeee424c6a1ab2ea539646"
   version = "v1.1.0"
 
+[[projects]]
+  digest = "1:64f6210405dce72b1218496fc46b1392eef4b7c1621f25d6ec05313cd13e6a0f"
+  name = "github.com/cloudevents/sdk-go"
```
radu-matei: Is this not defined in `Gopkg.toml`?

docs/topics/genericgateway.md
```diff
@@ -47,18 +64,60 @@ When calling the Generic Gateway `simpleevent` endpoint, you must include a cust
 }
 ```
 
-`Ref` and `commit` values are used to configure the specific revision that Brigade will pull from your repository, if you're using one. If both values are missing, then your Build's `ref` will be set to `master`. If you do not wish to provide any payload, just include an empty JSON object (like `{}`).
+`ref` and `commit` values are used to configure the specific revision that Brigade will pull from your repository, if you're using one. If both values are missing, your Build's `ref` will be set to `master`. Of course, you can also provide any other values you need.
```
radu-matei: As this gateway would be mostly used for non-Git scenarios, would it make sense to add this under a note?

docs/topics/genericgateway.md
```diff
@@ -47,18 +64,60 @@ When calling the Generic Gateway `simpleevent` endpoint, you must include a cust
...
+If you do not wish to provide any payload, just send an empty JSON object (`{}`).
```
radu-matei: Is this working if I don't include data at all?
Last time I checked it failed - do we want it to work?

docs/topics/genericgateway.md
```diff
@@ -47,18 +64,60 @@ When calling the Generic Gateway `simpleevent` endpoint, you must include a cust
...
+
+```json
+{
+    "type":   "com.example.file.created",
+    "source": "/providers/Example.COM/storage/account#fileServices/default/{new-file}",
+    "id":     "ea35b24ede421",
+    "specversion": "0.2",
+    "data": {
+        "ref": "refs/heads/changes",
+        "commit": "63c09efb6eb544f41a48901a6d0cc6ddfa4adb28",
```
radu-matei: Same thing about `ref` and `commit`.

docs/topics/genericgateway.md
```diff
@@ -47,18 +64,60 @@ When calling the Generic Gateway `simpleevent` endpoint, you must include a cust
...
+        "key1": "value1",
+        "key2": "value2"
+  }
+}
+```
+
+Bear in mind that 0.2 Cloud Events specification requires non-empty and proper values for "type","source","id" and "specversion".
```
radu-matei: Could you please link to the section of the CloudEvents spec that dictates this?

docs/topics/genericgateway.md
```diff
@@ -47,18 +64,60 @@ When calling the Generic Gateway `simpleevent` endpoint, you must include a cust
...
+
+A CloudEvent may include domain-specific information in the `data` field ([source](https://github.com/cloudevents/spec/blob/v0.2/spec.md#data-attribute)). Consequently, Brigade tries to read `ref` and `commit` values in the `data` field to configure the specific revision that Brigade will pull from your repository, if you're using one. If both values are missing, your Build's `ref` will be set to `master`.
```
radu-matei: This would be great under a `Note that if you plan to use this type of event with source control ...`.

docs/topics/genericgateway.md
```diff
@@ -47,18 +64,60 @@ When calling the Generic Gateway `simpleevent` endpoint, you must include a cust
...
 
-Last but not least, here is a sample Brigade.js file that could be used as a base for your own scripts that respond to Generic Gateway's `webhook` event. This script will echo the name of your project and 'webhook'.
+## Sample Brigade.js
+
+Here is a sample Brigade.js file that could be used as a base for your own scripts that respond to both Generic Gateway's event.
```
radu-matei: Aren't these actually two `brigade.js` files?

Do we want a single one that people can copy-paste?

docs/topics/genericgateway.md
```diff
@@ -47,18 +64,60 @@ When calling the Generic Gateway `simpleevent` endpoint, you must include a cust
...
+
+Here is a sample Brigade.js file that could be used as a base for your own scripts that respond to both Generic Gateway's event. 
 
 ```javascript
-  events.on("simpleevent", (e, p) => {
-  var echo = new Job("echo", "alpine:3.8");
+  events.on("simpleevent", (e, p) => { 
+  var echo = new Job("echosimpleevent", "alpine:3.8");
+  echo.storage.enabled = false;
```
radu-matei: Any reason why the storage is disabled?
(Only asking because since this is an example, it would be great if this was explained.)

docs/topics/genericgateway.md
```diff
@@ -47,18 +64,60 @@ When calling the Generic Gateway `simpleevent` endpoint, you must include a cust
...
+  echo.tasks = [
+    "echo Project " + p.name,
+    "echo SimpleEvent $EVENT_NAME"
+  ];
+
+  echo.env = {
+    "EVENT_NAME": e.type
```
radu-matei: FWIW, the environment variable `BRIGADE_EVENT_TYPE` is already set by the worker, so it's available on the job.

pkg/webhook/genericshared.go
```diff
@@ -0,0 +1,27 @@
+package webhook
```
radu-matei: Since it's the same package, does this require a separate file?

docs/topics/genericgateway.md
```diff
@@ -47,18 +64,60 @@ When calling the Generic Gateway `simpleevent` endpoint, you must include a cust
 }
 ```
 
-`Ref` and `commit` values are used to configure the specific revision that Brigade will pull from your repository, if you're using one. If both values are missing, then your Build's `ref` will be set to `master`. If you do not wish to provide any payload, just include an empty JSON object (like `{}`).
+`ref` and `commit` values are used to configure the specific revision that Brigade will pull from your repository, if you're using one. If both values are missing, your Build's `ref` will be set to `master`. Of course, you can also provide any other values you need.
+If you do not wish to provide any payload, just send an empty JSON object (`{}`). 
+
+### Calling the CloudEvent endpoint
```
radu-matei: Could you please include a simple `curl` POST command with the payload (both for this and for a simple event?)
It would be extremely helpful to try it out.

brigade-generic-gateway/cmd/brigade-generic-gateway/server.go
```diff
@@ -49,13 +49,15 @@ func newRouter(store storage.Store) *gin.Engine {
...
-		events.Use(gin.Logger())
-		events.POST("/:projectID/:secret", handler)
-	}
+	handlerSE := webhook.NewGenericWebhookSimpleEvent(store)
+	eventsSE := router.Group("/simpleevent")
+	eventsSE.Use(gin.Logger())
+	eventsSE.POST("/:projectID/:secret", handlerSE)
+
+	handlerCE := webhook.NewGenericWebhookCloudEvent(store)
+	eventsCE := router.Group("/cloudevent")
```
radu-matei: Do we want to pluralize this, as the spec is `CloudEvents`?
If we did pluralize this one, it would probably make sense to pluralize `simpleevents` as well.

docs/topics/genericgateway.md
dgkanatsios: definitely, fixing

docs/topics/genericgateway.md
dgkanatsios: I had that because it's not needed for the sample but yeah, this will create more questions than answers, removing.

docs/topics/genericgateway.md
dgkanatsios: ah, correct, gonna set this to another variable then :)

pkg/webhook/genericshared.go
dgkanatsios: this func is used by both SimpleEvent and CloudEvent, what would be the `idiomatic` :) way to have it?

brigade-generic-gateway/cmd/brigade-generic-gateway/server.go
dgkanatsios: I wasn't sure about that, to be honest. I used singular because we're just using POST and NOT GET (so, POST a - single - cloud event contrary to GET - many - cloud events). However, plural is much more common in APIs.
But, as I said, no strong opinions here... whatever you think, I'm good.

pkg/webhook/genericshared.go
radu-matei: Since it's the same package, and it's not such large functionality, I think you can leave it in one of the two files (ideally the first one that implemented it).

docs/topics/genericgateway.md
dgkanatsios: great idea, adding

Gopkg.lock
dgkanatsios: pinned commit, thanks!

docs/topics/genericgateway.md
dgkanatsios: this is on SimpleEvent, fixing it to allow empty payloads

docs/topics/genericgateway.md
dgkanatsios: Do you mean adding a note like [here](https://github.com/Azure/brigade/pull/818#discussion_r)?

docs/topics/genericgateway.md
radu-matei: Sorry, I should have been a bit more specific - not adding an event object results in a failure - passing `{}` works. Is this actually the behaviour we want?

dgkanatsios: yeah, I agree that it should allow for empty payloads, fixed it [here](https://github.com/dgkanatsios/brigade/blob/cloudevents/pkg/webhook/genericsimpleevent.go#L56) and the relevant test is [here](https://github.com/dgkanatsios/brigade/blob/cloudevents/pkg/webhook/genericsimpleevent_test.go#L195).

docs/topics/genericgateway.md
```diff
@@ -47,24 +64,105 @@ When calling the Generic Gateway `simpleevent` endpoint, you must include a cust
...
+  }
+}
+```
+
+You can use curl to test it:
+
+```bash
+curl --header "Content-Type: application/json" \
+  --request POST \
+  --data '{ "type":   "com.example.file.created",
```
vdice: If desired, we could reference the json above (if placed in file) and remove duplication in this curl cmd using `--data-binary @example.json` or `--data "$(cat example.json)"`

pkg/webhook/genericcloudevent.go
```diff
@@ -0,0 +1,114 @@
+package webhook
```
vdice: Would it be prudent to create a "Generic Webhook" interface that both the simple event and cloud event webhooks could then implement?  Arguments for this would be if a non-trivial amount of code (including the test code) would be consolidated in doing so and/or if additional generic webhook variants might be added in the future... 

Definitely defer to the author on this idea -- also, even if it might gain momentum, can surely be in the form of a follow-up refactor ticket.

docs/topics/genericgateway.md
dgkanatsios: Hm, in this case wouldn't we lose the benefit to simply copy/paste this command (for testing purposes)?

pkg/webhook/genericcloudevent.go
dgkanatsios: That's a great idea and definitely something I'm considering.
The only thing that pinches me if we go down this path is that I'm not sure whether in the future we would like to have any special treatment for a CloudEvent since it carries a [context](https://github.com/cloudevents/spec/blob/v0.2/spec.md#context-attributes) and we may want to do different things depending on its content. E.g. [source](https://github.com/cloudevents/spec/blob/v0.2/spec.md#source) might be an absolute path, a relative path, a urn, an email, etc. and we could want to have the CloudEvent handler take different actions depending on it.
On the contrary, a SimpleEvent has no pre-defined schema so I doubt we'll want to do anything more rather than just checking if there are ref/commit values present.
What do you think?

brigade-generic-gateway/cmd/brigade-generic-gateway/server_test.go
```diff
@@ -55,3 +55,49 @@ func TestNewRouter(t *testing.T) {
 		t.Fatalf("Expected 401 status, got: %s", res.Status)
 	}
 }
+
+func TestNewRouterCloudEvent(t *testing.T) {
```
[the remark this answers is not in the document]
dgkanatsios: Regarding a single set of test cases for both event types, I'm not sure whether this would be ideal, since CloudEvents has required properties etc. and we might want to do special checks on them (e.g. specversion which must be '0.2'). 
However, many thanks for the `iterates over an array of structs` proposal, just implemented that!
Only one caveat: if I do `t.Parallel()` in [this](https://github.com/dgkanatsios/brigade/blob/cloudevents/pkg/webhook/genericsimpleevent_test.go#L163) line, I had weird results :( Need to check it out, but I suppose it does not block this PR

dgkanatsios:
Amended the commit incorporating @vdice and @radu-matei feedback, thanks both!
One question: right now there is the `/cloudevent` (or `/cloudevents', whatever we decide) endpoint that supports cloudevents v0.2. How are we going to handle subsequent CloudEvents versions? As I see it, we have three options:
- specify in the documentation that `/cloudevent` endpoint is `alpha` and things will change till cloudevents spec reaches 1.0
- do `/cloudevent/v02`, '/cloudevent/v03` etc. endpoints
- go the K8s API way and rename the endpoint to `/alpha/cloudevent`

My concern with the first option is whether breaking changes on the Generic Gateway will be considered OK when Brigade 1.0 is released. On the other hand, second option might be an overkill. Thoughts?

radu-matei:
That is a really good point, here are my thoughts:

- we are releasing 1.0 for the Brigade core, gateways should (and  _will_) have their own release cycle.
- I think I would be ok with having `/cloudevents/v02` for the current version, and increment to adapt new versions of the specification as they are released.

Gopkg.toml
```diff
@@ -57,3 +57,7 @@
 [[constraint]]
   name = "gopkg.in/AlecAivazis/survey.v1"
   version = "1.6.1"
+
+[[constraint]]
+  name = "github.com/cloudevents/sdk-go"
+  revision = "feec6e002535db7561a364f82f67d2c9ebe4596a"
```
radu-matei: nit: it would be more visible if you used the 0.2.0 release tag - https://github.com/cloudevents/sdk-go/releases/tag/0.2.0 - but clearly not a big issue.

brigade-generic-gateway/cmd/brigade-generic-gateway/server.go
radu-matei: Yeah, the more I look at this, the more I like the pluralized versions.

Gopkg.toml
dgkanatsios: It's definitely more clear, fixing.

dgkanatsios:
thanks @radu-matei, switched endpoints to `simpleevents` and `cloudevents/v02`!

dgkanatsios:
I'd also like to ask a pair of eyes on [this](https://github.com/dgkanatsios/brigade/blob/cloudevents/pkg/webhook/genericsimpleevent_test.go#L180) method. It's necessary to do polling because currently Generic Gateway handlers are goroutines, so I'm giving it 3 seconds to check if a new Build has been created.
[Previous incarnation](https://github.com/dgkanatsios/brigade/blob/af4eec9bc50ac8ebcbd2b7fb3860595bb28b17f2/pkg/webhook/genericsimpleevent_test.go#L203) of this method would not do anything if the 3 second timeout would be reached and I *suspect* that we might have had a goroutine leak. So I did a modification (basically added the `stopChan` channel) which seems to work, but I'd definitely ask for another look.

pkg/webhook/genericcloudevent.go
radu-matei: I think for for we can keep it this way, and as it becomes more clear if the implementations for later versions of CloudEvents and our simple event converge or diverge, we can re-evaluate.

dgkanatsios:
> EDIT: as we're switching to cloudevents/vX for CloudEvents, would it make sense also for the /simpleevents endpoint to be versioned?
On one hand, it maintains consistency with the other endpoint -- on the other, as we don't have a schema, other than the behaviour itself, there isn't anything else to version. Thoughts?

I started writing that I don't think it is necessary but I thought of a scenario where (in the future) we may want to allow a Brigade user to **require** special fields to be present in the event payload (e.g. an `ID`). 
Good thinking, I'm updating `/simpleevent` endpoint to `/simpleevent/v01`.

pkg/webhook/genericcloudevent.go
vdice: Thanks @dgkanatsios, I'd figured you had considered both approaches.  The approach in this PR definitely seems the sensible path given the points you've mentioned.  In agreement with @radu-matei above -- makes sense to revisit at a later date.

docs/topics/genericgateway.md
vdice: You're right, we would lose that ability.  Let's keep it 👍

PR #838, opened 2019-03-08
If host.OS is Windows, use PowerShell as command instead of Bash

### LINKED ISSUE #823: If host.os is set to Windows, override bash script with a Windows alternative.
Right now, regardless of the operating system of the job specification, the entrypoint of the container is a bash script -- meaning it cannot run Windows containers.

radu-matei:
closes #823

This PR adds very early experimental Windows Server containers support.
It has been tested with a Virtual Kubelet node backed by Azure Container Instances, so the following are current known limitations:

- job and build caches are not supported (limitation of volumes in ACI)
- the Git sidecar is not supported (limitation of volumes in ACI)
- the entrypoint is passed directly to the command arguments, not constructed as a script (limitation of volumes in ACI)

To test this, you can add a Windows Virtual Node with Virtual Kubelet and test with a Brigade project that has the sidecar disabled, using the following pipeline:

```
const { events, Job } = require('@brigadecore/brigadier')

events.on("exec", (brigadeEvent, project) => {
    var hello = new Job("windows-job")
    hello.image = "mcr.microsoft.com/windows/nanoserver:1809"
    hello.tasks = ["Get-Process"]

    hello.host.name = "virtual-kubelet"
    hello.host.os = "windows"
    hello.host.shell = "powershell.exe"

    hello.storage.enabled = false;

    hello.resourceRequests.cpu = "1"
    hello.resourceRequests.memory = "1G"

    hello.run()
})

```

**If applicable**:

### DIFF
--- a/brigade-worker/src/k8s.ts
+++ b/brigade-worker/src/k8s.ts
@@ -346,35 +346,52 @@ export class JobRunner implements jobs.JobRunner {
 
     this.runner.spec.containers[0].env = envVars;
 
-    let mountPath = job.mountPath || this.options.mountPath;
 
-    // Add secret volume
-    this.runner.spec.volumes = [
-      { name: secName, secret: { secretName: secName } } as kubernetes.V1Volume
-    ];
-    this.runner.spec.containers[0].volumeMounts = [
-      { name: secName, mountPath: "/hook" } as kubernetes.V1VolumeMount
-    ];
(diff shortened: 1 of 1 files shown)


radu-matei:
This technically only sets the correct shell when the host OS is Windows - we should probably get this merged.

lippertmarkus:
Any updates on that?

PR #849, opened 2019-03-21
ref(brigade.js): remove acr publishing

vdice:
As part of the migration of Brigade to this brigadcore CNCF org, I propose we remove publishing of Docker images to ACR (which is managed on an internal Microsoft subscription). 

Images would therefore solely be published to the [brigadecore DockerHub org](https://hub.docker.com/u/brigadecore), which aligns with all of the registry/image defaults in documentation and [charts](https://github.com/brigadecore/charts).

### DIFF
--- a/brigade.js
+++ b/brigade.js
@@ -189,7 +189,7 @@ function release(p, tag) {
 }
 
 // Separate docker build stage as there may be multiple consumers/publishers,
-// For example, publishing to Dockerhub and ACR below
+// For example, publishing to Dockerhub below
 function goDockerBuild(e, p) {
   // We build in a separate pod b/c AKS's Docker is too old to do multi-stage builds.
   const builder = new Job(`${projectName}-docker-build`, goImg);
@@ -236,34 +236,6 @@ function dockerhubPublish(project, tag) {
   ];
 
   return publisher;
-}
-
-function acrBuild(project, tag) {
-  const acrImagePrefix = "public/deis/";
-
-  let registry = project.secrets.acrRegistry || projectName;
-  var acrBuilder = new Job("az-build", "microsoft/azure-cli:latest");
-
-  acrBuilder.imageForcePull = true;
-  acrBuilder.storage.enabled = true;
-
-  acrBuilder.tasks = [
-    addMake,
-    // Create a service principal and assign it proper perms on the ACR.
-    `az login --service-principal -u ${project.secrets.acrName} -p '${project.secrets.acrToken}' --tenant ${project.secrets.acrTenant}`,
-    `cd /src`,
-    `mkdir -p ./bin`,
-    "for i in $(make echo-images); do \
-      cd ${i} ; \
-      echo '========> Building ${i}' ; \
-      cp -av " + sharedMountPrefix + "${i}/rootfs ./ ; \
-      az acr build -r " + registry + " -t " + acrImagePrefix + "${i}:" + tag + " . ; \
-      echo '<======== Finished ${i}' ; \
-      cd .. ; \
-     done"
-  ];
-
-  return acrBuilder;
 }
 
 function slackNotify(title, msg, project) {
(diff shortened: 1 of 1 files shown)


squillace:
Hi @vdice @technosophos @radu-matei and crowd. As we fully transition to CNCF-land, I'm wondering whether there's a public repo we can use that actually delivers some telemetry that gives numbers we can count on. I no longer have any confidence in docker hub's numbers at all. I'm not suggesting MCR, which is for Microsoft images, but have we thought about quay.io? I mean this comment so that y'all can edumacate me.

bacongobbler:
Back in early Helm 3 discussions, we talked about moving Tiller over from DockerHub to quay for tracking download metrics. As far as I can remember, @mattfarina found that the metrics there were useful and was a potential landing zone for Helm.

However, because we removed Tiller from Helm's architecture in Helm 3, we decided it wasn't worth the time sink to migrate all our users over to quay for a short period of time, and decided we can track metrics based on how many times the Helm CLI is downloaded instead. We still need to re-visit that discussion as the Helm CLI is currently hosted on a GCS bucket owned by the Kubernetes org, but hopefully that provides some background context for y'all from another CNCF project. :)

squillace:
I'd like to consider it. although I understand the helm considerations.... this is brigade, right? I'm assuming you're just giving me the history of possible moves in the related community here? 

Here's what I'd like to do: I'd like to start being able to offer the entire community the telemetry from community images on projects that we start and/or contribute to. Anything that doesn't offer easy telemetry is in my view a poor idea. Maybe we don't start with this move, but I'd like to pick that up generally and drive toward a good registry that does telemetry well -- if possible.

SteveLasker:
ACR supports public repos (anonymous access). We haven’t lit this up in the portal as we want to finish the per repo permissions first so you can set a specific repo as public while others are private. ACR has telemetry as well.

vdice:
If I'm understanding correctly, I'm seeing 2 tickets.  The first, this one, proposes to remove publishing to ACR and the second one, to be created, is a potential feature request to migrate and/or add publishing to a repo that supports desired telemetry.

As a background, Brigade images have _always_ been pushed to (and pulled from) Dockerhub as the default, albeit via the `deis` org.  We've only recently changed to the `brigadecore` Dockerhub org.  In addition, docs and charts have _always_ referenced these Dockerhub-hosted images.  Therefore, if I'm understanding the complaints against this registry, we've _always_ been lacking in telemetry.

Within the last handful of months, we did add publishing to ACR, with vague intentions to update docs/charts to pull from this registry by default.  However, this no longer makes as much sense now that Brigade has moved from Azure into the brigadecore CNCF org.

technosophos:
So we definitely need to stop publishing to the current ACR, which is (a) private only, and (b) publishing to my subscription. For us to meet our CNCF obligations, we should deactivate this.

For the future, we probably should use a registry that has better telemetry. But, yes, @vdice -- you are right that that is a second ticket.

SteveLasker:
ACR has anonymous/public access with telemetry. We’re happy to support any of your community needs.

technosophos:
@SteveLasker Okay. I think we'd have to do what we did for Helm and add a new subscription owned by CNCF, and then activate ACR on that. I'll see what we need to do to get that process rolling /cc @squillace

vdice:
Thanks all.  I've created an issue in https://github.com/brigadecore/brigade/issues/856 to track telemetry needs and additional registry publishing candidates.

squillace:
Catching up here. @vdice @technosophos IFF ACR does do anonynous ("public") access, is there any need to move to dockerhub? Just confirming my understanding of https://github.com/brigadecore/brigade/pull/849#issuecomment-477660922

mattfarina:
@squillace Since my name was used I've been summoned.

Who pays for ACR in this model? With DockerHub and Quay there is a free tier so no one needs to. If this is a CNCF project, how does the money and control flow? What happens maintainers come on from other companies (a requirement for CNCF incubation)?

We used Quay for Helm because it provides better short term metrics.

PR #864, opened 2019-03-28
[docs] add analytics

flynnduism:
adds analytics for docs.brigade.sh


dgkanatsios:
Cookies are anonymous so I'm not sure whether GDPR does apply or not. What does apply is the [EU cookie law](http://ec.europa.eu/ipg/basics/legal/cookies/index_en.htm), however, Google Analytics *seems to* [comply](https://support.google.com/analytics/answer/7105316?hl=en).

flynnduism:
I've updated this PR to avoid any potential GDPR pitfalls:

* switched out the GA code to ensure it [anonymizes user IPs](https://support.google.com/analytics/answer/2763052?hl=en) 
* added a cookie notice banner 

---

Screen Shot 2019-05-09 at 1 48 50 PM

technosophos:
@flynnduism Forgive my total ignorance on GA, but where is the notice coming from?

flynnduism:
@technosophos I created that notice using general disclaimer wording, edited to disclose 'analytical purposes' as that's the only script the site is using here.

Really that notice is a catch all backup measure - as with this configuration of GA, there are no cookies being set. You can [test](https://deploy-preview-864--brigade-docs.netlify.com/) and inspect the site to see that no cookies are set.

Screen Shot 2019-05-09 at 2 27 53 PM

_(found under Chrome dev tools > Application > Cookies > Domain name)_

PR #869, opened 2019-04-02
Eliminates a yarn race condition when installing brigade.json deps

### LINKED ISSUE #867: Race condition when installing brigade.json dependencies
Output of `brig version`: `v1.0.0`

Output of `kubectl version`:
```
Client Version: v1.8.15
Server Version: v1.8.15-1+9f66f5c998b397
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): `AWS, custom`

---

When installing dependencies from `brigade.json`, the `addYarn` function returns a promise, but `addDeps` calls `addYarn` in a loop and does not wait on these promises to resolve.  This leads to multiple racing `yarn` processes that clobber each others' efforts.

To reproduce this, create a brigade project with a `brigade.json` that has multiple dependencies.

```json
{
  "dependencies": {
    "is-thirteen": "2.0.0",
    "moment": "2.24.0",
    "ramda": "0.26.1"
  }
}
```

Then, in `brigade.js`, try to `require` each:

```javascript
require('is-thirteen');
require('moment');
require('ramda');
```

Since it's a race condition, it's not always possible to observe, but in my tests it happens frequently.  It's also possible to connect to the running worker node and see that some modules are not installed into `node_modules`.

---

https://github.com/brigadecore/brigade/blob/ae56a2ab4bc93e663ab16a0e2034256ec3916647/brigade-worker/prestart.js#L53-L72

The simplest fix for this would be to use `child_process.execSync` rather than using `child-process-promise`, here:

https://github.com/brigadecore/brigade/blob/ae56a2ab4bc93e663ab16a0e2034256ec3916647/brigade-worker/prestart.js#L67

chris-chambers:
`yarn` was previously invoked using via `child-process-promise` without
waiting for the promises to resolve, which led to a race between
multiple `yarn` instances attempting to update package.json, yarn.lock
and node_modules simultaneously.  This led to failed package
installation for brigade.json files with more than one dependency.

Now, `yarn` is invoked synchronously and in a single invocation, rather
than via multiple asynchronous execs.  This prevents the race condition
and makes package installation faster.

Signed-off-by: Chris Chambers <chris.chambers@peanutcode.com>

Closes #867 

I need some help assessing the "If applicable" section below. I didn't see any existing unit tests for this functionality, and it may be tricky to mock.  For backward compatibility, are there functional tests I can run?

**If applicable**:

### DIFF
--- a/brigade-worker/package.json
+++ b/brigade-worker/package.json
@@ -21,12 +21,16 @@
   "devDependencies": {
     "@types/chai": "^4.0.1",
     "@types/mocha": "^5.2.6",
+    "@types/mock-require": "^2.0.0",
     "@types/node": "^11.9.6",
+    "@types/sinon": "^7.0.11",
     "@types/ws": "6.0.1",
     "chai": "^4.1.0",
     "mocha": "^6.0.2",
+    "mock-require": "^3.0.3",
     "prettier": "^1.9.1",
     "rimraf": "^2.6.2",
+    "sinon": "^7.3.1",
     "ts-node": "^8.0.2",
     "typedoc": "^0.14.2",
     "typescript": "^3.2.2"
--- a/brigade-worker/prestart.js
+++ b/brigade-worker/prestart.js
@@ -1,6 +1,6 @@
 const process = require("process")
 const fs = require("fs")
-const exec = require("child-process-promise")
+const { execFileSync } = require("child_process")
 
 // Script locations in order of precedence.
 const scripts = [
@@ -21,20 +21,22 @@ const scripts = [
 ];
 
 //checked out in repo
-const deps = "/vcs/brigade.json"
+const depsFile = "/vcs/brigade.json"
 
-addDeps()
+if (require.main === module)  {
+  addDeps()
 
-try {
-  var data = loadScript()
-  let wrapper = "const {overridingRequire} = require('./require');((require) => {" +
-    data.toString() +
-    "})(overridingRequire)"
-  fs.writeFileSync("dist/brigade.js", wrapper)
-} catch (e) {
-  console.log("prestart: no script override")
-  console.error(e)
-  process.exit(1)
+  try {
+    var data = loadScript()
+    let wrapper = "const {overridingRequire} = require('./require');((require) => {" +
+      data.toString() +
+      "})(overridingRequire)"
+    fs.writeFileSync("dist/brigade.js", wrapper)
+  } catch (e) {
+    console.log("prestart: no script override")
+    console.error(e)
+    process.exit(1)
+  }
 }
 
 // loadScript loads the first configured script it finds.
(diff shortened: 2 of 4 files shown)


chris-chambers:
I would very much like to add a test, but there are a couple of problems to talk through:

**addDeps is called at the top level**

The top-level call to `addDeps` means that when `prestart.js` is `require`d, it will immediately try to read `/vcs/brigade.json` and search for dependencies.  To test the file without a significant refactor or a strange setup step for the test, we'll need to mock the `fs` module.

**`yarn` is invoked and touches the file system**

Because `yarn` is invoked via `execFileSync`—and that may be tricky to control in a test—we'd need to mock `child_process.execFileSync`.  But, then someone might refactor to use a different kind of `exec`, and any checking against/spying on `execFileSync` would be meaningless.

**None of these ideas covers the actual regression**

What we'd really like to do is verify that if a dependency is specified in `brigade.json`, then after the execution of `prestart.js`, that dependency has been installed into `package.json` and `node_modules`.  Any other kind of test we do is just a proxy for guaranteeing this behavior.

So, maybe the test should focus in this direction.  We could:

- Create a temporary directory (say, `/tmp/foo`).
- Add a well-known `brigade.json`.
- Execute `prestart.js` in `/tmp/foo` (this is tricky because of the hard-coded `"/vcs/brigade.json"`.
- Check `/tmp/foo/package.json`
- Check `/tmp/foo/node_modules`

---

None of these ideas are very satisfying to me.  The mocking approach leads to brittle tests that will almost always need to be updated during a refactor of the dependency installation code.  And the temporary directory approach feels more like testing `yarn` than testing Brigade.

Any preference?  Or perhaps a nicer idea?

chris-chambers:
Oh, another question:

How is `prestart.js` invoked?  I never found where it actually happens.  If we can guarantee that it is the main module (ie, `require.main === module` inside `prestart.js`), then a small refactoring could make it possible to do most of the testing unit-test style rather than system-test style.

radu-matei:
`prestart.js` is called by Node directly, before starting the worker -- see https://github.com/brigadecore/brigade/blob/master/brigade-worker/package.json#L10

chris-chambers:
How far are we from being able to run a functional test as part of CI?  Is there anything existing to hook into?

radu-matei:
I think @vdice has been playing with `kind`, but not sure how far he is running a functional test on top of Brigade.

radu-matei:
Opened #879 to track end to end testing for Brigade.

chris-chambers:
@radu-matei, since it seems like functional tests are a pretty big lift, maybe we should go ahead and land this PR so it doesn't get stale.  What do you think?

chris-chambers:
@radu-matei, is anything else needed to get this PR merged?  Roll back the tests?  Simplify them?

chris-chambers:
@vdice, sorry about that.  I've pushed a commit that fixes the test runs.  Not sure how I missed that.  If you're happy with it after this change, I'd like to squash before merging, if that's alright.

vdice:
@chris-chambers not a problem!  Indeed, all looks good on this branch now.  Proceed w/ squashing commits when convenient!

cc @radu-matei to make sure feedback has been addressed.

chris-chambers:
Squashed and rebased to the latest master!

PR #870, opened 2019-04-02
Use `module-alias` to improve module resolution in Brigade scripts

### LINKED ISSUE #868: Module resolution in local dependencies is counter-intuitive
Inside [local dependencies](https://docs.brigade.sh/topics/dependencies/#using-local-dependencies-from-the-project-repository), `require` behaves counter-intuitively.

There are two problems:

### Brigadier Aliases
In `prestart.js`, `require` is replaced by a custom function that allows several aliases of `brigadier` to be resolved.  But, this aliasing is not available in any required local dependencies.

### Module Resolution in Local Deps
In all local dependencies, `require` sees `modules.path` as:
```javascript
[ '/vcs/node_modules', '/node_modules' ]
```

But, `yarn` installs dependencies into `/home/src/node_modules`.  The module search path from within `brigade.js` is correct, since it is copied-and-modified into `/home/src/dist`.  But local dependencies are left in-place in `/vcs/`, so they aren't able to resolve modules in `/home/src/node_modules`.

---

I'm working around both these issues using the `module-alias` module, and configuring it like:

```javascript
const modAlias = require('module-alias');

// Add the search paths for the current module (brigade.js) so they are
// available to all modules, everywhere.
module.paths.forEach(modAlias.addPath);

// Globally perform the same brigade aliasing that prestart.js does.
modAlias.addAlias('brigade', __dirname + '/brigadier.js');
modAlias.addAlias('brigadier', __dirname + '/brigadier.js');
modAlias.addAlias('@brigadecore/brigadier', __dirname + '/brigadier.js');

// Enable module-alias
modAlias();
```

The semantics of this may not be perfect, but I think something like this is pretty reasonable, given the desire to leave local dependencies in `/vcs`.  In fact, I think with a change like this, `brigade.js` could also be left in place, rather than being copied and modified.

chris-chambers:
Previously, local dependency resolution and Brigadier aliasing were done
by a special `require` replacement, which was only available in the main
Brigade script (eg, `brigade.js`).  This meant that local dependencies
could not themselves depend on packages installed via `brigade.json`.

Now, `module-alias` takes care of Brigadier aliasing and applies the
correct module resolution paths to allow local dependencies to require
packages from the `node_modules` that `brigade.json` targets.

Signed-off-by: Chris Chambers <chris.chambers@peanutcode.com>

Closes #868 

Consider this a proposal to address #868.  In particular, it leaves the Brigade script (eg, `brigade.js`) wherever it was discovered, rather than wrapping it and writing to `dist/brigade.js`.  Someone with more knowledge of the `require` wrapper should assess whether the behavior changes here are all positive.  I can't think of a downside to executing the discovered script via (the real) `require`, rather than copying its contents into the dummy `brigade.js` and importing it, but a second opinion would be very welcome.

**If applicable**:
_I've tested some minimal examples, but automated end-to-end tests would be better (#879)._


brigade-worker/src/index.ts
```diff
@@ -28,15 +28,57 @@
...
+  "brigadier": __dirname + "/brigadier",
+  "@brigadecore/brigadier": __dirname + "/brigadier",
+});
+// Add the current module resolution paths to module-alias, so the node_modules
+// that prestart.js adds to will be resolvable from the Brigade script and any
+// local dependencies.
+module.paths.forEach(moduleAlias.addPath);
+moduleAlias();
+
+// Script locations in order of precedence.
```
chris-chambers: Maybe this whole block (through the `findScript` function) would be better as a separate module? It feels a little distracting to me inline like this.

chris-chambers:
Anyone have thoughts on this PR/proposal?  It's working well for me, but I'd like to get more perspective on the other ways that brigade.js files can be invoked.

chris-chambers:
@vdice, my pleasure!   This PR will almost certainly need to be rebased if #869 lands first.

Should I update the docs to match the new behavior here?

vdice:
> Should I update the docs to match the new behavior here?

Good call!  Yes, that'd be great.

chris-chambers:
@vdice, I've rebased and squashed, and updated the dependencies documentation to cover the new behavior.  If you're happy with it, merge when ready.

PR #881, opened 2019-04-07
brig check initial implementation

### LINKED ISSUE #815: brig check command
It would be really helpful to have a `brig check` command, similar to the [Linkerd Check](https://linkerd.io/2/reference/cli/check/) command, that makes sure you have access to the cluster, and that all components required for Brigade are running properly.

dgkanatsios:
Fixes #815 as per https://github.com/brigadecore/brigade/issues/815#issuecomment-466362299

It's a trivial PR that implements of `brig check` which offers a quick check on the health of the system by checking for
-  at least one active API Server Pods
-  at least one active Controller Pods
-  at least one active Kashti Pods
-  if Vacuum CronJob is not suspended and has a non-empty schedule

Do you think we can add anything else?

### DIFF
--- a/brig/cmd/brig/commands/check.go
+++ b/brig/cmd/brig/commands/check.go
@@ -1,0 +1,118 @@
+package commands
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	apps_v1 "k8s.io/api/apps/v1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func init() {
+	Root.AddCommand(check)
+}
+
+const (
+	checkUsage = `Checks the status of your Brigade installation
+
+Specifically, it reports Desired/Current/Running/Up-to-date/Available/Unavailable Pods for the Controller, API Server and Kashti deployments.
+`
+)
+
+var check = &cobra.Command{
+	Use:   "check",
+	Short: "Checks the status of your Brigade installation",
+	Long:  checkUsage,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return checkBrigadeSystem()
+	},
+}
+
+func checkBrigadeSystem() error {
+	/*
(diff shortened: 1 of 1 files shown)


brig/cmd/brig/commands/check.go
```diff
@@ -0,0 +1,121 @@
...
+
+	apiDeploymentFound := false
+	controllerDeploymentFound := false
+	kashtiDeploymentFound := false
+
+	for _, deployment := range deployList.Items {
+		if lbl := deployment.Labels["app.kubernetes.io/name"]; lbl != "" {
+			if strings.HasSuffix(lbl, "-brigade-api") && deployment.Labels["role"] == "api" {
+				apiDeploymentFound = true
+				reportDeployStatus(deployment, "Brigade API Server")
```
radu-matei: One question -- do we want to error out if Kashti or the API are unavailable?
As they are not essential, this might be informational -- thoughts?

dgkanatsios: yeah, you're right. I also altered the Deploy message to make it more consistent with `kubectl get deploy`, let me know what you think.

brig/cmd/brig/commands/check.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+func init() {
+	Root.AddCommand(check)
+}
+
+const (
+	checkUsage = `Checks the status of your Brigade installation
+
+Specifically, it checks if the Controller, API Server and Kashti deployments have more than one available replicas (pods).
```
vdice: Maybe `at least one` (or `exactly one`?) instead of `more than one`?

dgkanatsios: oh, nice catch, editing!

dgkanatsios:
rebased, thanks!

radu-matei:
I really like the output of `linkerd check` - https://linkerd.io/2/reference/cli/check/

Can we look at adding some checks in the future? :smile:

dgkanatsios:
sure, should we open an issue to discuss what to add? Apart from version check, what else should we check?

PR #892, opened 2019-04-12
brig project create templates

### LINKED ISSUE #816: brig project create templates
It would be really helpful if `brig project create` would have a template for a Git vs non-Git projects at creation time. When selecting the non-Git option, the following would happen:

- repository information would not be configurable
- clone URL information would not be configurable
- the custom VCS sidecar would be automatically set to `NONE`
- Git submodules inforrmation would not be configurable
- a Brigade.js file would be mandatory (either config map, or local)
- a secret for the generic gateway would be required

dgkanatsios:
Fixes #816, #917 

This PR modifies `brig project create` so that
- User is asked whether they do want to create a VCS or a no-VCS project
- If a no-VCS project is selected, some defaults are set (based on [comment](https://github.com/brigadecore/brigade/issues/816#issue-413206378))
  - repository information would not be configurable
  - clone URL information would not be configurable
  - the custom VCS sidecar would be automatically set to NONE
  - Git submodules inforrmation would not be configurable
  - a Brigade.js file would be mandatory (either config map, or local)
  - a secret for the generic gateway would be required

This needs TONS of testing, so pinging @technosophos @radu-matei @vdice @adamreese  :) Happy to say that VS Code for Kubernetes extension is pretty awesome for that, since you can run the modified code and immediately check what has been created.

The approach I've taken is to split the VCS and no-VCS code in two files (project_create_vcs.go and project_create_no_vcs.go) since I presume we'll add more stuff in the future so it will be cool to have keep two `templates` separate.

~~This PR is created as WIP till~~
- ~~we do agree on code/file format and messages~~
- ~~once previous point is agreed, a separate commit will be added that changes all docs that reference `brig project create`~~

~~Just for reference, files I've seen that document `brig project create` are~~
- ~~https://docs.brigade.sh/intro/quickstart/~~
- ~~https://github.com/brigadecore/brigade/blob/master/brig/README.md~~
- ~~https://docs.brigade.sh/intro/install/~~
- ~~https://docs.brigade.sh/intro/tutorial03/~~
- ~~https://docs.brigade.sh/topics/developers/~~
- ~~https://docs.brigade.sh/topics/genericgateway/~~
- ~~https://docs.brigade.sh/topics/projects/~~
- ~~https://docs.brigade.sh/topics/scripting/~~
- ~~https://docs.brigade.sh/topics/secrets/~~
- ~~https://docs.brigade.sh/topics/workers/~~

~~All these files will probably need to be amended with new `brig project create` messages and options.~~

Added a fix for #917, even though I suspect that this might be a breaking change for `brigade.js` files that for (whatever reason) got the repo URL from project.name and the actual project name from project.repo.name.

### DIFF
--- a/brig/cmd/brig/commands/project_create_test.go
+++ b/brig/cmd/brig/commands/project_create_test.go
@@ -7,15 +7,32 @@ import (
 
 const testProjectSecret = "./testdata/project_secret.json"
 
-func TestInitProject(t *testing.T) {
-	p := newProject()
-	if p.Name != defaultProject.Name {
+func TestInitProjectVCS(t *testing.T) {
+	p := newProjectVCS()
+	if p.Name != defaultProjectVCS.Name {
 		t.Fatal("newProject is not cloning default project")
 	}
 
 	p.Name = "overrideName"
-	if p.Name == defaultProject.Name {
+	if p.Name == defaultProjectVCS.Name {
 		t.Fatal("newProject returned the pointer to the default project.")
+	}
+}
+
+func TestInitProjectNoVCS(t *testing.T) {
+	p := newProjectVCS()     // VCS is the default
+	setDefaultValuesNoVCS(p) // switch the defaults to a no VCS project
+
+	if p.Kubernetes.VCSSidecar != "NONE" {
+		t.Fatal("VCSSidecar should be NONE")
+	}
+
+	if p.Repo.CloneURL != "" {
+		t.Fatal("CloneURL should be an empty string")
+	}
+
+	if p.Repo.Name != "" {
+		t.Fatal("Repo.Name should be an empty string")
 	}
 }
 
@@ -36,7 +53,7 @@ func TestParseSecret(t *testing.T) {
 }
 
 func TestLoadProjectConfig(t *testing.T) {
-	proj, err := loadProjectConfig(testProjectSecret, newProject())
+	proj, err := loadProjectConfig(testProjectSecret, newProjectVCS())
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/brigade-worker/test/k8s.ts
+++ b/brigade-worker/test/k8s.ts
@@ -21,16 +21,16 @@ describe("k8s", function() {
     });
   });
 
-  describe("secretToProject", function() {
-    it("converts secret to project", function() {
-      let s = mockSecret();
+  describe("secretToProjectVCS", function() {
+    it("converts secret to project - with a VCS", function() {
+      let s = mockSecretVCS();
       let p = k8s.secretToProject("default", s);
       assert.equal(
         p.id,
         "brigade-7e3d1157331f6726338395e320cffa41d2bc9e20157fd7a4df355d"
       );
-      assert.equal(p.name, "github.com/brigadecore/test-private-testbed");
-      assert.equal(p.repo.name, "brigadecore/test-private-testbed");
+      assert.equal(p.name, "brigadecore/test-private-testbed");
+      assert.equal(p.repo.name, "github.com/brigadecore/test-private-testbed");
       assert.equal(
         p.repo.cloneURL,
         "https://github.com/brigadecore/empty-testbed.git"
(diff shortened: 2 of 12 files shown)


brig/cmd/brig/commands/project_create_test.go
```diff
@@ -8,13 +8,13 @@ import (
 const testProjectSecret = "./testdata/project_secret.json"
 
 func TestInitProject(t *testing.T) {
-	p := newProject()
-	if p.Name != defaultProject.Name {
+	p := newProjectVCS()
```
vdice: Perhaps we were intending to do so after this PR exited Draft status, but can we add unit tests added around a NoVCS project?  (Checking corresponding defaults, etc.)

radu-matei: Yay for more tests.

radu-matei:
Manually tested, everything seems to be in order.
Once you address Vaughn's comments, I think we can go ahead and have an initial version of this we can iterate on.

Thanks!

radu-matei:
@dgkanatsios - is this ready for a re-review?
Thanks!

dgkanatsios:
not yet, did a rebase, will finish it up in the following days, thanks!

radu-matei:
You realise how awesome this PR is when creating some workshop material and still have to go through explaining how to set `VCS Sidecar` to `NONE`.

Thanks a lot for the work here, @dgkanatsios!

dgkanatsios:
Added a simple test plus amended the documentation.

radu-matei:
I think `REPO` should be blank for `no-VCS` projects when you do a `brig project list`:

Screen Shot 2019-05-22 at 01 59 44

dgkanatsios:
Nice catch @radu-matei. So the error is in [this line](https://github.com/brigadecore/brigade/blob/master/pkg/storage/kube/project.go#L197). What I'm gonna do is remove the `def` method call, so `brig project list` will return an empty string if a repo is not specified. Any arguments here?

dgkanatsios:
Do NOT merge till #917 is resolved

brigade-worker/test/k8s.ts
```diff
@@ -66,6 +66,26 @@ describe("k8s", function() {
...
+    it("converts secret to project - without a VCS", function() {
+      let s = mockSecretnoVCS();
+      let p = k8s.secretToProject("default", s);
+      assert.equal(
+        p.id,
+        "brigade-0e0ae80bb1243d95ea707257baebda6ccf96094cd0353d875ae903"
+      );
+      assert.equal(p.name, "noVCSProject");
+      assert.equal(p.repo, undefined);
+      assert.equal(s.data["genericGatewaySecret"], "SThkQ1g=") // Project class does not contain genericGatewaySecret field - this is because we do not want to expose the genericGatewaySecret to any job
```
dgkanatsios: At first I tried to asset `p.genericGatewaySecret` but I discovered that this property is not exposed in Project class in Brigadier ([link](https://github.com/brigadecore/brigadier/blob/master/src/events.ts#L199))
I debated whether to PR Brigadier and add it, but ultimately I decided against it. Reasoning is that I can't find any legit reason to expose `genericGatewaySecret` to a job running within a brigade.js file. WDYT?

radu-matei:
I tried again to create a job cache, and I still get the same failure:

```
body: 
      { kind: 'Status',
        apiVersion: 'v1',
        metadata: {},
        status: 'Failure',
        message: 'PersistentVolumeClaim "-cacher" is invalid: metadata.name: Invalid value: "-cacher": a DNS-1123 subdomain must consist of lower case alphanumeric characters, \'-\' or \'.\', and must start and end with an alphanumeric character (e.g. \'example.com\', regex used for validation is \'[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*\')',
        reason: 'Invalid',
        details: [Object],
        code: 422 } }
```

And I think I built the right `brig`:

```
$ brig version
8b5931d
```

dgkanatsios:
Weird, I tested that using kind and it worked :( I suppose you used the brigade.js from #917, right? Can you share the values you used on brig project create?

PR #851, opened 2019-03-25
Added Generic Gateway quickstart

dgkanatsios:
Fixes #806

This is a quickstart for the Generic Gateway.

At the time of issue, this PR should NOT be merged since `helm install -n brigade brigade/brigade --set genericGateway.enabled=true` does not install the latest version of Brigade/Generic Gateway. It should be merged after the relevant charts get updated. 

- [X] this PR contains documentation

PS. I also think that both quickstarts should be moved to docs.brigade.sh at a later time.

PR #893, opened 2019-04-12
fix config.toml to include quickstart link

dgkanatsios:
#865 was merged during #851 review. #851 did not include the necessary changes to `config.toml` so that `Quickstart` documentation would appear in docs index. This PR fixes that.


flynnduism:
LGTM thanks @dgkanatsios!

PR #908, opened 2019-05-09
allow global configuration of default storage classes

### LINKED ISSUE #905: Enhancement: Add build and cache storage class flags to controller
Currently, build and cache storage classes cannot be configured globally.

This has the unfortunate side-effect that if the default storage class on your cluster isn't adequate (e.g. doesn't support `AccessMode: ReadWriteMany`), the users of `brig project create` are forced down an advanced configuration path that they otherwise may not want to mess with.

By exposing these flags, we can, in turn, expose these options through Brigade's Helm chart, thus putting default build and cache storage class selection in the hands of whoever administers the Brigade installation instead of individual project owners.

krancour:
Fixes #905

### DIFF
--- a/brigade-worker/test/k8s.ts
+++ b/brigade-worker/test/k8s.ts
@@ -556,6 +556,119 @@ describe("k8s", function() {
         });
       });
     });
+    describe("cachePVC", () => {
+      let jr: k8s.JobRunner;
+      beforeEach(function() {
+        let j = new mock.MockJob("pequod", "whaler", ["echo hello"]);
+        let p = mock.mockProject();
+        let e = mock.mockEvent();
+        jr = new k8s.JobRunner().init(j, e, p);
+      });
+      context("when global default cache storage class is specified", () => {
+        beforeEach(function() {
+          jr.options.defaultCacheStorageClass = "foo";
+        });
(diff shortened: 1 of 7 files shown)


krancour:
@vdice I added the unit tests you suggested. JS/TS is not my strong suit. lmk if these look ok.

brigade-worker/test/k8s.ts
```diff
@@ -556,6 +556,119 @@ describe("k8s", function() {
...
+            assert.equal(pvc.spec.storageClassName, jr.project.kubernetes.cacheStorageClass)
+          });
+        });
+        context("when the cache storage class is not overridden at the project level", () => {
+          beforeEach(function() {
+            jr.project.kubernetes.cacheStorageClass = "";
+          });
+          it("it falls back on the cluster default", () => {
+            let pvc = jr['cachePVC']()
+            // Undefined means k8s will use the cluser's default storage class
```
vdice: typo: `cluser's` (same on line... 666 😅 )

krancour: Good eye. Fixing.

krancour: Fixed

PR #914, opened 2019-05-17
Log unhandled rejection with error message

### LINKED ISSUE #913: Brigade should log thrown error message
Consider the following example brigade.js file:

```
const { events } = require("brigadier")

events.on("exec", function() {
  throw new Error("This should be logged when failing the event handler.")
})
```

When this file is used to run `brig run <test project>`, although the pipeline fails as intended, the error message isn't logged:

```
 $ brig run test -f test-project/throw-error.js
Event created. Waiting for worker pod named "brigade-worker-01dast93c17zp0h3ftvqh6n9q1".
Build: 01dast93c17zp0h3ftvqh6n9q1, Worker: brigade-worker-01dast93c17zp0h3ftvqh6n9q1
[brigade] brigade-worker version: 1.0.0
build failed. (Build ID: 01dast93c17zp0h3ftvqh6n9q1)
```

I'm wondering if we can fix this such that the error message is logged by the worker.

radu-matei:
closes #913 

This PR logs the error message when an error is thrown but not handled.

One area I'd like to dig a bit deeper is around the different log levels, but for the time being, this is the easiest way to get the error to show up in the pod logs.

To test:

```
const { events } = require("brigadier")

events.on("exec", function() {
  throw new Error("This should be logged when failing the event handler.")
});
```

This generates:

```
[brigade] brigade-worker version: 1.0.0 
Unhandled rejection: Error: This should be logged when failing the event handler.
error Command failed with exit code 1
```

### DIFF
--- a/brigade-worker/src/app.ts
+++ b/brigade-worker/src/app.ts
@@ -130,6 +130,7 @@ export class App {
     // Run if an uncaught rejection happens.
     process.on("unhandledRejection", (reason: any, p: Promise<any>) => {
       this.logger.error(reason);
+      console.log("Unhandled rejection: " + reason);
       this.fireError(reason, "unhandledRejection");
     });
 


brigade-worker/src/app.ts
```diff
@@ -130,6 +130,7 @@ export class App {
     // Run if an uncaught rejection happens.
     process.on("unhandledRejection", (reason: any, p: Promise<any>) => {
       this.logger.error(reason);
+      console.log("Unhandled rejection: " + reason);
```
vdice: Just curious: do we know where the pre-existing `this.logger.error(reason);` is actually logged?

radu-matei: That's why I left the comment to investigate the log levels a bit more.

vdice: Ok, I have an update here.  While testing some work on the Brigade GitHub App, I noticed that in events created by the app (Provider: github), the log level was such that all of the `this.logger*` calls in this file _were_ being printed to the logs, whereas the are not by default when running events via `brig` (Provider: brigade-cli).

When I describe the corresponding pods for each, `BRIGADE_LOG_LEVEL` is empty for the former while it is `log` for the latter.

Therefore, using this hint for running our same test with `brig`, I can get the error to show properly if I set log level to an empty string (as it is when produced by Provider: github currently:

```
 $ brig run vdice/empty-testbed -l "" -f throw-error.js
Event created. Waiting for worker pod named "brigade-worker-01db3k51tbq748hrzckby1hhdg".
Build: 01db3k51tbq748hrzckby1hhdg, Worker: brigade-worker-01db3k51tbq748hrzckby1hhdg
[brigade] brigade-worker version: 1.0.0
[brigade:k8s] Creating PVC named brigade-worker-01db3k51tbq748hrzckby1hhdg
[brigade:app]   Error: This should be logged when failing the event handler.

  - script:4 EventRegistry.<anonymous>
    /etc/brigade/..2019_05_17_19_08_59.317237096/script:4:9

  - brigadier.js:34 Object.fire

    /home/src/dist/app.js:157:23


[brigade:app] error handler is cleaning up
[brigade:k8s] Destroying PVC named brigade-worker-01db3k51tbq748hrzckby1hhdg
```

To be noted, if `brig`'s log level is set to _anything_ besides `""`, we don't see the logs.  For instance, you'd think `error` would show them:
```
 $ b run vdice/empty-testbed -l error -f throw-error.js
Event created. Waiting for worker pod named "brigade-worker-01db3kf62vv6grqpgse7sv4y99".
Build: 01db3kf62vv6grqpgse7sv4y99, Worker: brigade-worker-01db3kf62vv6grqpgse7sv4y99
[brigade] brigade-worker version: 1.0.0
```

So perhaps the real issue is in `brig`?  Somehow the log levels not taking affect?  And/or why an empty string might default to actually showing more logs?

radu-matei: I would absoutely be in favour for a real fix rather than an additional `console.log` - so feel free not to merge this if you have the fix, I'll probably _not_ get around to this for at least another week..

vdice: A lot of time spent on investigation and understanding the flow of things, but I believe https://github.com/brigadecore/brigade/pull/915 is the fix we're looking for.  Let me know what you think, @radu-matei

PR #920, opened 2019-05-29
improve the build and release processes

### LINKED ISSUE #918: proposal: make images smaller
Currently, we're basing images for binaries built with Go on Alpine, but other than using its root CAs, I'm not sure why we need it.

It should be pretty easy to build `FROM scratch` and simply add root CAs.

If / when we transition to multi-stage Dockerfile builds, this is _especially_ easy to do by copying the root CAs over from a previous stage.

### LINKED ISSUE #919: proposal: track vendored code
This is increasingly recognized as a good practice, especially when using dep.

A huge benefit to this is that `make bootstrap` goes away. When you check out a branch / tag / commit, or whatever, you just automatically have all the right dependencies and a build can proceed with no fuss.

It does necessitate validating, as part of CI, that vendored code hasn't willfully or accidentally been modified, but `dep check` can easily do just that.

Not sure yet what this looks like for JS...

krancour:
Fixes #918 
Fixes #919 

Similar to https://github.com/brigadecore/brigade-github-app/pull/59

Highlights:

1. multi-stage builds for docker images
1. test targets consistently execute in a container
1. swap dood docker image builds for dind
1. track vendored code (both go and js)
1. base images on `scratch`

I tested this end-to-end in a fork. Tested PRs, pushes to branches, merges to master, semver tagged releases.

I also dog-fooded and tested all of the above using the new, smaller images that this PR creates.

~~Note: I'd like #923 merged first, then I need to rebase this.~~ This is all set.


brigade-controller/Dockerfile
```diff
@@ -1,7 +1,16 @@
...
+COPY pkg/ pkg/
+COPY vendor/ vendor/
+RUN go build -ldflags "$LDFLAGS" -o bin/brigade-controller ./brigade-controller/cmd/brigade-controller
 
-RUN apk add --no-cache ca-certificates && update-ca-certificates
-
-COPY rootfs/brigade-controller /usr/bin/brigade-controller
-
-CMD /usr/bin/brigade-controller
+FROM scratch
```
vdice: I think we may need to bring in a `/tmp` dir for this image as well (as is done in `brigade-vacuum`'s).

As it stands, I'm seeing the ctlr pod CrashLoopBackOff w/:
```
 $ k logs brigade-brigade-ctrl-7c5495577d-kkq2v
W0531 16:21:32.534596       1 client_config.go:533] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
log: exiting because of error: log: cannot create log: open /tmp/brigade-controller.brigade-brigade-ctrl-7c5495577d-kkq2v.unknownuser.log.WARNING.20190531-162132.1: no such file or directory
```

vdice:
Re: the changes to the `brigade.js` file (unfortunately, I don't see a way to see them as GH only shows the first 3000 files changed), it looks like a few more changes will be needed; the CI run for this PR shows a failure at the initial brigade-worker level (hence lack of updating check statuses here):
```
 $ k logs brigade-worker-01dc7bryaz2w8j0pnqsad6g04e
[brigade] brigade-worker version: 1.0.0
[brigade:k8s] Creating PVC named brigade-worker-01dc7bryaz2w8j0pnqsad6g04e
[brigade:app]   ReferenceError: checkRun is not defined

  - brigade.js:85 EventRegistry.runSuite
    /vcs/brigade.js:85:20

  - brigadier.js:34 Object.fire

    /home/src/dist/app.js:157:23


[brigade:app] error handler is cleaning up
[brigade:k8s] Destroying PVC named brigade-worker-01dc7bryaz2w8j0pnqsad6g04e
```
(Need to define `checkRun` and/or replace w/ a diff strategy)

Makefile
```diff
@@ -1,79 +1,185 @@
...
+BRIGADE_VERSION := $(VERSION)
+
+# DOCKER_IMAGE_PREFIX would have a trailing slash that we won't want here
+# because the charts glue the registry (registry + org, really), image name.
+# and tag together themselves and do not account for the possibility of registry
+# or org being "". So, we create a new variable here called BRIGADE_REGISTRY
+# that is DOCKER_IMAGE_PREFIX minus any trailing slash. If DOCKER_IMAGE_PREFIX
+# is the empty string, we default to "brigadecore", since, again, the charts
+# cannot currently accommodate a blank registry or org.
+ifeq ("$(DOCKER_IMAGE_PREFIX)","")
```
vdice: I'm wondering if we should add a note around needing to set `DOCKER_ORG` to `brigadecore` when running `make build-all-images` such that the images can be pulled when/if moving on to `make helm-install/upgrade test-functional`, etc.?  Or, I suppose, more generally, needing to set `DOCKER_ORG` to _something_...

(My particular use case was running/using the minikube docker daemon but, when no DOCKER_REGISTRY/ORG vals were set in my env, `make build-all-images` produced the bare, non-org'd images, e.g., `brigade-worker:edge`, etc. -- however, the helm release defaults to `brigadecore/brigade-worker:edge`)

krancour:
@vdice... will respond that other feedback on Monday, but in the meantime... I did deliberately break this PR into two commits where the addition of vendored code should be isolated to just one of them. i.e. You can look at the other commit (instead of the PR as a whole) to see _just_ the reviewable changes. lmk if that doesn't work out as I think it should.

radu-matei:
What is the size difference between `scratch` and `alpine`?

Is the 5.5 MB difference worth? Would we ever need to exec into a component and install something to debug?

krancour:
> Is the 5.5 MB difference worth? Would we ever need to exec into a component and install something to debug?

🤷‍♂ Perhaps not. But, reduced image size isn't the only benefit to scratch images. There's a substantially smaller attack surface.

brigade-controller/Dockerfile
krancour: @vdice, this is what the size of those dirs looks like:

```
156M	bin/
128K	brig/
 16K	brigade-api/
 72K	brigade-controller/
 20K	brigade-cr-gateway/
 24K	brigade-generic-gateway/
 28K	brigade-vacuum/
153M	brigade-worker/
3.6M	docs/
 32K	git-sidecar/
476K	pkg/
 20K	scripts/
 12K	tests/
101M	vendor/
```

Your idea would work, but would drag along the very large `vendor/` dir even for images that don't involve any Go code being built-- e.g. the worker. I think maybe what we can do it `.gitignore` `.dockerignore` (for good measure) and then copy an image-specific `.dockerignore` from each component's directory over to the root of the project right before the build is kicked off.

Regardless of your strategy or mine, one minor complication will be that building multiple images cannot be parallelized because there will be a race over the contents of `.dockerignore` in the root or the project. Of course, this is a bridge we could cross if and when we come to it. `make build-all-images` builds images serially right now.

krancour: > I think we may need to bring in a /tmp dir for this image as well (as is done in brigade-vacuum).

Thought I did. I will check on that.

vdice:
>  I did deliberately break this PR into two commits

Oh!  Great thinking; missed this.  Will re-review non-vendor files.  Thank you!

brigade-controller/Dockerfile
krancour: > I think we may need to bring in a /tmp dir for this image as well (as is done in brigade-vacuum).

@vdice not sure what happened that you are seeing that error. I've double-checked that the new image works. The `--logtostderr` seems to be honored by glog, so writes to `/tmp` aren't coming into play.

Can you double check that this is still happening to you?

krancour: I also just pushed some new commits and one of them reduces the build context for each image, as you suggested @vdice

Makefile
krancour: `checkRun` was blindly copied from the github gw's `brigade.js`. Whoops. This is fixed in a recent commit.

krancour: > make build-all-images produced the bare, non-org'd images, e.g., brigade-worker:edge, etc.

We had discussed this briefly... this was a deliberate choice to reduce (albeit not eliminate) the possibility of individuals with write permissions to the official Docker repo from accidentally building and pushing images from local source. I stand by this being a good idea...

But, I think you're absolutely right that some doc updates, at least are in order, and that maybe the `helm-upgrade` make target should be updated to honor the same env vars. I'm on it...

krancour: Or, actually, in your specific case, you can probably skip the push and not even bother at all with setting `DOCKER_ORG` at all since you're building images and and pulling them in the same VM.

Just `make build-all-images helm-upgrade` should work, I think. lmk if I am wrong. It happens frequently.

Definitely need to (and will) document this better.

brigade-controller/Dockerfile
vdice: Very nice -- the build context solution added in https://github.com/brigadecore/brigade/pull/920/commits/ce72e8f738662e1695b6f0c6883d7aa6ce18768e is _much_ more elegant than my suggestion above.  👍

Makefile
krancour: Updated docs in a recent commit.

vdice: > Just make build-all-images helm-upgrade should work

This doesn't quite work on account of the `brigade` chart concatenating values to construct the full image:tag string (and so, always prefacing an un-org'd image name with a `/`, as you pointed out in comments, I believe.)  However, the added/updated docs give good examples of how to run these targets with `DOCKER_REGISTRY`/`DOCKER_ORG` overrides, so I think we're good.

Makefile
```diff
@@ -1,79 +1,185 @@
...
-# Cross-compile for Docker+Linux
-build-docker-bins: $(addsuffix -docker-bin,$(BINS))
+.PHONY: format
+format: format-go format-js
 
-%-docker-bin: vendor
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -ldflags '$(LDFLAGS)' -o ./$*/rootfs/$* ./$*/cmd/$*
+.PHONY: format-go
+format-go:
+	$(GO_DOCKER_CMD) sh -c "go list -f '{{.Dir}}' ./... | xargs goimports -w -local github.com/brigadecore/brigade"
```
vdice: I think this is the last outstanding item: bumping the `GO_DEV_IMAGE` to a version that includes `goimports` so that this command can run successfully.

krancour: @vdice covered in the latest commit.

krancour:
Ok...

So before we merge this... both the Go and JS dependency trees here are quite large. While I am generally _in favor_ of tracking vendored code (as we discussed this morning), I am forced to reconsider that one aspect of this PR.

```
$ du -sh vendor/
101M	vendor/
$ du -sh brigade-worker/node_modules/
152M	brigade-worker/node_modules/
```

@vdice @radu-matei thoughts?

vdice:
> both the Go and JS dependency trees here are quite large.

I'm fine with this change, although I admit I don't have a suitable reference point.  This project is hit with the double whammy of having both Go deps and JS deps.  Other Go-based projects appear to have `vendor/` dirs of ~50MB.  Perhaps there is the possibility of vendor/Go-dep tuning in a follow-up PR?

krancour:
> I'm fine with this change, although...

@vdice I still do prefer this, tbh. Just wanted to raise it.

Over in kashti, if I am not mistaken, we're looking at a whopping 450MB of JS dependencies. Kinda thinking it's better not to track _those_ (I think GH prefers repos stay under 1GB). Not tracking here would have been a play at consistency, but now that I am rethinking it, we shouldn't do something a non-preferred way here only because we had to elsewhere.

Ok... I'm convinced to leave this be.

PR #923, opened 2019-05-30
make runSuite() fail build when any job fails

### LINKED ISSUE #921: bug: project's own brigade.js swallows job failures
The `notificationWrap(...)` helper sandwiches a job between two notifications to Github. If the sandwiched job fails, the second notification communicates that failure to Github, _but as long as that second notification succeeds, the promise that is returned resolves (it's from the second notification; not from the failed job)._

Even if this is corrected, another problem is lurking: the `runSuite(..)` handler is also swallowing errors-- logging them, but not propagating them.

Bottom line is that these two issues currently stand in the way of a build failing when the go or JS unit tests fail.

krancour:
Fixes #921

### DIFF
--- a/brigade.js
+++ b/brigade.js
@@ -59,10 +59,24 @@ function jsTest(e, project) {
 // Here we can add additional Check Runs, which will run in parallel and
 // report their results independently to GitHub
 function runSuite(e, p) {
+  // Important: To prevent Promise.all() from failing fast, we catch and return
+  // all errors. This ensures Promise.all() always resolves. We then iterate
+  // over all resolved values looking for errors. If we find one, we throw it
+  // so the whole build will fail.
+  //
+  // Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all#Promise.all_fail-fast_behaviour
+  //
   // Note: as provided language string is used in job naming, it must consist
   // of lowercase letters and hyphens only (per Brigade/K8s restrictions)
-  checkRun(e, p, goTest, "go").catch(e  => {console.error(e.toString())});
-  checkRun(e, p, jsTest, "javascript").catch(e  => {console.error(e.toString())});
+  return Promise.all([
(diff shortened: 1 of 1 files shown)


brigade.js
```diff
@@ -59,10 +59,24 @@ function jsTest(e, project) {
...
+  // over all resolved values looking for errors. If we find one, we throw it
+  // so the whole build will fail.
+  //
+  // Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all#Promise.all_fail-fast_behaviour
+  //
   // Note: as provided language string is used in job naming, it must consist
   // of lowercase letters and hyphens only (per Brigade/K8s restrictions)
-  checkRun(e, p, goTest, "go").catch(e  => {console.error(e.toString())});
-  checkRun(e, p, jsTest, "javascript").catch(e  => {console.error(e.toString())});
+  return Promise.all([
```
vdice: Can/should we use brigadier's `Group.runAll(..)` here?

technosophos: I _think_ that either way should work.

krancour: No... `Group.runAll(...)` runs _jobs_, but a call to `checkRun(...)` does not return a job; rather it returns a promise chain that is the execution of multiple jobs (a notification sandwhich) strung together.

vdice: Ah, indeed.  This makes sense. 👍 

I'm eager to brainstorm/iterate on this updated/corrected logic and see how we can bundle it up (or, abstract the details) for ease of use for brigade.js authors wanting to run multiple check runs (via notification sandwiches 🍔).

I believe we hope to add the `Notification` class and `notificationWrap` helper (perhaps even the `checkRun` helper) as re-usable library code once https://github.com/brigadecore/brigade-utils gets off the runway... when we do so, perhaps we can revisit the previously-established paradigm and find the best way to wrap jobs that require updating PRs via the Checks API and reduce error handling overhead, etc. to a minimum.

Which is to say, thanks for making this right and provoking further thoughts on refinement/usability.

PR #929, opened 2019-06-07
make devx more win-friendly

### LINKED ISSUE #927: Local build targets / Build fails on Windows
#920 brought significant improvements to the build and release process.

However, there is one scenario where now using the Make targets to build locally became significantly more difficult is on Windows, including WSL.

on Windows:
```
make -f .\Makefile build-brig
docker run -it --rm -e SKIP_DOCKER=true -v $(pwd):/go/src/github.com/brigadecore/brigade -w /go/src/github.com/brigadecore/brigade quay.io/deis/lightweight-docker-go:v0.7.0 bash -c 'LDFLAGS="-X github.com/brigadecore/brigade/pkg/version.Version=5559a34-dirty" scripts/build-brig.sh'
C:\Program Files\Docker\Docker\Resources\bin\docker.exe: Error response from daemon: Mount denied:
The source path "C:/projects/go/src/github.com/brigadecore/brigade;C"
doesn't exist and is not known to Docker.
See 'C:\Program Files\Docker\Docker\Resources\bin\docker.exe run --help'.
make: *** [build-brig] Error 125
```

Because the build target is using `$(pwd)`, this fails on Windows, and because WSL can't mount directories in Docker properly, if you're on Windows it means the current build process doesn't work.

That being said, I used to use the local targets to build and test components locally quite a bit, `make brig` in particular, and generally without Docker being a requirement.

Apologies, I should have tested this as part of the the PR review for #920.

cc @krancour, @vdice

krancour:
Fixes #927 

@radu-matei can you please lmk how this works for you?

__Update:__ I've confirmed that these changes work across Mac OS, Windows (Powershell), and WSL.


vdice:
Awesome!  Thanks for theses updates @krancour .

@radu-matei  to help confirm.

radu-matei:
Not relying on `pwd` to get the working directory is awesome.

That being said, the way the Makefile is structured, the build for Windows still assumes WSL is configured, or requires Docker builds - and I'm not particularly happy with either requirement, especially when the build should work without.

krancour:
I have to be honest in saying I don't generally understand the aversion to containerized builds. My instinct (which can be wrong, of course) says that it's a positive thing for all contributors to have their builds, tests, lint checks, etc, etc all occur in a consistent sandbox that we control. Beyond the dev environment being the same from one contributor to the next, ensuring local execution of build, tests, etc. occurs within a container, more significantly, means developers are utilizing the same containers for these tasks that our CI pipelines utilize-- meaning there's a pretty reasonable guarantee that if a build, test, etc. passes locally, it will also pass CI. These sort of guarantees don't exist if we pass on the opportunity to isolate these tasks to a sandbox.

krancour:
I'm going to merge this... whatever controversy there may be over the notion of containerized builds, it isn't introduced by this PR; rather this PR adds docs, fixes some bugs, and improves the situation overall.

Issue #917, opened 2019-05-22
PVC creation failure for job caches


radu-matei:
Trying to run a script with a job cache fails:

```
const { events, Job } = require("brigadier")

events.on("exec", (e) => {
    var job = new Job("cacher", "alpine:3.7")
    job.cache.enabled = true

    job.tasks = [
        "echo " + e.buildID + " >> /mnt/brigade/cache/jobs.txt",
        "cat /mnt/brigade/cache/jobs.txt"
    ]

    job.run()
})

```

```
     message: 'PersistentVolumeClaim "-cacher" is invalid: metadata.name: Invalid value: "-cacher": a DNS-1123 subdomain must consist of lower case alphanumeric characters, \'-\' or \'.\', and must start and end with an alphanumeric character (e.g. \'example.com\', regex used for validation is \'[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*\')',

```

radu-matei:
Related - function to get the cache PVC name (from [here](https://github.com/brigadecore/brigade/blob/master/brigade-worker/src/k8s.ts#L467)):

```
  /**
   * cacheName returns the name of this job's cache PVC.
   */
  protected cacheName(): string {
    // The Kubernetes rules on pvc names are stupid^b^b^b^b strict. Name must
    // be DNS-like, and less than 64 chars. This rules out using project ID,
    // project name, etc. For now, we use project name with slashes replaced,
    // appended to job name.
    return `${this.project.name.replace(/[.\/]/g, "-")}-${
      this.job.name
      }`.toLowerCase();
  }
```

dgkanatsios:
I've isolated the error, seems to be caused by [this line](https://github.com/brigadecore/brigade/blob/502eb908975c4cc3d516708c7a0adcb422845c75/brigade-worker/src/k8s.ts#L1009). Is there any reason that project name is set to the repo name? I'd be tempted to replace it with `secret.metadata.annotations["projectName"]`. Any objections?

dgkanatsios:
Fixed by #892

PR #936, opened 2019-06-18
fix(brigade-worker): ensure backwards compatibility for project name and repo name values

vdice:

This PR restores the legacy values for project name and repo name for a VCS project, albeit somewhat reluctantly.  The values were updated, logically so, in https://github.com/brigadecore/brigade/pull/892.  However, for users who may be relying on the the previous/legacy values for these two, say, in corresponding `brigade.js` files for a given project, breaking changes may result.

As a concrete example, if a URL in a `brigade.js` file is derived using `project.repo.name`, the legacy value of this interpolated string (restored in this PR) would have been `myorg/myrepo`, whereas in 892, it would be `github.com/myorg/myrepo`, thus invalidating the URL.  

If we feel strongly about following the (sane) precedent/updates of 892 at some point in the future, we'd have to point out the breaking changes to users (and thus might strictly require a `v2.x.x` release.)

### DIFF
--- a/brig/cmd/brig/commands/project_create_no_vcs.go
+++ b/brig/cmd/brig/commands/project_create_no_vcs.go
@@ -87,7 +87,4 @@ func setDefaultValuesNoVCS(p *brigade.Project) {
 	p.Name = "myproject"
 	// setting the sidecar to NONE
 	p.Kubernetes.VCSSidecar = "NONE"
-	// empty values for the repo
-	p.Repo.CloneURL = ""
-	p.Repo.Name = ""
 }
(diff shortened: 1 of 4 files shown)


vdice:
Ok, @dgkanatsios , the most recent commit pushes changes to brig's no-VCS project create logic so that the logic in brigade worker (determining whether or not to set project name to the legacy repository value) works properly (and passes the test in https://github.com/brigadecore/brigade/issues/917).  Let me know what you think.

dgkanatsios:
LGTM!

Issue #879, opened 2019-04-06
End to end testing for Brigade


radu-matei:
I think it's time we started end-to-end tests for Brigade. To achieve this, we first need to build and push container images for Brigade components for all pull requests (#761).

Then, I think we have the following approaches we could take:

- we can deploy Brigade in a new namespace, following [this really cool article on how to create preview environments with Brigade](https://medium.com/devopslinks/development-in-the-cloud-4aa2cabd3880). I think this would be the most straightforward and fastest way to run the tests, since the Kubernetes cluster would already be provisioned and running. However, the tests would be restricted to the Kubernetes version running on our Brigade instance.

- use `kind` to create a cluster and configure Brigade. This has the advantage of allowing us to easily test against multiple Kubernetes versions (at least whatever versions `kind` supports), but the disadvantage of not running in a real-world environment.

- create a fully fledged cloud-provided Kubernetes cluster (AKS/EKS/GKE), then deploy Brigade (optionally, we could do this using CNAB). For this, we would need knowledge and credits for various cloud platforms (we already have these for AKS).

Then, once we have a cluster / namespace and a Brigade instance, we can define some projects and run jobs.

Thoughts?

dgkanatsios:
I've been playing a bit with [kind](https://github.com/kubernetes-sigs/kind) and I really like it.
[Here's](https://github.com/dgkanatsios/azuregameserversscalingkubernetes/tree/master/e2e) what I've been doing with my project. Steps:

- use the [run.sh](https://github.com/dgkanatsios/azuregameserversscalingkubernetes/blob/master/e2e/run.sh) to `kubectl create` my Deployments. Similarly, we could use this to `helm install Brigade`
- wait till these Deployments are up and running with [this](https://github.com/dgkanatsios/azuregameserversscalingkubernetes/blob/master/e2e/wait-for-deployment.sh) script
- then, I do have [this](https://github.com/dgkanatsios/azuregameserversscalingkubernetes/tree/master/e2e/cmd) app that basically creates stuff in Kubernetes and then waits (well, polls) till they are up and running and reports accordingly. We could use a similar app to
   - create a bunch of Projects (by creating Secrets). 
   - Trigger some Builds on them and wait for the expected outcome
   - we could probably test Generic/GitHub Gateway as well by issuing expected HTTP requests and wait for the expected outcome

In any case, if we go down this way (since I think that this is huge), we should begin with a simple skeleton and then built on top of it. Should we make the e2e tests run on every PR?

[...]

PR #937, opened 2019-06-19
added instructions to develop with kind

dgkanatsios:

This PR includes instructions for Brigade development with [kind](https://github.com/kubernetes-sigs/kind). Same commands will also be used in building [e2e tests for Brigade](https://github.com/brigadecore/brigade/issues/879).

I added a line on Makefile to `kind load` images, this should run only if kind is installed. Is there a better way to do that?

**If applicable**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,10 @@ ifneq ($(SKIP_DOCKER),true)
 		-w /code/$(BASE_PACKAGE_NAME) $(JS_DEV_IMAGE)
 endif
 
+# Allow for users to supply a different helm cli name,
+# for instance, if one has helm v3 as `helm3` and helm v2 as `helm`
+HELM ?= helm
+
 ################################################################################
 # Binaries and Docker images we build and publish                              #
 ################################################################################
@@ -141,6 +145,14 @@ build-all-images: $(addsuffix -build-image,$(IMAGES))
 		.
 	docker tag $(DOCKER_IMAGE_PREFIX)$*:$(IMMUTABLE_DOCKER_TAG) $(DOCKER_IMAGE_PREFIX)$*:$(MUTABLE_DOCKER_TAG)
 
+.PHONY: load-all-images
+load-all-images: $(addsuffix -load-image,$(IMAGES))
+
+%-load-image:
+	@echo "Loading $(DOCKER_IMAGE_PREFIX)$*:$(IMMUTABLE_DOCKER_TAG)"
+	@kind load docker-image $(DOCKER_IMAGE_PREFIX)$*:$(IMMUTABLE_DOCKER_TAG) \
+			|| echo >&2 "kind not installed or error loading image: $(DOCKER_IMAGE_PREFIX)$*:$(IMMUTABLE_DOCKER_TAG)"
+
 # Cross-compile binaries for brig
 build-brig:
 	$(GO_DOCKER_CMD) bash -c 'LDFLAGS="$(LDFLAGS)" scripts/build-brig.sh'
@@ -187,7 +199,7 @@ helm-install: helm-upgrade
 
 .PHONY: helm-upgrade
 helm-upgrade:
-	helm upgrade --install $(BRIGADE_RELEASE) brigade/brigade --namespace $(BRIGADE_NAMESPACE) \
+	$(HELM) upgrade --install $(BRIGADE_RELEASE) brigade/brigade --namespace $(BRIGADE_NAMESPACE) \
 		--set brigade-github-app.enabled=true \
 		--set controller.registry=$(BRIGADE_REGISTRY) \
 		--set controller.tag=$(BRIGADE_VERSION) \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -140,6 +140,7 @@ build-all-images: $(addsuffix -build-image,$(IMAGES))
 		--build-arg LDFLAGS='$(LDFLAGS)' \
 		.
 	docker tag $(DOCKER_IMAGE_PREFIX)$*:$(IMMUTABLE_DOCKER_TAG) $(DOCKER_IMAGE_PREFIX)$*:$(MUTABLE_DOCKER_TAG)
+	kind load docker-image $(DOCKER_IMAGE_PREFIX)$*:$(IMMUTABLE_DOCKER_TAG) || echo >&2 "kind not installed or error loading image"
```
vdice: I'd definitely prefer a separate set of targets for loading images into a kind cluster.  How about `load-all-images` and `%-load-image` for each?  I was playing around a little and came up with this:
```
.PHONY: load-all-images
load-all-images: $(addsuffix -load-image,$(IMAGES))

%-load-image:
	@echo "Loading $(DOCKER_IMAGE_PREFIX)$*:$(IMMUTABLE_DOCKER_TAG)" && \
		KUBECONFIG="$(shell kind get kubeconfig-path --name="kind")" \
			kind load docker-image $(DOCKER_IMAGE_PREFIX)$*:$(IMMUTABLE_DOCKER_TAG) \
				|| echo >&2 "kind not installed or error loading image: $(DOCKER_IMAGE_PREFIX)$*:$(IMMUTABLE_DOCKER_TAG)"
```

There I added an informative echo, automatically switching kube context to kind (that way, I can load images to my kind cluster even if my current context is on a different one.  Maybe this is overkill/not needed... or perhaps it isn't Windows compatible?  Feel free to ignore, if so), and then run the `kind load` command.

(The `@` symbol prefixing the command string just silences printing out the actual command string when running)

Either way, if this approach sounds good, we can update the docs, e.g, `... make build-all-images load-all-images`

docs/content/topics/developers.md
```diff
@@ -262,6 +262,18 @@ $ helm install -n brigade brigade/brigade -f myvalues.yaml
 
 Don't forget to also create a project.  Check out [projects](./projects.md) to see how it's done.
 
+## Developing brigade with kind
+
+You can also use [kind](https://github.com/kubernetes-sigs/kind) for your day to day Brigade development workflow.
+
+- Run `kind create cluster` to create the cluster 
+- Run `export KUBECONFIG="$(kind get kubeconfig-path --name="kind")"`
+- Run `helm init` to install helm on the `kind` cluster. The latest [Helm 3](https://github.com/helm/helm) version is recommended. However, if you're running Helm 2, check [here](https://helm.sh/docs/using_helm/#role-based-access-control) for proper Helm/Tiller installation instructions
```
vdice: Thank you for providing the excuse to play around with helm 3!

Since I still require helm v2 to interact with other clusters, I placed helm v3 into `helm3` locally.  Is this a common use case?  If so, I wonder if it'd be worth a follow-up PR enabling switching out helm versions for the `make helm-install` command... something like:
```
# Allow for users to supply a different helm cli name,
# for instance, if one has helm v3 as `helm3` and helm v2 as `helm`
HELM ?= helm
...
helm-upgrade:
	$(HELM) upgrade --install $(BRIGADE_RELEASE) brigade/brigade --namespace $(BRIGADE_NAMESPACE) \
...
```

docs/content/topics/developers.md
```diff
@@ -262,6 +262,18 @@ $ helm install -n brigade brigade/brigade -f myvalues.yaml
 
 Don't forget to also create a project.  Check out [projects](./projects.md) to see how it's done.
 
+## Developing brigade with kind
+
+You can also use [kind](https://github.com/kubernetes-sigs/kind) for your day to day Brigade development workflow.
```
radu-matei: Could you please also point to the Kind quickstart here?
Thanks!

docs/content/topics/developers.md
radu-matei: Oh, this is on me and Ralph... we pointed Dimitris to Helm 3 for this.
Can we maybe have two targets? I really think it's valuable for people to get familiar with Helm 3, but still provide the Helm 2 target.
WDYT?

Makefile
dgkanatsios: Great addition, thanks! I copied this to the Makefile, works like a charm. Regarding the KUBECONFIG set on `%-load-image`, not sure if it's needed, since you should set the KUBECONFIG before calling `load-all-images`, provided that you follow the instructions, of course.

docs/content/topics/developers.md
dgkanatsios: I opted for @vdice's option, since it's easy to change the variable on the Makefile and by adding more targets we would have 4 in total: `helm2-install`, `helm2-upgrade`, `helm3-install`, `helm3-upgrade`.

docs/content/topics/developers.md
dgkanatsios: done, thanks!

Issue #941, opened 2019-06-20
brig: rerun should have the same flags as run


itowlson:
`brig run` accepts the flags `--no-color --no-progress --background` but these are not available on `brig rerun`.  It seems like we should support most of the `run` flags on `rerun` too.

PR #948, opened 2019-07-03
feat(brig): achieve flag parity for rerun command (to run); add unit tests

vdice:

Proposed implementation to address https://github.com/brigadecore/brigade/issues/941, including refactoring for unit testability.

**Special Notes for the Reviewer**:

I went the route of using the same set of flags for `run` and `rerun`, as I didn't see harm in enabling overriding any of the `run` flags/fields for `rerun`.  Open to suggestions/revisiting if others have feedback.

**If applicable**:

Closes https://github.com/brigadecore/brigade/issues/941

### DIFF
--- a/brig/cmd/brig/commands/rerun.go
+++ b/brig/cmd/brig/commands/rerun.go
@@ -2,11 +2,14 @@ package commands
 
 import (
 	"errors"
+	"io"
 	"io/ioutil"
 	"os"
 
 	"github.com/spf13/cobra"
 
+	"github.com/brigadecore/brigade/pkg/brigade"
+	"github.com/brigadecore/brigade/pkg/decolorizer"
 	"github.com/brigadecore/brigade/pkg/script"
 )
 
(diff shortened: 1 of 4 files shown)


brig/cmd/brig/commands/rerun.go
```diff
@@ -17,22 +20,39 @@ must still be accessible. Once a build is vacuumed, it can no longer be re-run.
...
 
 func init() {
+	rerun.Flags().StringVarP(&rerunFile, "file", "f", "", "The JavaScript file to execute")
+	rerun.Flags().StringVarP(&rerunEvent, "event", "e", "", "The name of the event to fire")
+	rerun.Flags().StringVarP(&rerunPayload, "payload", "p", "", "The path to a payload file")
+	rerun.Flags().StringVarP(&rerunCommitish, "commit", "c", "", "A VCS (git) commit")
+	rerun.Flags().StringVarP(&rerunRef, "ref", "r", "", "A VCS (git) version, tag, or branch")
+	rerun.Flags().BoolVar(&rerunNoProgress, "no-progress", runNoProgress, "Disable progress meter")
+	rerun.Flags().BoolVar(&rerunNoColor, "no-color", runNoColor, "Remove color codes from log output")
+	rerun.Flags().BoolVarP(&rerunBackground, "background", "b", runBackground, "Trigger the event and exit. Let the job rerun in the background.")
```
mumoshu: I think we'd better add a utility func that adds these flags to a given flagset, so that the func can be used from here and [run.go](https://github.com/brigadecore/brigade/blob/master/brig/cmd/brig/commands/run.go#L53) to reduce code duplication.

But that's an another story 😉

vdice: Yes, good point.  Though there is a subtle difference in some of the flag's default values between rerun and run (for instance, we want the default rerun ref, commit, etc. to be empty so as not to override the initial run values), this perhaps could be hashed out in said utility function.  I agree we can save this for a later PR, ideally with retrofitting run.go with unit tests...

PR #955, opened 2019-07-19
ci(brigade.js): add e2e job

vdice:
**Update 10/28/19  The various kinks/issues seem to have been ironed out and, as seen in the Check suite for this PR, the e2e job is running successfully.  This PR is now ready for review.**

This is where I'm at with adding @dgkanatsios 's e2e tests to CI.  The only real change to enable running in a container was the bin dir setup.

As kind requires a Docker daemon to run (Kubernetes In Docker!), I went the route of running the tests in a Docker-In-Docker fashion, using an image that wraps docker:stable-dind with some utilities we have use for. (https://github.com/vdice/go-dind)

The part I'm stuck on is getting the kind cluster to properly launch in the context of a container _in a k8s pod_, as is the case when running via Brigade.

It variously fails on setup as in:
```
...
Creating cluster "kind" ...
 • Ensuring node image (kindest/node:v1.15.0) 🖼  ...
 ✓ Ensuring node image (kindest/node:v1.15.0) 🖼
 • Preparing nodes 📦  ...
time="2019-07-19T21:48:29.177656516Z" level=info msg="shim containerd-shim started" address="/containerd-shim/moby/62dc5e06a6120ab4ac36556056afd6f40d00cdf079c55f64a9b9a6eb0978da1d/shim.sock" debug=false pid=427
 ✓ Preparing nodes 📦
 • Creating kubeadm config 📜  ...
 ✓ Creating kubeadm config 📜
 • Starting control-plane 🕹️  ...
 ✓ Starting control-plane 🕹️
 • Installing CNI 🔌  ...
 ✗ Installing CNI 🔌
time="2019-07-19T21:49:25.656763378Z" level=info msg="shim reaped" id=62dc5e06a6120ab4ac36556056afd6f40d00cdf079c55f64a9b9a6eb0978da1d
time="2019-07-19T21:49:25.666962354Z" level=info msg="ignoring event" module=libcontainerd namespace=moby topic=/tasks/delete type="*events.TaskDelete"
Error: failed to create cluster: failed to apply overlay network: exit status 1
```
Or, the cluster creation may be successful but any attempts at contacting the api server fail... 

As a comparison, the kind cluster launches just fine (and e2e tests pass!) when running via the same Docker image directly (using Docker for Mac):

```
[...]

PR #966, opened 2019-08-04
 Add experimental volumes and volumeMounts properties to jobs

radu-matei:
depends on https://github.com/brigadecore/brigadier/pull/22

This PR enables experimental support for adding volumes and volume mounts to jobs. 

Note that for the simple use case of sharing a volume across all jobs, `storage.enabled` should still be used.

A pre-built image for the worker: `radumatei/brigade-worker:volume-mounts-check`.

cc @krancour - this is a very Kubernetes-specific addition - how would this impact non-Kubernetes targets?

Use cases: https://github.com/brigadecore/brigade/pull/955#issuecomment-518002970 and  [Mounting volumes for ETL jobs](https://hackmd.io/@-IwbEKEMQBWBBEpIBone5g/Bk5zvEJEV)

Example:

```
    const kind = new Job("kind", "radumatei/golang-kind:1.11-0.4")
    kind.privileged = true;
    kind.volumes = [
        {
            name: "modules",
            hostPath: {
                path: "/lib/modules",
                type: "Directory"
            }
        },
        {
            name: "cgroup",
            hostPath: {
                path: "/sys/fs/cgroup",
                type: "Directory"
            }
        },
        {
            name: "docker-graph-storage",
            emptyDir: {}
        }
    ];

    kind.volumeMounts = [
        {
            name: "modules",
            mountPath: "/lib/modules",
            readOnly: true
        },
        {
            name: "cgroup",
            mountPath: "/sys/fs/cgroup"
        },
        {
            name: "docker-graph-storage",
            mountPath: "/var/lib/docker"
        }
    ];
```


brigade-worker/src/k8s.ts
```diff
@@ -442,6 +442,16 @@ export class JobRunner implements jobs.JobRunner {
       }
     }
 
+    for (let cfg of job.volumeConfig) {
+      if (cfg.volume.hostPath != undefined && !project.allowHostMounts) {
+        throw new Error(`allowHostMounts is false in this project, not mounting path ${cfg.volume.hostPath.path}`)
+      }
+      this.runner.spec.volumes.push(cfg.volume);
```
radu-matei: I think this should check for unique volumes?

What is the default Kubernetes behaviour if:
- a volume name is present multiple times
- a PVC is present multiple times
- a host path mount is present multiple times

mumoshu: I have never got to do it myself. But I see some use with subpath, as mentioned in the issue.

Just curious, but does whether to support mounting the same volume multiple times affect our design/implementation? 🤔

mumoshu: Ah I got it! You've paired a volume and its corresponding mount in each volumeConfig.

I'd rather suggest exposing `volumes` and `volumeMounts` without any abstraction first, so that we can provide a standard library feature like this `volumeConfig` on top of that, or users can build their own.

IMHO for the standard library, something that solves 80% of the use-cases will be enough, that justifies dropping support for mounting the same volume multiple times.

mumoshu: Reading https://github.com/radu-matei/brigadier/commit/94f56c2446c67e817e69730ea7cf6e9f2cdd3db0 - If you agree, implementation-wise, I think adding `volumeMounts` and `volumes` that are allowed/used only when `volumeConfig` doesn't exist would suffice.

radu-matei: That totally makes sense - I initially thought maybe we could simplify and boil things down to a single property, but yeah, I you're right, we should keep in line with the Kubernetes abstraction.

Updating.

radu-matei: I actually don't think we should keep both `volume` + `volumeMounts` and `volumeConfig`, as it just invites confusion - what do you think?

mumoshu: Btw I like the idea of pairing volumes and mounts.

I even think we can happily merge the volume and the mount as in many cases each volume is mounted only once:

```
job.volumeConfig = [
    {
        name: "modules",
        mountPath: "/lib/modules",
        readOnly: true,
        hostPath: {
            path: "/lib/modules",
            type: "Directory"
        }
    },

    {
        name: "cgroup",
        mountPath: "/sys/fs/cgroup",
        hostPath: {
            path: "/sys/fs/cgroup",
            type: "Directory"
        }
    },
    {
        name: "docker-graph-storage",
        mountPath: "/var/lib/docker",
        emptyDir: {}
    }
]
```

And I believe `name`s can be usually omitted because they can be generated from mount-path(which are guaranteed to not collide:

```
job.volumeConfig = [
    {
//        name: "lib-modules",
        mountPath: "/lib/modules",
        readOnly: true,
        hostPath: {
            path: "/lib/modules",
            type: "Directory"
        }
    },

    {
//        name: "sys-fs-cgroup",
        mountPath: "/sys/fs/cgroup",
        hostPath: {
            path: "/sys/fs/cgroup",
            type: "Directory"
        }
    },
    {
//        name: "var-lib-docker",
        mountPath: "/var/lib/docker",
        emptyDir: {}
    }
]
```

mumoshu: > I actually don't think we should keep both volume + volumeMounts and volumeConfig, as it just invites confusion - what do you think?

So I'd vote for starting with only `volumes` + `volumeMounts` if I need to choose one.

// But I do believe something like `volumeConfig` has a lot of potential, and worth it, even if it needs an extensive documentation to not confuse users.

radu-matei: Yeah, I initially thought that something like `volumeConfig` could be an abstraction over how volumes and volume mounts work in Kubernetes and simplify setting them up for a job.

But I agree, let's start with plain `volumes` and `volumeMounts`, then decide whether to add a higher level abstraction, how you suggested.

Thanks a lot, this has been really helpful!

radu-matei: Brigadier change: https://github.com/brigadecore/brigadier/pull/22/commits/8c53a1c91bf165c69ff8205712800f120223b82b

radu-matei: Updated - could you please give this another look?
Thanks!

brigade-worker/package.json
```diff
@@ -38,7 +38,7 @@
     "typescript": "^3.2.2"
   },
   "dependencies": {
-    "@brigadecore/brigadier": "^0.5.0",
+    "@brigadecore/brigadier": "radu-matei/brigadier#job-volume-config-out",
```
radu-matei: After https://github.com/brigadecore/brigadier/pull/22 is merged and a new version is published, update this.

brigade-worker/src/k8s.ts
```diff
@@ -12,6 +12,7 @@ import * as fs from "fs";
 import * as path from "path";
 import * as request from "request";
 import * as byline_1 from "byline";
+import { V1Volume } from "@kubernetes/client-node";
```
radu-matei: Remove this.

brigade-worker/src/k8s.ts
```diff
@@ -442,6 +443,23 @@ export class JobRunner implements jobs.JobRunner {
...
+    // and throw an error if it it does not.
+    for (let v of job.volumes) {
+      if (v.hostPath != undefined && !project.allowHostMounts) {
+        throw new Error(`allowHostMounts is false in this project, not mounting ${v.hostPath.path}`);
+      }
+      this.runner.spec.volumes.push(v);
+    }
+
+    // If the job defines volume mounts, add them to every container's spec.
+    for (let m of job.volumeMounts) {
```
vdice: Since `volumeMounts` and `volumes` appear to be un-coupled, should we add the same check here to be sure `project.allowHostMounts` is true?

And/or, in general, should we somehow check for proper API usage?  (If, say, configured with a volumeMount setup but no corresponding volume defined, etc. -- or is that allowed?)  I suppose as long as we can show the user the k8s library error that should result on mis-configuration, we'd be good.

radu-matei: Since you need to declare both volume and volume mount, I think checking is unnecessary - it would end up just taking the volume mount name and iterating through the volumes to check if it's mounting a host path - which already do.

As for checking the existence of the volume you reference in the volume mount, that is a good idea, I'll add that.

radu-matei: Added a check for the referenced volume, together with a test.

radu-matei: Pushed `radumatei/brigade-worker:volume-mounts-check` if you need to try it on a project.

mumoshu: I love the extra validation with `volumeExists` because that provides us a faster feedback loop 👍

vdice: Perfect, thanks Radu!

radu-matei:
Merged https://github.com/brigadecore/brigadier/pull/22, rebasing, updating, and adding a doc.

krancour:
> cc @krancour - this is a very Kubernetes-specific addition - how would this impact non-Kubernetes targets?

Do we imagine being able to one day orchestrate job containers using something other than Kubernetes? It wasn't on my radar, but even if the answer were yes, I think this could be accommodated with other orchestrators.

radu-matei:
Rebased and added a simple doc.

@krancour - not really, or at least not the JS worker anyway. Just trying to make sure we're not overlooking something that might be affected by this - thanks!

radu-matei:
Thanks a lot for the help, everyone!

Issue #938, opened 2019-06-19
Improve error reporting in brig


radu-matei:
For newly created events, if anything wrong is happening in the init container, you don't get any visibility into what is wrong - same goes for other types of failures, which forces the users to go searching through logs.

We could compile a list of common failures that can happen, and surface the relevant logs directly though `brig`.

mumoshu:
> We could compile a list of common failures that can happen, and surface the relevant logs directly though brig.

@radu-matei Can we start by just making `brig` stream last N lines(10 or so) of the init container logs if it failed?

Or do you have something more user-friendly in your mind? I'd appreciate an example of "a list of common failures" to see how granular each list item would be!

mumoshu:
@radu-matei I've submitted the MVP of this feature in #967 that just stream a brig flag to stream all the init container logs. Would you mind taking a look?

Maybe we can gradually enhance it by perhaps enabling it by default, adding a nice header line before the init container logs.

radu-matei:
^^ Tested and LGTM-ed, I think we can go ahead and merge it. I would keep this issue open though, until we are satisfied with the overall error reporting in `brig`, what do you think?

mumoshu:
I agree that we should keep this open 👍 

Also posting a link to my comment https://github.com/brigadecore/brigade/pull/967#discussion_r310600394 as a basis for discussion.

Enabling `--init` by default sounds good. But output from vcs-sidecar seems a bit verbose. Can we e.g. prevent every bash script line being printed when there's no error? (Not sure how though

vdice:
Removing from 1.3; we can consider for a future milestone if/when PRs appear.

krancour:
Closing due to staleness. Please do re-open if appropriate.

PR #967, opened 2019-08-05
feat(brig): option to show logs from vcs-sidecar

mumoshu:
You can now add `--init` like `brig build logs --init BUILD_ID` to view logs from the init container of a worker(=vcs-sidecar).

When the init-container failed due to reasons like a wrong ssh key or invalid git repo reference,
`brig build logs` has been giving us nothing useful.
So you had to run `kubectl logs -c vcs-sidecar $BRIGADE_WORKER` to finally catch it.

The `--init` flag just makes this process simple.
If your build failed, you always run `brig build logs --init` (or perhaps `brig build logs --init --jobs`) to catch any issues regardless of in which container in the worker pod the fatal error happened.

Ref #938

Adds the `--init` flag to `brig`.

This is the MVP or a foundational work to fully address #938

**If applicable**:

### DIFF
--- a/brig/cmd/brig/commands/build_logs.go
+++ b/brig/cmd/brig/commands/build_logs.go
@@ -22,11 +22,13 @@ build.
 )
 
 var (
+	logsInit bool
 	logsJobs bool
 	logsLast bool
 )
 
 func init() {
+	buildLogs.Flags().BoolVarP(&logsInit, "init", "i", false, "Show init container logs as well as the worker log")
 	buildLogs.Flags().BoolVarP(&logsJobs, "jobs", "j", false, "Show job logs as well as the worker log")
 	buildLogs.Flags().BoolVarP(&logsLast, "last", "l", false, "Show last build's log (ignores BUILD_ID)")
 	build.AddCommand(buildLogs)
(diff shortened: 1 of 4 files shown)


dgkanatsios:
Nice!

brig/cmd/brig/commands/build_logs.go
```diff
@@ -22,11 +22,13 @@ build.
 )
 
 var (
+	logsInit bool
 	logsJobs bool
 	logsLast bool
 )
 
 func init() {
+	buildLogs.Flags().BoolVarP(&logsInit, "init", "i", false, "Show init container logs as well as the worker log")
```
radu-matei: I'm wondering whether we should enable both `--init` and `--jobs` by default.

mumoshu: For `--init` I'm thinking that enabling it by default makes sense, as that's the only way to grab logs from vcs-sidecar, AND you have no way to access it from within worker/brigade.js.

For `--jobs`, it's really useful for debugging jobs. But usually any job failure should be handled correctly by `brigade.js` with appropriate error logs to `console`. With correctly authored `brigade.js`, `--jobs` by default may result in too verbose logs.

mumoshu:
Rebased.

brig/cmd/brig/commands/build_logs.go
mumoshu: That said - Please let me defer this to https://github.com/brigadecore/brigade/issues/938 for further discussion!

radu-matei: Absolutely, we shouldn't keep this PR from getting merged because of this.

mumoshu:
@vdice Yeah, the dedicated doc sounds good! I have no idea where to add it at the moment, so just created the issue #970 to not forget about it

Issue #886, opened 2019-04-10
document worker image requirements better


krancour:
This section covers custom worker images: https://docs.brigade.sh/topics/workers/

But it also seems to make an assumption that customizing a worker image is only accomplished by extending an existing image and adding in additional JS libraries and really only covers this scenario.

I've been creating a new worker from scratch (as part of an experiment to integrate with a new pipeline model) and it's been difficult for me to understand the expected inputs, outputs, and behaviors. If these expectations were documented somewhere, we open the door for a greater variety of custom workers that can bring their own unique innovations to the table.

krancour:
@radu-matei can you assign this issue to me, please?

krancour:
Starting a short list of things as I uncover them. I'll circle back and document later. Just don't want to lose these:

1. Document the environment variables that the brigade controller will inject into the worker. These are an implicit interface between controller and worker.
1. Document how custom workers should name and label any pods (and other things?) that it creates to ensure other Brigade and Kashti processes interact with them as expected. For instance, a pod needs to be labeled a certain way to count as a "job" in the eyes of the vacuum process, Kashti, etc.

Will add to this list as I uncover things.

radu-matei:
@krancour - I added this issue to the 1.2 milestone - feel free to remove it if you think it won't be done in time, or is not in scope.

radu-matei:
Closed in #968

PR #968, opened 2019-08-05
more thorough custom worker documentation

krancour:
Fiexes #886 

Note to reviewers-- probably don't try to look at the diffs here; rather look at the revised page as a whole.

### DIFF
--- a/docs/content/topics/workers.md
+++ b/docs/content/topics/workers.md
@@ -9,162 +9,57 @@ aliases:
   - /topics/scripting/workers.md
 ---
 
-# Adding custom libraries to a Brigade worker
+# What is a Brigade Worker?
 
-A worker is the component in Brigade that executes a `brigade.js`. Brigade ships
-with a general purpose worker focused on running jobs. This generic worker
-exposes a host of Node.js libraries to `brigade.js` files, as well as the
-`brigadier` Brigade library.
+A worker is a Brigade component that the Brigade controller launches in response
+to an event in order to execute a build in accordance with
+project/event-specific configuration or logic. There is a one-to-one
+relationship between events and workers.
 
-Sometimes, though, it is desirable to include additional libraries -- perhaps even
-custom libraries -- to your workers. 
-There are two ways of adding custom dependencies to a Brigade worker:
+_Typically_, the action of a worker is driven by the contents of project's
+`brigade.js` file, however, _custom_ workers aren't strictly limited to this
+approach and can utilize alternative mechanisms for defining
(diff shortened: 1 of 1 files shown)


docs/content/topics/workers.md
```diff
@@ -9,64 +9,225 @@ aliases:
...
+We can build this file into our `Dockerfile` by copying it into the image:
+
+```
+FROM brigadecore/brigade-worker:v1.1.0
 
 RUN yarn add xml-simple
+COPY mylib.js /home/src/dist
+```
+
+Next, skip to the section on [building and publishing a custom worker image](#building-and-publishing-a-custom-worker-image).
```
vdice: This looks like an errant copied line/section (first seen on line 62)?

krancour: I can see that it may appear that way, but it was deliberate that both of those spots have a "go to" that bypasses the next, significantly more advanced/involved section in order to get on with the business of building and pushing the image.

krancour:
I'm going to mark this WIP because some more research has revealed that my characterization of `BRIGADE_PROJECT_NAMESPACE` as being able to differ from the worker's own namespace was not factually correct.

docs/content/topics/workers.md
vdice: Ah gotcha... but if we already may've skipped to that section higher up (line 62), it may not be obvious when/where to return to and then skip again?  Wonder if we re-org a bit?

docs/content/topics/workers.md
```diff
@@ -9,64 +9,225 @@ aliases:
...
+const {Job} = require("./brigadier");
+
+exports.alpineJob = function(name) {
+  j = new Job(name, "alpine:3.7", ["echo hello"])
+  return j
+}
+```
+
+**Note:** Because we are loading our code directly into Brigade, we import
+`./brigadier`, not `brigadier`. This may change in the future.
```
radu-matei: Not sure I'm following this - the default implementation of the worker takes care of this, and allows importing `@brigadecore/brigadier` as well.

krancour: That line was copied verbatim from the previous version of this doc. This PR just moves it. I'm all for removing old/superfluous details. If you think this can go, I'm happy to remove it.

krancour:
@vdice @radu-matei https://github.com/brigadecore/brigade/pull/968/commits/430c22a6c5ed6c33ff93437560915655d06da577 contains a few small updates for accuracy and clarity.

PR #975, opened 2019-08-12
add support for optional build-level cloneURL field

### LINKED ISSUE #973: Proposal: Add cloneURL field to the Build struct
Today I was working on adding Bitbucket support to [brigdrake](https://github.com/lovethedrake/brigdrake) (which is really only a custom worker). I had trouble getting pull request workflows to work across forks. I did some digging...

The [brigade-bitbucket-gateway](https://github.com/lukepatrick/brigade-bitbucket-gateway) was emitting events/builds into Brigade with revision ref referencing a commit _from the fork_, however, when spawning the worker, the Brigade controller unconditionally sets the `BRIGADE_REMOTE_URL` environment variable of the "vcs sidecar" to the `cloneURL` field _from the project--_ i.e. the vcs sidecar _always_ references the upstream. The commit in question does not exist there. The vcs sidecar fails. Whoops!

(Aside: Nothing like this happens with the brigade-github-app gateway because the way GitHub handles pull requests, behind the scenes, is fundamentally different from how Bitbucket does.)

The fix I'd like to PR to the [brigade-bitbucket-gateway](https://github.com/lukepatrick/brigade-bitbucket-gateway) involves adding the clone URL for the fork to the event/build that it emits into Brigade, but for that to actually have any effect, some changes are required to Brigade itself...

What I'd like to propose is adding an _optional_ `cloneURL` field to the `Build` struct. When this is set, the Brigade controller will honor it when setting the `BRIGADE_REMOTE_URL` on the worker's vcs sidecar. If left unset, it will fall back to the value in the `cloneURL` field of the `Project`.

Last, note that this may happen to improve Brigade's Bitbucket integration significantly, but in general, this undoes a narrow assumption about _where_ the source code for a given build should be obtained from and puts more control over this into the hands of the gateways that are scheduling the builds. Generally, they will know more than Brigade itself about the right place (e.g. a fork) to get the code from.

krancour:
Fixes #973

This will give gateways the power to be more specific about where source code for a build is obtained from-- i.e. it creates the possibility of getting source code from forks. This isn't necessary for github (because of how github implements pull requests behind the scenes), but may be necessary for other VCS-- notably, bitbucket.

### DIFF
--- a/brigade-controller/cmd/brigade-controller/controller/handler.go
+++ b/brigade-controller/cmd/brigade-controller/controller/handler.go
@@ -245,6 +245,15 @@ func workerEnv(project, build *v1.Secret, config *Config) []v1.EnvVar {
 		serviceAccount = string(project.Data["serviceAccount"])
 	}
 
+	// Try to get cloneURL from the build first. This allows gateways to override
+	// the project-level cloneURL if the commit that should be built, for
+	// instance, exists only within a fork. If this isn't set at the build-level,
+	// fall back to the project-level default.
+	cloneURL := bsv.String("clone_url")
+	if cloneURL == "" {
+		cloneURL = string(project.Data["cloneURL"])
+	}
+
 	envs := []v1.EnvVar{
 		{Name: "CI", Value: "true"},
 		{Name: "BRIGADE_BUILD_ID", Value: build.Labels["build"]},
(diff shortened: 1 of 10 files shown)


krancour:
I have verified these changes by creating a test project in Bitbucket as well as a fork of that project, then forked the bb-gw to add support for this new, optional build-level field. Everything worked as expected.

I have also verified that all current functionality is preserved. i.e. When this new, optional build-level field is not set, things still work the same as they always have.

krancour:
Needs a little cleanup.

krancour:
Since `Event` is defined in brigadier, a change is required to that library first.

krancour:
Ready for review.

Issue #802, opened 2019-02-07
Pipefail not set on job


radu-matei:
According to the docs, by default, all job tasks are running in a shell with `set -eo pipefail`:

> Tasks can be an arbitrary shell command that is supported by the image you use. Tasks are concatenated together and executed as one shell script. (In Linux, they are executed as /bin/sh commands, with `set -eo pipefail`.)

> `tasks: string[]`: Tasks to be run in the job, in order. Tasks are concatenated together and packaged as a Borne (`/bin/sh`) shell script with `set -eo pipefail`.

However, when the script that eventually runs all tasks is created, `pipefail` never seems to be actually set (https://github.com/Azure/brigade/blob/master/brigade-worker/src/k8s.ts#L972-L977):

```
  let newCmd = "#!" + job.shell + "\n\n";

  // if shells that support the `set` command are selected, let's add some sane defaults
  if (job.shell == "/bin/sh" || job.shell == "/bin/bash") {
    newCmd += "set -e\n\n";
  }
```

This results in wrong exit code for piped tasks (see https://github.com/Azure/brigade/pull/801/checks?check_run_id=60201784).

PR #982, opened 2019-08-28
fix(brigade-worker): ensure set -o pipefail is set if shell is /bin/bash

vdice:
Closes https://github.com/brigadecore/brigade/issues/802

Note that `-o pipefail` is not a valid option for `/bin/sh`, so we only currently add it for `/bin/bash` scripts.

**If applicable**:


vdice:
I'd love to get this into the forthcoming 1.2 release if possible :)  Or perhaps we can cut a 1.3 soon after...

PR #991, opened 2019-10-11
Add brig term command that starts Brigadeterm

### LINKED ISSUE #796: Integration with brigadeterm
We've been discussing over in https://github.com/slok/brigadeterm/issues/56 about integrating [`brigadeterm`](https://github.com/slok/brigadeterm) into a `brig` command.

`brigadeterm` is a really helpful tool that, and in the current context we require users to configure two tools to interact with Brigade from the CLI. The purpose of the integration is to only configure and use one tool.

Ideally, we would have a `brig term` command that starts the `brigadeterm` dashboard.

radu-matei:
closes #796 

cc @slok


radu-matei:
Added two sections in the CLI section about `brig dashboard` and `brig term`.

Issue #928, opened 2019-06-07
Ability to send brigade.json with `brig run`


Mister101:
This is a feature request.

I am prototyping some things that have external dependencies. I saw that I need to have `brigade.json` in my repository in order for the build to pick it up and install them.

Since the dependencies are downloaded on every build anyway (not just when the project is created), would it be possible to add a flag to `brig run` similar to `--file` but for the `brigade.json` file?

radu-matei:
Sure, that is a valid request, and would be really helpful for prototyping, as well as for enabling projects that don’t have a repo associated with them to use dependencies. 

Are you interested in contributing this feature?
I’d be more than happy to help you get started. 

(That being said, I just want to point out that you can also create a custom worker for your project and have the dependencies added in the container image)

adinunzio84:
It will probably be a little while before I am able to contribute this, but if noone else does by then, I can try. In the meantime, I just created a repository. I created this request when I was going through the process and figured it would probably help others prototyping as well.

radu-matei:
As I said, you don't actually _have_ to create a repo - you can create a new worker image, and add the dependencies there and use that worker image for your project, [as described in the documentation](https://docs.brigade.sh/topics/workers/).

adinunzio84:
Yep, I did consider that, thanks. Seemed like the same amount of work either way so I just went with the repo approach.

radu-matei:
ref #774

PR #1017, opened 2019-11-08
feat(*): enable passing brigade.json via brig run

vdice:

Enables passing a local `brigade.json` file to  a `brig run` command, overriding the version found in the project repo, if exists.

Closes https://github.com/brigadecore/brigade/issues/928

**If applicable**:

### DIFF
--- a/brigade-worker/prestart.js
+++ b/brigade-worker/prestart.js
@@ -2,24 +2,28 @@ const process = require("process")
 const fs = require("fs")
 const { execFileSync } = require("child_process")
 
-const mountedDepsFile = "/etc/brigade/brigade.json";
-const vcsDepsFile = "/vcs/brigade.json";
-// Deps file locations in order of precedence.
-const depsFiles = [
+const configFile = "/config.json";
+const mountedConfigFile = "/etc/brigade/config";
+const vcsConfigFile = "/vcs/brigade.json";
+
+// Config file locations in order of precedence.
+const configFiles = [
   // data mounted from event secret (e.g. brig run)
-  mountedDepsFile,
+  mountedConfigFile,
 
   // checked out in repo
-  vcsDepsFile,
+  vcsConfigFile,
 ];
 
-function findDeps() {
-  for (let src of depsFiles) {
+function createConfig() {
+  for (let src of configFiles) {
     if (fs.existsSync(src) && fs.readFileSync(src, "utf8") != "") {
-      return src;
+      // Node's require will complain/fail if the file does not have a .json/.js extension
+      // Here we create the appropriately named file using the contents from src
+      fs.writeFileSync(configFile, fs.readFileSync(src, "utf8"));
+      return;
     }
   }
-  return "";
 }
 
 if (require.main === module)  {
@@ -27,12 +31,15 @@ if (require.main === module)  {
 }
 
 function addDeps() {
-  const depsFile = findDeps();
-  if (!depsFile) {
+  createConfig();
+  if (!fs.existsSync(configFile)) {
     console.log("prestart: no dependencies file found")
     return
   }
-  const deps = require(depsFile).dependencies || {}
+
+  // Parse the config file
+  // Currently, we only look for dependencies
+  const deps = require(configFile).dependencies || {}
 
   const packages = buildPackageList(deps)
   if (packages.length == 0) {
@@ -66,8 +73,10 @@ function addYarn(packages) {
 }
 
 module.exports = {
-  mountedDepsFile,
-  vcsDepsFile,
+  configFile,
+  mountedConfigFile,
+  vcsConfigFile,
+  createConfig,
   addDeps,
   buildPackageList,
   addYarn,
--- a/brigade-worker/prestart.js
+++ b/brigade-worker/prestart.js
@@ -2,15 +2,33 @@ const process = require("process")
 const fs = require("fs")
 const { execFileSync } = require("child_process")
 
-//checked out in repo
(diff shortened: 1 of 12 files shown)


pkg/storage/kube/build.go
```diff
@@ -102,8 +102,9 @@ func (s *store) CreateBuild(build *brigade.Build) error {
 		},
 		Type: secretTypeBuild,
 		Data: map[string][]byte{
-			"script":  build.Script,
-			"payload": build.Payload,
+			"script":       build.Script,
+			"brigade.json": build.Deps,
```
mumoshu: Nit but I think this should better be `deps` rather than `brigade.json` to be consistent with other fields. Other fields seem to be named after their contents, rather than the file names.

radu-matei: I would steer away from `deps`, since we plan to add more things to `brigade.json` in the future (such as environment variables for the worker, for example).

vdice: @mumoshu I agree the discrepancy looks odd.  I discovered if the file doesn't have an extension of `.json`, node's `require` throws an error attempting to parse.  We could work around this for a file with no extension (e.g., `/etc/brigade/deps`) by renaming or reading/rewriting to a local `*.json` file.  WDYT?

```
 $ k logs brigade-worker-01ds62rc3kctgs3sd0agxg123r
/etc/brigade/..2019_11_08_17_59_55.496562507/deps:2
    "dependencies": {
                  ^

SyntaxError: Unexpected token :
    at createScript (vm.js:80:10)
    at Object.runInThisContext (vm.js:139:10)
    at Module._compile (module.js:617:28)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at addDeps (/home/src/prestart.js:35:16)
```

vdice: @radu-matei ah I see.  What should we rename the flag and internal fields as?  Currently `-d`/`--deps-file` and `Deps`...  something connoting a Brigade config file?  `--config`/`Config`?

radu-matei: Yeah, I think `--config` would work.

radu-matei:
This is great, the only comment I'd have here is that we should maybe look at bringing parity for how `brigade.js` and `brigade.json` are stored - i.e. - also add support for a default file, stored in a ConfigMap.

vdice:
> also add support for a default file, stored in a ConfigMap.

@radu-matei good idea!  I'd prefer to have this be a follow-up PR, as the original ticket just mentioned adding functionality to pass in via cli.  WDYT?

radu-matei:
Oh, surely a follow-up PR.

vdice:
PR updated to incorporate review feedback.  PTAL when convenient @mumoshu @radu-matei .  Thanks!

vdice:
Pushed an add'l commit to clean up/add to the `createConfig` units...

pkg/script/delegated_runner.go
```diff
@@ -94,7 +94,8 @@ func (a *Runner) SendBuild(b *brigade.Build) error {
 }
 
 // SendScript converts a script into a Brigade Build object and submits it
-func (a *Runner) SendScript(projectName string, data []byte, event, commitish, ref string, payload []byte, logLevel string) error {
+// TODO: ripe for refactor to reduce amount of params... create/use Script struct?
```
mumoshu: Yeah.. or perhaps go for `SendBuild(projectName string, build *brigade.Build) error {`
because it's almost `brigade.Build` or omitting this function at all?

vdice: @mumoshu I've decided to create a separate ticket for this work, as I was reminded about `rerun` as well.  My thoughts on the refactor are in https://github.com/brigadecore/brigade/issues/1021

vdice:
OK, this should now be ready for final review.  Forgot to add the same `--config` flag to `brig rerun`.  In doing so, I was reminded of a refactor that https://github.com/brigadecore/brigade/issues/1021 should hopefully cover, to consolidate logic between the two commands, etc.

brigade-worker/prestart.js
```diff
@@ -2,19 +2,44 @@ const process = require("process")
 const fs = require("fs")
 const { execFileSync } = require("child_process")
 
-//checked out in repo
-const depsFile = "/vcs/brigade.json"
+const configFile = "/config.json";
```
radu-matei: nit: why the naming difference between the file on disk (or in repo) and the file created by the worker?

brigade-worker/test/prestart.ts
```diff
@@ -2,6 +2,8 @@ import "mocha";
 import { assert } from "chai";
 import * as mock from "mock-require";
 import * as sinon from "sinon";
+import { getDefaultCompilerOptions } from "typescript";
```
radu-matei: Is this import statement used?

brigade-worker/prestart.js
vdice: No particular reason; shall we change it back to `/brigade.json`?

brigade-worker/test/prestart.ts
vdice: Oops, thank you; both this and the line below were unused.  Removed.

brigade-worker/prestart.js
vdice: Updated to be `/brigade.json`

PR #1019, opened 2019-11-11
take default command from image instead of controller

krancour:


technosophos:
Yeah, this looks fine to me. I tried to think of how this might manifest as a breaking change, but I don't think it will. The early implementation was just to make it easier for us to run tests quickly, and it was never really intended that we would keep the container this way always.

PR #1023, opened 2019-11-14
Revert "take default command from image instead of controller"

krancour:
This reverts #1019  because it has been determined to have caused a breaking change in cases where a custom worker might be extended from an older version of our default worker-- in which case its `CMD` remains `yarn run test`. In such cases, that will run unexpectedly since the controller will no longer be overriding it. All such builds will pass with no jobs being executed.

We should be sure to re-apply #1019 in the next major release.

### DIFF
--- a/brigade-controller/cmd/brigade-controller/controller/controller_test.go
+++ b/brigade-controller/cmd/brigade-controller/controller/controller_test.go
@@ -115,8 +115,10 @@ func TestController(t *testing.T) {
 		}
 	}
 
-	if len(c.Command) > 0 {
-		t.Errorf("Expected command to be empty, got %q", c.Command)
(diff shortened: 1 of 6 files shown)


vdice:
> We should be sure to re-apply #1019 in the next major release.

I wonder if we should create a loosely-defined `post-1.0` or other milestone that we can add tickets like these to (for reminding ourselves of some known features/changes we'd like to add in the next major version)... WDYT?

krancour:
> I wonder if we should create a loosely-defined post-1.x/2.0 or other milestone that we can add tickets like these to (for reminding ourselves of some known features/changes we'd like to add in the next major version)... WDYT?

Absolutely. I'm _definitely_ getting around to that one of these days. 😉 Unless anyone beats me to it.

PR #1027, opened 2019-11-18
ref(brigade-worker): remove yarn docs target

vdice:

This removes the `docs` script from brigade-worker's `package.json`, thereby removing a number of dependencies, one of which, `handlebars`, has proved more effort than its worth with frequent security updates needing a separate `resolutions` section (since `typedoc` didn't appear to be updating), frequently failing `yarn audit`/CI.

Since we have the Netlify GH App at our service to generate the docs for all PRs, we're thinking this `docs` script and associated dependencies can be removed.

**If applicable**:


radu-matei:
The only thing to note in this PR is that the `docs` target would create the TypeScript documentation for the worker, not the docs website - however, I still agree that we should remove the target from the worker, and maybe consider adding it to Brigadier.

PR #1031, opened 2019-11-26
add optional title fields to build type

### LINKED ISSUE #976: Proposal: Add human readable title to builds
At present, it's _really_ hard for humans to visually identify which opaquely named Brigade build(s) is/are associated with a given upstream event.

For instance, if I have opened PR#n in my GitHub project, there's no way to quickly identify the build(s) that are related to that.

It seems this could be improved by permitting gateways to "label" builds with a human readable title. `brig` and Kashti could both be improved to display this field if/whenever it is set.

krancour:
Fixes #976 

This adds fields to the build that gateways can _optionally_ populate. Follow-up changes to the gateways and to Kashti can make good use of these fields.

### DIFF
--- a/pkg/storage/kube/store_test.go
+++ b/pkg/storage/kube/store_test.go
@@ -127,8 +127,10 @@ var (
 
 	// stubBuild is a build
 	stubBuild = &brigade.Build{
-		ID:        stubBuildID,
-		ProjectID: stubProjectID,
+		ID:         stubBuildID,
+		ProjectID:  stubProjectID,
+		ShortTitle: "this is a short title",
+		LongTitle:  "this is a long title",
 		Revision: &brigade.Revision{
 			Commit: "abc123",
 			Ref:    "refs/heads/master",
--- a/pkg/brigade/build.go
+++ b/pkg/brigade/build.go
@@ -12,6 +12,18 @@ type Build struct {
 	Type string `json:"type"`
 	// Provider is the name of the service that caused the event (github, vsts, cron, ...)
 	Provider string `json:"provider"`
+	// ShortTitle is an optional field for a short (and not necessarily unique)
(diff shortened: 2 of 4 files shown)


pkg/storage/kube/store_test.go
```diff
@@ -127,8 +127,10 @@ var (
 
 	// stubBuild is a build
 	stubBuild = &brigade.Build{
-		ID:        stubBuildID,
-		ProjectID: stubProjectID,
+		ID:         stubBuildID,
+		ProjectID:  stubProjectID,
+		ShortTitle: "this is a short title",
+		LongTitle:  "this is a long title",
```
vdice: But the short title is longer than the long title! 😜

krancour: > Is the vision for the Kashti dashboard to show the short title for each entry on the build list and then the long title (perhaps in addition to the short) on each individual build page?

That is exactly what I had in mind. Of course, I am open to feedback on that (or in this PR) if that doesn't jive with other maintainers.

krancour:
@vdice not sure what is going on with e2e. Failures look unrelated to the change.

vdice:
> @vdice not sure what is going on with e2e. Failures look unrelated to the change.

Shoot, it appears to be an error that _should_ have been fixed w/ the latest chart releases (for k8s 1.16.x compatibility).  I'll investigate.

vdice:
@krancour fix for e2e failure (in charts repo): https://github.com/brigadecore/charts/pull/58

Sorry for the disruption!  I'll cut a 1.4.2 chart release once the fix is merged.

pkg/storage/kube/store_test.go
radu-matei: This makes sense - the one thing we should make sure in `brig` and Kashti is to fallback to displaying the build ID if the title is not available.

PR #1061, opened 2020-03-05
feat(brig): accept inline payload in brig run

### LINKED ISSUE #774: Feature request: brig run should directly accept the payload as flag
This is useful when running from CI environments and the payload needs information about the current build:

```
$ brig run --json-payload "<my-json-formatted-payload>"
```
For example, I'd like to be able to do the following from a Makefile (most likely wrong quotes there, but along the lines of):

```
brig run --json-payload {"image": "node-demo:$(TAG)"}
```

shubham1172:
Closes #774 - supports adding an inline payload for events in brig run

NA

**If applicable**:

### DIFF
--- a/brig/cmd/brig/commands/run.go
+++ b/brig/cmd/brig/commands/run.go
@@ -13,17 +13,17 @@ import (
 )
 
 var (
-	runFile        string
-	runConfigFile  string
-	runEvent       string
-	runPayload     string
-	runJsonPayload string
-	runCommitish   string
-	runRef         string
-	runLogLevel    string
-	runNoProgress  bool
-	runNoColor     bool
-	runBackground  bool
+	runFile          string
+	runConfigFile    string
+	runEvent         string
+	runPayloadFile   string
+	runInlinePayload string
+	runCommitish     string
+	runRef           string
+	runLogLevel      string
+	runNoProgress    bool
+	runNoColor       bool
+	runBackground    bool
 )
 
 const (
@@ -47,15 +47,15 @@ To send a local JS file to the server, use the '-f' flag:
 While specifying an event is possible, use caution. Many events expect a
 particular payload.
 
-A payload can be either specified as a file or an inline JSON.
+A payload can be either specified inline or in a payload file.
 
-To specify a file, use the '-p' flag:
+To specify a payload file, use the '-p' flag:
 
 	$ brig run -p payload.json
 
-Alternatively to specify inline JSON, use the '-j' flag:
+Alternatively to specify it inline, use the '-i' flag:
 
-	$ brig run -j {"key": "value"}
+	$ brig run -i {"key": "value"}
 
 To run the job in the background, use -b/--background. Note, though, that in this
 case the exit code indicates only whether the event was submitted, not whether
@@ -65,8 +65,8 @@ the worker successfully ran to completion.
 func init() {
 	run.Flags().StringVarP(&runFile, "file", "f", "", "The JavaScript file to execute")
 	run.Flags().StringVarP(&runEvent, "event", "e", "exec", "The name of the event to fire")
-	run.Flags().StringVarP(&runPayload, "payload", "p", "", "The path to a payload file")
-	run.Flags().StringVarP(&runJsonPayload, "json-payload", "j", "", "The payload specified as a JSON")
+	run.Flags().StringVarP(&runPayloadFile, "payload-file", "p", "", "The path to a payload file")
+	run.Flags().StringVarP(&runInlinePayload, "inline-payload", "i", "", "The payload specified inline")
 	run.Flags().StringVar(&runConfigFile, "config", "", "The brigade.json config file")
 	run.Flags().StringVarP(&runCommitish, "commit", "c", "", "A VCS (git) commit")
 	run.Flags().StringVarP(&runRef, "ref", "r", defaultRef, "A VCS (git) version, tag, or branch")
@@ -98,14 +98,15 @@ var run = &cobra.Command{
 			return err
 		}
 
-		if runPayload != "" && runJsonPayload != "" {
-			return errors.New("Both payload and json-payload should not be specified")
+		if runPayloadFile != "" && runInlinePayload != "" {
+			return errors.New("Both payload-file and inline-payload should not be specified")
 		}
 
-		if runJsonPayload != "" {
-			payload := runJsonPayload
+		var payload []byte
+		if runInlinePayload != "" {
+			payload = []byte(runInlinePayload)
 		} else {
-			payload, err := readFileParam(runPayload)
+			payload, err = readFileParam(runPayloadFile)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


brig/cmd/brig/commands/run.go
```diff
@@ -13,16 +13,17 @@ import (
...
-	runRef        string
-	runLogLevel   string
-	runNoProgress bool
-	runNoColor    bool
-	runBackground bool
+	runFile        string
+	runConfigFile  string
+	runEvent       string
+	runPayload     string
+	runJsonPayload string
```
krancour: The older of these two variables, `runPayload`, references a file. The newer one `runJsonPayload`, is meant to directly contain the payload... but, the payload is JSON either way, so the "json" in the new variable `runJsonPayload` doesn't accurately disambiguate how each of these two similarly named variables are used. I would suggest rethinking the naming here and maybe changing the older variable to `runPayloadFile` and the newer one to `runInlinePayload`... or something like that.

brig/cmd/brig/commands/run.go
```diff
@@ -46,6 +47,16 @@ To send a local JS file to the server, use the '-f' flag:
...
 
+A payload can be either specified as a file or an inline JSON.
+
+To specify a file, use the '-p' flag:
+
+	$ brig run -p payload.json
+
+Alternatively to specify inline JSON, use the '-j' flag:
+
+	$ brig run -j {"key": "value"}
```
krancour: See [other comment](https://github.com/brigadecore/brigade/pull/1061/files#r388383249)-- `j` might not be the ideal flag to use here since I am assuming "j" stands for JSON and the payload is JSON no matter what.

brig/cmd/brig/commands/run.go
```diff
@@ -55,6 +66,7 @@ func init() {
 	run.Flags().StringVarP(&runFile, "file", "f", "", "The JavaScript file to execute")
 	run.Flags().StringVarP(&runEvent, "event", "e", "exec", "The name of the event to fire")
 	run.Flags().StringVarP(&runPayload, "payload", "p", "", "The path to a payload file")
+	run.Flags().StringVarP(&runJsonPayload, "json-payload", "j", "", "The payload specified as a JSON")
```
krancour: [See other comment](https://github.com/brigadecore/brigade/pull/1061/files#r388384154)

brig/cmd/brig/commands/run.go
```diff
@@ -86,9 +98,17 @@ var run = &cobra.Command{
...
 
-		payload, err := readFileParam(runPayload)
-		if err != nil {
-			return err
+		if runPayload != "" && runJsonPayload != "" {
+			return errors.New("Both payload and json-payload should not be specified")
+		}
+
+		if runJsonPayload != "" {
+			payload := runJsonPayload
```
krancour: This doesn't look right. Did this compile successfully? It should not have, because it appears that this `payload` variable goes out of scope before it is used.

brig/cmd/brig/commands/run.go
```diff
@@ -86,9 +98,17 @@ var run = &cobra.Command{
...
+		} else {
+			payload, err := readFileParam(runPayload)
```
krancour: Same issue as [here](https://github.com/brigadecore/brigade/pull/1061/files#r388386345)

brig/cmd/brig/commands/run.go
krancour: To clarify, a symbol named `payload` _is_ used down on line 135, but it's actually undefined because the `payload` variable you think you're referencing (i.e. the one here on line 106) is out of scope by then.

shubham1172: Yup, fixed it!

brig/cmd/brig/commands/run.go
```diff
@@ -13,17 +13,17 @@ import (
...
-	runRef         string
-	runLogLevel    string
-	runNoProgress  bool
-	runNoColor     bool
-	runBackground  bool
+	runFile          string
+	runConfigFile    string
+	runEvent         string
+	runPayloadFile   string
+	runInlinePayload string
```
krancour: 👍

krancour:
@shubham1172 thanks for the prompt fixes.

/brig run

shubham1172:
@krancour the PR still seems to be under review.

krancour:
@vdice 

> My only comment is: can we add the same functionality to rerun? run and rerun share many opts/flags and I can see this being a reasonable one to supply to rerun as well. What do you think? Perhaps can be a follow-up... not sure if logic consolidation would result between the two.

We should probably have a conversation about this. I posit this question: If you change the payload on a "rerun" is it _really_ a _re_-run? Or is it a whole new event? I submit it's the latter, _however_ I realize you can _already_ change the payload, `brigade.js` contents, etc. with existing flags on the `rerun` command...

Maybe, for consistency, we do what you've suggested here, but we reconsider how "rerun" works in 2.0?

krancour:
In fact, I am going to open a separate issue about that.

krancour:
I opened #1062

krancour:
> As there isn't consensus on the need/wish for similar functionality in rerun, I definitely don't want to block this great addition in its current form.

@vdice sorry... I was calling out something that I'd (personally) wish to revisit in 2.0. I _do_ think that if `rerun` _currently_ supports all the same options as `run`, it should continue to do so for the life of 1.x.

vdice:
Apologies, I had forgotten about this standing PR.  I rebased off of master and triggered CI.  Assuming all looks good, we'll be set to merge.

shubham1172:
Looks like the merging is blocked because of a pending review comment. I guess @krancour can resolve it now.

brig/cmd/brig/commands/run.go
```diff
@@ -54,7 +65,8 @@ the worker successfully ran to completion.
 func init() {
 	run.Flags().StringVarP(&runFile, "file", "f", "", "The JavaScript file to execute")
 	run.Flags().StringVarP(&runEvent, "event", "e", "exec", "The name of the event to fire")
-	run.Flags().StringVarP(&runPayload, "payload", "p", "", "The path to a payload file")
+	run.Flags().StringVarP(&runPayloadFile, "payload-file", "p", "", "The path to a payload file")
```
krancour, line 68: This appears to be a breaking change. The longer form of the `-p` flag should remain `payload` to avoid this breakage.

krancour: Once this is fixed, the PR will LGTM.

shubham1172: Does it look good?

shubham1172: Updated!

vdice:
We had some blips in CI this morning, but I've verified that all checks passed w/ latest commit (https://github.com/brigadecore/brigade/pull/1061/checks).

It looks like @krancour 's final requested change has been addressed, so merging on his behalf.  Thanks again @shubham1172 !

PR #1068, opened 2020-04-15
docker node upgrade

### LINKED ISSUE #1002: Update Node version for worker
Currently, we're basing our worker image on Node 8, whose deprecation is coming at the end of the year ([source](https://nodejs.org/en/about/releases/))

image

I think it makes sense to migrate to a stable LTS - 10 or 12.

https://github.com/brigadecore/brigade/blob/2218bcf12b4005256713c4caa0ca9e5cb87e89ab/brigade-worker/Dockerfile#L1

veeramarni:
This PR closes #1002

### DIFF
--- a/brigade-worker/Dockerfile
+++ b/brigade-worker/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8-alpine
+FROM node:10-alpine
 
 WORKDIR /home/src
 COPY brigade-worker/ /home/src/
--- a/brigade-worker/Dockerfile
+++ b/brigade-worker/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:10-alpine
+FROM node:12.16.2-alpine3.11
 
 WORKDIR /home/src
 COPY brigade-worker/ /home/src/


krancour:
Is there any reason to upgrade only to Node 10 and not to something even more recent?

My suggestion would be `node:13.12.0-alpine3.11`

veeramarni:
@krancour Looking at https://nodejs.org/en/about/releases/ we can select Node 10 or Node 12 as they being long term support. Let me know if you want me to update the PR to Node 12?
If we node 13 then, it will out of support in few months and we need to update to Node 14 later.

krancour:
@veeramarni 12 then, but specifically the `12.16.2-alpine3.11` tag, please. The `12-alpine` tag is "mutable" and over time will reference newer images. That implicitly can result in different base images for the Brigade worker image from release to release _without_ us explicitly opting to do so. We'd like to avoid that. `12.16.2-alpine3.11`, on the other hand, can reasonable be expected to _always_ point to Node 12.16.2 on Alpine 3.11 and never surprise us with an upgrade we didn't know we were getting.

veeramarni:
@krancour Just pushed with the change which you have suggested. Let me know if there is anything else I can do to get the new worker image created.

krancour:
Thanks @veeramarni!

veeramarni:
@krancour Just checking if this docker image will be released soon or there is some hold on it?

krancour:
@veeramarni I'll meege this, but I'll defer to @vdice on any release plans. We had a minor release just recently, so I fairly certain we don't have another on our radar at the moment.

That said, Brigade can be configured globally or at the project level to use a custom worker image, so you shouldn't end up blocked even if we don't cut a release soon.

PR #1075, opened 2020-04-29
Dependency updates

krancour:
Once again, I've been compelled to upgrade `client-go`. Getting the dependency tree into a workable state was proving difficult with `dep`. I made a judgement call to bite the bullet and move of on to `go mod`. This necessitates an upgrade to a newer version of Go. Once all that was accounted for, wouldn't you know it, the latest `client-go` also had breaking changes to account for.

Even if this passes CI, I'd likely to manually verify the new images that result from these changes prior to merging.


Makefile
```diff
@@ -24,7 +24,7 @@ CLIENT_ARCH ?= $(shell go env GOARCH)
 
 ifneq ($(SKIP_DOCKER),true)
 	PROJECT_ROOT := $(dir $(realpath $(firstword $(MAKEFILE_LIST))))
-	GO_DEV_IMAGE := quay.io/deis/lightweight-docker-go:v0.7.0
+	GO_DEV_IMAGE := krancour/go-tools:v0.1.0
```
vdice, line 27: What does this image provide beyond the official `golang` image?  I'm assuming handy utilities, also possibly with a slimmer footprint?

Although I'd place preference on using an official `golang` image, perhaps if we decide to go with this one, we can add a note and link to the source code somewhere in this repo?  Or, alternatively, we could house a Dockerfile "clone" of `krancour/go-tools` here in this repo, that would serve as the base image for all of the updated Dockerfiles as seen in this commit.  It may start out as a clone, but could potentially diverge as is needed by this project... and then the Dockerfile contents would be even easier to see/update/etc. WDYT?

Makefile
```diff
@@ -79,7 +79,7 @@ IMMUTABLE_DOCKER_TAG := $(VERSION)
 
 .PHONY: dep
```
vdice, line 80: nit: I wonder if we should update the name of this target from `dep` to... `bootstrap`?

Makefile
```diff
@@ -79,7 +79,7 @@ IMMUTABLE_DOCKER_TAG := $(VERSION)
...
 dep:
-	$(GO_DOCKER_CMD) dep ensure -v
+	$(GO_DOCKER_CMD) go mod tidy && go vendor
```
vdice, line 82:
```suggestion
	$(GO_DOCKER_CMD) sh -c "go mod tidy && go mod vendor"
```

Makefile
krancour: `lightweight-docker-go` has been used for a long time (since the Deis days) and has been periodically updated. It includes Go + Go-related tools commonly used (then and now). Things like dep, packr, linters, etc.

It was inconvenient for me to make further updates to that image (for reasons I can explain offline), so I started a new project that is exactly what I have described above, except newer versions of everything.

I am completely unopposed to moving this somewhere under brigadecore and, in fact, wanted to talk to you about that. It is the way it is now simply to unblock me, but I didn't imagine this to be the way it would be left. Perhaps I should have tagged this PR as WIP.

Makefile
krancour: When I originally created that make target, I saw it as coincidental that its name matches the dependency management tool we were using. It's job was (and still is) to manage the dependencies and I had felt `dep` was a succinct and descriptive name. imho, it still is.

fwiw, I am opposed to `bootstrap` because I feel it carries a connotation that it's a target that must be executed before one undertakes any development upon a repo (and perhaps needs to re-execute when switching branches?) Any need for such a thing has been obviated by the fact that we vendor our Go dependencies. So I've long been steering clear of the name `bootstrap` in an attempt to head off any possible confusion.

Makefile
krancour: Yup... _very_ nice catch. Will fix.

Makefile
krancour: Upon further reflection, I would be in favor of changing the name of this target to `resolve-dependencies`. It may be more verbose, but its function is crystal clear and does not (even coincidentally) hint at what the underlying dependency manager is.

Makefile
krancour: Upon further reflection, I would be in favor of changing the name of this target to `resolve-dependencies`. It may be more verbose, but its function is crystal clear and does not (even coincidentally) hint at what the underlying dependency manager is.

Makefile
krancour: Fixed by https://github.com/brigadecore/brigade/pull/1075/commits/4702842c5047de059a9bd7c34c2dd779d5eb4ca3

Makefile
vdice: Good points all.  No strong opinion here; I'll leave it up to you.  I'm fine with leaving it as is, as well.

Makefile
vdice: I'd be fine with a link for now to the Dockerfile that this image is based on in this suite of changes and then a follow-up to perhaps bring in said Dockerfile, either to this repo itself or perhaps a separate repo in the `brigadecore` org.  All in the name of visibility and open access to make updates/changes to this Dockerfile.

vdice:
Testing with Brigade built from this branch looks good on AKS 1.17.3 and minikube 1.18.0 🎉

Makefile
krancour: The metadata in the Docker Hub already links to the Dockerfile in GitHub-- and the builds are automated directly via Docker Hub.

https://hub.docker.com/repository/docker/krancour/go-tools/general

krancour: > then a follow-up to perhaps bring in said Dockerfile

Follow up sounds good, but can we make it its own top-level "utility" project in brigadecore since we're using it at across at least two other projects? Seems it would keep things simpler also if we don't have to build the build image at the start of the pipeline and can use a pre-built image that has all our tools.

krancour:
> Testing with Brigade built from this branch looks good on AKS 1.17.3 and minikube 1.18.0

I've had good results with k8s 1.15.5 and 1.18.2. 1.18.2 is the latest available and 1.15.5 was selected arbitrarily to confirm that the newer go-client remains compatible with (much) older clusters.

Makefile
vdice: Yup ^^, I like it.

krancour:
@vdice I applied the changes we discussed to the Makefile in the latest commit.

scripts/build-brig.sh
```diff
@@ -2,7 +2,7 @@
 
 set -euo pipefail
 
-oses="linux windows darwin"
+oses="darwin"
```
vdice, line 5: Was this change intended?  We still want to build linux and windows brig binaries, for uploading to each GitHub release, correct?

krancour: Whoa. That was a local hack that snuck in. Eagle eye @vdice.

krancour: Fixed it. Force pushed https://github.com/brigadecore/brigade/pull/1075/commits/433a1ce2d5d19c435df4eccf86dc56ad4f721b83 to replace the commit where that hack snuck in. Thanks again.

PR #1095, opened 2020-06-25
Affinities support

vmasutin:
Adding support for brigade worker affnities.
This allows to schedule workers to the specific nodepool, for example:

Cluster X
nodepools:
-default
-workers
-something-else

Brigade components are running on nodes in nodepool default, but brigade workers, are utilising more powerful nodes in workers nodepool

**If applicable**:

### DIFF
--- a/brigade-controller/cmd/brigade-controller/controller/controller.go
+++ b/brigade-controller/cmd/brigade-controller/controller/controller.go
@@ -32,6 +32,8 @@ type Config struct {
 	WorkerRequestsCPU          string
 	WorkerRequestsMemory       string
 	WorkerLimitsCPU            string
+	WorkerNodePoolKey          string
+	WorkerNodePoolValue        string
 	WorkerLimitsMemory         string
 	DefaultBuildStorageClass   string
 	DefaultCacheStorageClass   string
--- a/brigade-controller/cmd/brigade-controller/controller/handler.go
+++ b/brigade-controller/cmd/brigade-controller/controller/handler.go
@@ -164,6 +164,28 @@ func NewWorkerPod(build, project *v1.Secret, config *Config) v1.Pod {
 		RestartPolicy:  v1.RestartPolicyNever,
 	}
 
+	if config.WorkerNodePoolKey != "" && config.WorkerNodePoolValue != "" {
+		log.Printf("Adding affinity, worker pods will be scheduled to nodepool %s", config.WorkerNodePoolValue)
+		affinities := v1.Affinity{
+			NodeAffinity: &v1.NodeAffinity{
+				RequiredDuringSchedulingIgnoredDuringExecution: &v1.NodeSelector{
+					NodeSelectorTerms: []v1.NodeSelectorTerm{
+						{
+							MatchExpressions: []v1.NodeSelectorRequirement{
+								{
+									Key:      config.WorkerNodePoolKey,
+									Operator: v1.NodeSelectorOpIn,
+									Values:   []string{config.WorkerNodePoolValue},
+								},
+							},
+						},
+					},
+				},
+			},
+		}
+		spec.Affinity = &affinities
+	}
+
 	if scriptName := project.Data["defaultScriptName"]; len(scriptName) > 0 {
 		attachConfigMap(&spec, string(scriptName), "/etc/brigade-default-script")
 	}
@@ -314,6 +336,14 @@ func workerEnv(project, build *v1.Secret, config *Config) []v1.EnvVar {
 		} else {
 			envs = append(envs, v1.EnvVar{Name: "BRIGADE_CONFIG", Value: filepath.Join("/vcs", brigadeConfigPath)})
 		}
+	}
+
+	if config.WorkerNodePoolKey != "" {
+		envs = append(envs, v1.EnvVar{Name: "BRIGADE_WORKER_NODEPOOL_KEY", Value: config.WorkerNodePoolKey})
+	}
+
+	if config.WorkerNodePoolValue != "" {
+		envs = append(envs, v1.EnvVar{Name: "BRIGADE_WORKER_NODEPOOL_VALUE", Value: config.WorkerNodePoolValue})
 	}
 
 	return envs
--- a/brigade-controller/cmd/brigade-controller/controller/handler_test.go
+++ b/brigade-controller/cmd/brigade-controller/controller/handler_test.go
@@ -188,3 +188,79 @@ func TestNewWorkerPod_WorkerEnv_ServiceAccount(t *testing.T) {
 		})
 	}
 }
+
+func TestNewWorkerPod_Affinity(t *testing.T) {
+	testcases := []struct {
+		name   string
+		data   map[string][]byte
+		config Config
+	}{
+		{"without affinity",
+			map[string][]byte{},
+			Config{
+				WorkerNodePoolKey:   "",
+				WorkerNodePoolValue: "",
+			},
+		},
+		{"working affinity",
+			map[string][]byte{},
+			Config{
+				WorkerNodePoolKey:   "nodepool-key",
+				WorkerNodePoolValue: "nodepool-value",
+			},
+		},
+		{"broken affinity - no nodepool-key",
+			map[string][]byte{},
+			Config{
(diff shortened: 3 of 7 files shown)


krancour:
@vmasutin thanks for the contribution. I can tell a lot of work went into this.

I want to make sure I have absolute clarity on the value proposition here. At the job level, you can already choose from a subset of nodes using label selectors. This control is conspicuously absent at the worker level and perhaps ought to be implemented. Of course, job-level node selection is _opt-in_ by the author of the `brigade.js` and, were it to be implemented, worker-level node selection would be opt-in by the Brigade user who defined the project (e.g. the user who ran `brig project create`)...

Is the intent behind this feature to allow the _cluster operator_ to place _additional_ constraints on placement that are enforced across the board instead of leaving it entirely up to developers?

vmasutin:
> I want to make sure I have absolute clarity on the value proposition here. At the job level, you can already choose from a subset of nodes using label selectors.

Correct, I have added affinity for worker mostly to have uniform approach with helm charts (e.g. all brigade related pods could be inspected with  `kubectl get pod` and affinity on them will look similar, the only difference would be  the nodepool name)

> Is the intent behind this feature to allow the _cluster operator_ to place _additional_ constraints on placement that are enforced across the board instead of leaving it entirely up to developers?

Yes, also it allows to manage resources in a more dynamic fashion .
Let me describe a use case the way I see it:
- cluster initially created with a single nodepool
- more projects being added  over the time, more builds started
- builds start to interfere with other software running on the cluster
- new dedicated nodepool added 
- brigade installation updated
- resource hungry brigade workers are being scheduled to a new nodepool
- no intervention is required from the developer side, all they see is a reduced build time
- at some point in the future (e.g. provider drops prices on faster vm's), new nodepool is being added, and brigade installations updated to utilise it 

Topologies could be very different, some examples:
- Dev team A is using namespace dev-A, and nodepool _pooldevA_  to run both brigade components and builds
- Dev teams B and C, use their own namespaces dev-B and dev-C, that are being scheduled to nodepool - _default_, but their brigade workers are using nodepool _workers_
- Dev teams D and E, have their own respective namespaces, all pods are being scheduled to nodepool  _poolX_, but their brigade jobs are using shared nodepool _workers_ too

krancour:
@vmasutin first let me say that I'm supportive of this change.

A proposal for what Brigade 2.0 might look like will be made public within the next few days. Among the many things in the proposal is using an agent-per-node log aggregator. To prevent that from becoming a resource drain, one may wish to label a subset of the nodes in their cluster as suitable for hosting Brigade workers (and jobs). i.e. The feature you're adding here is something we definitely want. _Ideally_, we'd like the approach to be one that can be carried over to 2.0, which means we want to be certain that whatever we land on here is something we can live with long term.

So again, I am very supportive of this change. Based on the context above, I ask you to forgive me if I end up seeming picky about implementation details.

One thing I am iffy on here is the user of the term "node pool." You seem to intend it as "a group of nodes designated for a particular purpose" whilst at least two popular k8s distributions (AKS and GKE, others?) use the same terminology to reference a group of VMs with the same configuration that can be scaled up or down as a group. There's a nuanced difference between how you are using the term and how they are using the term and I have a slight worry about the confusion that could result from overloading the term with two meanings.

I'll think about alternative terminology that you could employ here. I'd invite you to suggest alternatives as well.

vmasutin:
> So again, I am very supportive of this change. Based on the context above, I ask you to forgive me if I end up seeming picky about implementation details.
Thats absolutely fine.
> One thing I am iffy on here is the user of the term "node pool." You seem to intend it as "a group of nodes designated for a particular purpose" whilst at least two popular k8s distributions (AKS and GKE, others?) use the same terminology to reference a group of VMs with the same configuration that can be scaled up or down as a group. There's a nuanced difference between how you are using the term and how they are using the term and I have a slight worry about the confusion that could result from overloading the term with two meanings.
From the point of view of kubernetes scheduler - its just nodes that share common set of labels (either custom or added by provider). Azure AKS if I remember correctly, does not support custom node labels (or they wont be preserved during k8s upgrade), so basicaly one is stuck with labels that were added by provider, so you can hardly designate specific nodes from the nodepool to be used for specific workloads.
But its a good catch, if you have any ideas on how to document usage of nodepools in a best way, just add them to the comments and I'll try to produce a draft, e.g. 
- step-by-step guide for specific providers (Azure, GCP) + notes about limitations ?
- disclaimer about usage of terms ?
- something else

krancour:
@vmasutin what do you think about changing `WorkerNodePoolKey` and `WorkerNodePoolValue` to something like `WorkerNodeLabel` wherein the value is of the format `key=value`?

This scheme doesn't overload the meaning of "node pool" (which we've noted is a term used by some existing, managed k8s distros) and (imo) it doesn't lose any clarity. Combining the two fields together may actually even improve the clarity. It says pretty clearly that "worker nodes should be labeled this way."

I'd like to hear @vdice's feedback on this name change as well.

Similar changes would need to be applied to your chart PR, too.

vmasutin:
> @vmasutin what do you think about changing `WorkerNodePoolKey` and `WorkerNodePoolValue` to something like `WorkerNodeLabel` wherein the value is of the format `key=value`?

I am fine with that idea in general, but I think there's more to it:
- normal use case - we split `key=value` on `=`, we are up and running
- somebody supplies `keyvalue`, we fail to split it, what should we do at that point -  send pod  into crashloop ? or just omit `keyvalue` alltogether, spit warning and continue  ?
- nodeaffinity allows to pass multiple values, e.g.
```affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: kubernetes.io/e2e-az-name
            operator: In
            values:
            - e2e-az1
            - e2e-az2
```
 code could be modified to support that case if we are using separated vars for key and value, e.g.
`WorkerNodePoolKey=kubernetes.io/e2e-az-name` and `WorkerNodePoolValue=e2e-az1,e2e-az2`
- pass affinity alltogether as a string ? - most flexible way, but most error-prone and hard to debug

Thats just from the top of my head, different opinions are welcome.

> 
> This scheme doesn't overload the meaning of "node pool" (which we've noted is a term used by some existing, managed k8s distros) and (imo) it doesn't lose any clarity. Combining the two fields together may actually even improve the clarity. It says pretty clearly that "worker nodes should be labeled this way."
> 
> I'd like to hear @vdice's feedback on this name change as well.
> 
> Similar changes would need to be applied to your chart PR, too.

vdice:
No real preference on naming, though lean slightly towards `WorkerNodeLabel`.  I would be fine if we end up with supporting the simpler case of passing `key=value` for `WorkerNodeLabel` or supporting separate configuration of key and value via `WorkerNodeLabelKey` and `WorkerNodeLabelValue` - though, I'd expect if we went the latter route, we'd indeed have support for multiple, comma-delimited label values as @vmasutin points out above.

krancour:
@vmasutin 

> somebody supplies keyvalue, we fail to split it, what should we do at that point

The controller receives this configuration so it can apply it to pods it creates, right? The configuration can be validated at startup. The only thing that could every go into a crashloop would be the controller-- at deployment time; not runtime. Should be ok.

As for how to accommodate multiple labels... are we sure that's warranted at this stage? I'm thinking about how I'd use this feature in practice and I feel I'd be adding _one_ label to eligible nodes. i.e. Is it good enough? And it may not be. I'm only asking the question.

tbqh, this is making me wonder also if the "simplest thing that works" might actually be to simply have a boolean on/off in the configuration to restrict workers to a certain subset of nodes (or not) that are labeled in a particular, _hardcoded_ way. For instance, when enabled, the the controller adds affinity to all worker pods so that they look for specifically `brigade.sh/worker-node=true`.

I know that's _less_ configurable than what you originally proposed, but sometimes I think it's worthwhile to ask whether the additional configuration buys us anything.

krancour:
Amending my previous comment to note that on the flip side of affinity is anti-affinity.

Operators may wish to add some taint to the nodes that should host workers so that those nodes will automatically repel _non-Brigade_ workloads. Thus, the controller should probably _always_ add a toleration to workers for that (hardcoded?) taint and workers should probably _always_ add a toleration to jobs for that same taint.

PR #1101, opened 2020-07-09
Brigade 2.0 proposal for maintainer and community review

### LINKED ISSUE #1100: Draft a Brigade 2.0 proposal for maintainer and community review
Among the maintainers, there have been thoughts for some time now (and many issues opened) dealing with the idea of moving toward a major release-- Brigade 2.0.

This issue calls for a formal proposal to:

* Highlight issues that may prompt architectural pivots
* Recommend a path forward
* Be made public for review, comment, and ratification by:
    * Peer maintainers
    * The broader Brigade community

krancour:
Fixes #1100 

This PR will remain open for _at least_ two weeks and will not be merged until a majority of maintainers are satisfied that feedback from themselves and the community has been sufficiently addressed.

__Edit:__ Anyone wishing to comment, when possible, please make your comments in-line with the relevant proposal markdown so as to provide context for anyone trying to follow the discussion.


brandonkal:
I've looked over this. Some thoughts:
1. Have you considered NATS and Postgres?
2. Automatic namespace creation will make Brigade unusable in certain environments such as https://okteto.com/ and similar services. Brigade should not rely on having the permissions to create and manage namespaces.
3. The term Project can be confusing and you should consider renaming it to something else. Because it is really not a project but an isolated config for launching workers and storing secrets. Perhaps something like Dispatcher or Subscription.
4. Beyond the backend changes, there should be a focus on reducing boilerplate in `brigade.js` files. I've noticed that a lot of non-js projects are marked as JavaScript in GitHub because of the length of the CI config.
5. It would be interesting to abstract the kubernetes operations for the worker.

This is what I am having to do for my Deno worker. The worker's entrypoint launches Deno with higher permissions (can talk on network, read files, read from environment). It then builds the event and launches a subprocess whose only permission is to talk on localhost:8080 to communicate with the parent process through a websocket. Then ops such as requesting a pod run are sent to the privileged parent which speaks kubernetes. Once Deno allows for permissions restrictions in WebWorkers, this could be more lightweight. V8 isolates could be an alternative to launching a pod to run a worker script on every incoming event.

krancour:
@brandonkal that's a lot of feedback. If you don't mind, I'm going to copy/paste individual items from your comment into new, in-line comments on the relevant parts of the proposal markdown so as to provide proper context for anyone trying to follow along. Let's move most discussion to those new comments.

2.0-PROPOSAL/README.md
```diff
@@ -0,0 +1,885 @@
...
+
+To bridge the gap between the synchronous and asynchronous components of this
+subsystem, a message bus capable of ensuring reliable transmission of messages
+with at-least-once delivery guarantees will be required.
+
+After careful consideration of Kafka, ActiveMQ, RabbitMQ, and Redis (using the
+[reliable queue](https://redis.io/commands/rpoplpush#pattern-reliable-queue)
+pattern), Kafka is proposed to be used for this purpose based on several
+factors:
+
```
krancour, line 548: From @brandonkal:

> Have you considered NATS?

With finite bandwidth and a very large number of contenders in this field, summary dismissal of some candidates was necessary in order to allow closer examination of others that were not as easily dismissed. (There was a primary, if you will.) NATS didn't receive closer scrutiny because upon initial consideration, it appeared too k8s-centric. You'll note that there's a theme to this proposal of making Brigade _less_ k8s-centric overall.

As a courtesy, I will give NATS a closer look.

2.0-PROPOSAL/README.md
```diff
@@ -0,0 +1,885 @@
...
+persistence of domain objects such as projects and events. Document-based
+storage is proposed since relatively few relationships exist between domain
+objects and among those relationships that do exist, composition is highly
+favored. i.e. A relational database is unlikely to yield any significant
+advantages and would also introduce object-relational impedance that can
+otherwise be avoided.
+
+Further, after careful consideration of MongoDB, Cassandra, and CouchDB, as well
+as cursory consideration of a few other options, MongoDB is proposed as the
+underlying data store based on several factors:
```
krancour, line 504: From @brandonkal:

> Have you considered Postgres?

With finite bandwidth and a very large number of contenders in this field, summary dismissal of some candidates was necessary in order to allow closer examination of others that were not as easily dismissed. (There was a primary, if you will.)

From the start, it was thought that document-based storage would offer the least impedance with the domain model. Brigade 1.x can be cited as historical proof of this, for storing project and event data in Kubernetes secrets is tantamount to a makeshift document store.

I presume you're not suggesting using Postgres in a relational capacity (please correct me if I am wrong) and are instead referring to its support for storing JSON documents.

Again, having being necessary to dismiss some candidates so that more promising ones might receive closer scrutiny, Posgtres was admittedly dismissed on the basis that document storage seemed an afterthough-- something "tacked on" to their existing product.

You're the second person to suggest a closer look at Postgres (with the other suggestion having come offline from a former colleague). I'm willing to give it further scrutiny.

2.0-PROPOSAL/README.md
```diff
@@ -0,0 +1,885 @@
...
+substrate's (i.e. Kubernetes') native secret storage facilities are deemed
+inadequate, that problem is not Brigade's and can be addressed at a lower level,
+since Kubernetes supports pluggable backends for secret management.
+
+#### Preventing Escalation of Privileges
+
+Isolating each Brigade project to its own namespace has been among Brigade's
+most requested features. To fulfill this, Brigade 2.0 is proposed to (via the
+scheduling subsystem) _automatically create a brand new Kubernetes namespace for
+each new project_.
```
krancour, line 728: From @brandonkal:

> Automatic namespace creation will make Brigade unusable in certain environments such as [https://okteto.com/](https://okteto.com/) and similar services.

Earlier sections of the proposal emphasize that Brigade could be offering a lot of value to people who lack Kubernetes expertise if the de facto requirements for said experience and direct access to a cluster were dropped. To whatever extent is achievable, Kubernetes is meant to fade into the background as an _implementation detail_. You can then view Brigade 2.0 as something much closer to a PaaS than 1.x was.

Viewed in this context, I think Brigade 2.0 and a service like Okteto are mutually exclusive.

__Okteto:__ Small k8s sandboxes (one namespace) aaS for developers who know Kubernetes.

__Brigade 2.0:__ Small event-driven scripting sandboxes (one project) aaS for developers _in general_. k8s is an implementation detail.

> Brigade should not rely on having the permissions to create and manage namespaces.

Perhaps the proposal needs to be more explicit about this, but there was a series of cascading design decisions, each necessitating the next, with this being just about the last in the chain. If Brigade cannot create new namespaces, pretty much the whole proposal falls apart.

Here is the rationale:

1. Today Brigade 1.x is subject to data loss and resource contention issues because of Kubernetes being used as a makeshift document store and message bus.
1. Introducing a real document store and message bus solves those issues.
1. This means clients like the `brig` CLI and gateways can no longer communicate directly with the k8s API server.
1. This means Brigade 2.0 needs the full-featured API that Brigade 1.x lacked.
1. If we're going to have an API, there needs to be a way to authenticate users and authorize actions.
1. But there is no way for random APIs to delegate authn or authz to Kubernetes.
    1. We spent a lot of time trying to figure out how this could be done and ultimately concluded that it can not be done currently. In fact, we discovered along the way that it was proposed feature of Kubernetes at one time, but appears to have long since been abandoned.
    1. Even if this were possible, it undermines our goal of making Brigade 2.0 less k8s-centric and _not_ requiring users to have direct access to a cluster.
1. Authn and authz would have to be achieved through other means.
1. The Brigade 2.0 API doesn't know what permissions (if any) a user has within the k8s cluster.
1. Allowing a user to "bring their own namespace" by telling Brigade what namespace to use for their project would create the possibility of completely hijacking someone else's namespace.
1. The solution is that Brigade 2.0 creates _fresh_ namespaces for each project. It's proposed that it can at least tell you the name of that namespace so that _if_ you have sufficient permissions within the cluster, you can go tweak the namespace, the service accounts therein, etc. to your heart's content.

Hopefully, you can see how _not_ permitting Brigade to create namespaces unwinds this entire stack and would put us back at square one.

An alternative that I will dismiss now is having everything happen in just one namespace as in Brigade 1.x (which incidentally was not meant to work that way-- it's actually a bug). _Many_ users have asked to have projects isolated to their own namespaces. Delivering on that is a high priority.

2.0-PROPOSAL/README.md
```diff
@@ -0,0 +1,885 @@
...
+Incidental types required to address non-functional requirements (such as access
+control) notwithstanding, Brigade 2.0 is proposed to deal with only two "top
+level" domain elements: projects and events.
+
+__Projects__ pair _subscriptions_ to various events with a _template_ for a
+worker that will handle such events. Projects additionally play a role similar
+to that of a Kubernetes namespace in that they define a logical scope for access
+controls. For instance, rather than granting a user permissions to list or view
+_all_ events in Brigade, a user might be granted permissions to list or view
+events _for a particular project_.
```
krancour, line 420: From @brandonkal:

> The term Project can be confusing and you should consider renaming it to something else. Because it is really not a project but an isolated config for launching workers and storing secrets. Perhaps something like Dispatcher or Subscription.

This is a legacy of Brigade 1.x and it is a stated goal to make 2.0 feel familiar to existing Brigade users, despite the many changes.

imho, "project" is also colloquially understood across many different contexts in our industry to denote a unit of isolation or independence. Specifically wrt Brigade 1.x, I've never heard of any prior confusion over the use of this term. (Contrast this with "builds," and the frequent complaint, "but I'm not _building_ anything in my use case.")

krancour:
From @brandonkal:

> Beyond the backend changes, there should be a focus on reducing boilerplate in brigade.js files. I've noticed that a lot of non-js projects are marked as JavaScript in GitHub because of the length of the CI config.

Can you give a specific example of the boilerplate you're talking about?

The biggest offender I am aware of wrt to "boilerplate" is code that is written (or more often copied/pasted) to report job statuses to the upstream systems from which events originated. A great example of this is with workflows triggered by events in a GitHub repository. It's not uncommon to see a "job sandwich" where a job runs to use the GitHub Checks API to report "this job is in-progress", then the job executes, then another job runs to report success or failure.

To reduce this sort of boilerplate, a helper library was introduced quite some time ago that implements these common patterns. tbh, I consider this a half-measure, because I think whether you have a library to help you with it or not, reporting status upstream to the system from which an event originated shouldn't be the responsibility of a worker / your script to begin with. _It should be the responsibility of the gateway that brokered the event._ And, in fact, this is covered in this proposal. Smarter gateways should mean less code in your scripts.

In case I am misinterpreting the sort of boilerplate you are talking about, again, please give me an example.

krancour:
From @brandonkal:

> It would be interesting to abstract the kubernetes operations for the worker.

I'm not sure what is meant by this exactly. Even today in 1.x, there are but a scant few k8s-specific options at the job level, but generally speaking, `brigade.js` scripts are not loaded with Kubernetes-specific details. It should be fairly easy to look at most `brigade.js` scripts and imagine it executing (unchanged) on a different substrate (if something like Mesos or Nomad were supported; which I should state is _not_ a goal).

> This is what I am having to do for my Deno worker. The worker's entrypoint launches Deno with higher permissions (can talk on network, read files, read from environment). It then builds the event and launches a subprocess whose only permission is to talk on localhost:8080 to communicate with the parent process through a websocket. Then ops such as requesting a pod run are sent to the privileged parent which speaks kubernetes. Once Deno allows for permissions restrictions in WebWorkers, this could be more lightweight. V8 isolates could be an alternative to launching a pod to run a worker script on every incoming event.

I'm clearly missing a lot of the context here. Although I don't know why you're doing all of that, I don't see any reason that the design proposed for 2.0 would preclude you from implementing a custom worker that works just as you described.

2.0-PROPOSAL/README.md
brandonkal: Yes, I am referring to it as a JSON document store. Some useful features:
1. RLS. The database itself can determine if a tenant has access to a particular document.
2. It's simple and less Node.js centric
3. You can run [TRIGGERS](https://www.postgresql.org/docs/current/trigger-example.html) (i.e. written with JS or another language)
4. It can optionally act as a [message broker](https://github.com/andywer/pg-listen) which is inline with the simplicity goals here.

2.0-PROPOSAL/README.md
brandonkal: It seems to me this could easily be solved by creating a Project CRD. It wouldn't actually have to store the Project config (just the name). Then if a user has permissions to create that resource in their chosen namespace, brigade would launch pods for that project etc in that namespace. It's odd that the existing design has namespace specified anywhere but metadata.namespace

2.0-PROPOSAL/README.md
brandonkal: The issue is that when I think of project, I consider the code to be part of that project. But in brigade terms a Project is only a subset of that. It's more of something that binds your project to brigade.
Consider a user starting out with brigade:
developer: I wish to start a new project. Let's create a repo to store my code.
developer: I've heard about brigade, let's try hooking up that.
The developer reads the docs, where it suggests storing config in a separate repo. I love that these can be separate but it is a bit confusing when learning brigade the first time.
developer: So I need to create another project just to configure brigade?

And that is the issue. Because many consider the code repository the project's home, it is odd having another project somewhere else.
Example confusion: https://youtu.be/hreCyC7XnOw?t=1848
It's not that bad once you get the hang of it but 2.0 would be an opportunity to give it a more descriptive name.

Even a change to ProjectConfig or Pipeline would help imho.

brandonkal:
> Can you give a specific example of the boilerplate you're talking about?

A few opportunities for reducing boilerplate.

1. events only receive one argument which could be destructured if desired ([e, p])
https://github.com/deislabs/porter/blob/main/build/brigade.js#L19
Note here that all functions are called with the same repeating parameters. If it was one parameter, there would be less typing.
2. Sharing functionality (i.e. GoTest job lib etc) across projects.
3. Abstract the GitHub notification

brandonkal:
> I'm not sure what is meant by this exactly. Even today in 1.x, there are but a scant few k8s-specific options at the job level, but generally speaking, brigade.js scripts are not loaded with Kubernetes-specific details.

This is why I brought this up. The current worker logic does a lot to abstract toe kubernetes-specific details for the `brigade.js` script it executes. The script only has a few operations available. But the current implementation requires that the worker perform all of that work itself (handled by the brigadier file).

The current API is that you can:

1. Submit a job run request to brigadier
2. Receive notification when that job run request completes (the promise resolves or rejects)
3. Submit a log request to brigadier
4. Receive notification when that log request resolves
5. Perform work in response to an event
6. Throw to notify brigadier to end execution.

> Although I don't know why you're doing all of that

My main motivation for using Deno are:
1. It is simpler to develop.
2. One of my motivations for using TypeScript for CI config is the rich auto-complete.
3. I don't want to have to npm install or publish npm packages just to share small bits of functionality across projects. This is especially true when the project is not a node project. Because of that hassle, what I've seen happen across GitHub is a lot of copy-and-paste which makes the files longer.
2. A brigade.ts file is now self-contained much like a insert-ci-platform-yaml-file-here. Its dependencies can be resolved using https imports.
3. Because of (2) this will allow easier sharing of utilities (i.e. goTest function) for my projects
4. It adds some level of security and sanity. My brigade.ts script cannot talk to kubernetes by default, launch processes or read files. No npm installs required for development of the CI config (assuming auto-complete is desired)

My point is that while it is possible to achieve what I want using the current model there are some unfortunate side-effects of the current approach:
1. I had to reimplement much of brigadier.ts for the different runtime. I've made a Deno library with kubernetes types for generating YAML but there is no `@kubernetes/client-deno`. So this increased the barrier to entry for this custom worker.
2. It's highly dependent on being able to speak kubernetes in your custom worker-runtime language.

Consider an alternative worker API where my worker only needs to:
1. Receive JSON event data (which contains project)
2. Notify server which event names it will act on (optional)
3. Submit a well-defined JSON job run request to server with a ulid
4. Respond to a JSON JobRunResult (reject or resolve the promise with that ulid depending on an ok field)
5. Notify server when work is done or on error

That is the approach I ended up building, but it occurred to me that this makes sense in general for 2.0. Given that the proposal wants to abstract more of kubernetes away, I brought this up.

With a secure websocket API as an example, the brigade.ts doesn't even have to run on kubernetes.

As you say, there was a lot of work to abstract the kubernetes details from the `brigade.js` script's perspective. It interacts with the Job abstraction which has a minimal API. brigadier is stubbed out for the npm package but for the worker it is intermingled. Given that this is the desired abstraction, a simpler means of submitting a job run request makes sense in this context.

2.0-PROPOSAL/README.md
krancour: > It seems to me this could easily be solved by creating a Project CRD. Then if a user has permissions to create that resource in their chosen namespace, brigade would launch pods for that project etc in that namespace.

I will re-iterate that there is a goal for this to be useful to people who don't have k8s experience or (direct) cluster access.

2.0-PROPOSAL/README.md
krancour: You're arbitrarily subscribing to a narrow definition of "project" as being equivalent to "repository." Even within our industry, it's not difficult to find many other uses of the word that have nothing to do with a source code repository. 

It seems that some minor fixes to the docs (which would be completely overhauled for 2.0 anyway) would remediate the sort of confusion Joe Beda seems to have in that clip.

krancour:
@brandonkal 

> A few opportunities for reducing boilerplate.
>
> 1. events only receive one argument which could be destructured if desired ([e, p])
https://github.com/deislabs/porter/blob/main/build/brigade.js#L19
Note here that all functions are called with the same repeating parameters. If it was one parameter, there would be less typing.

Perhaps I'm reading you incorrectly, but you're bothered that some functions take two parameters (an event and a project) instead of one??? Also, please realize that there's a lot of flexibility in how you write a `brigade.js` script. You're going to run across some that are well written, succinct, and DRY... and others that aren't.

To be frank, however, this level of implementation detail is _not_ what this proposal is about. We shouldn't be talking about anything as nitty gritty as method signatures right now.

> 2. Sharing functionality (i.e. GoTest job lib etc) across projects.

Brigade has been and will remain completely agnostic of what you choose to do with it. Language specific chunks of functionality for reuse across multiple projects isn't a concern for Brigade itself. Nothing stops you from extracting patterns of interest into your own npm packages. And in fact, some maintainers on our own team have done that.

> 3. Abstract the GitHub notification

Same response as above, however, in this case I've _also_ already noted that if the API enables gateways to have greater insight into the status of events they have brokered, they can take more responsibility for sending status notifications back to upstream systems.

> 4. Defined jobs (much like drone plugin images)

Again-- not a concern for core Brigade. Feel free to create a new project that bundles this sort of stuff in an npm package.

It's also worth noting that Brigade isn't tightly coupled to the default worker (or to JS even) and there's a lot in the proposal about improving the custom worker story further. We really want to focus on "core" Brigade and not the default worker quite as much. Just around the corner, there's room for an ecosystem of alternative workers and you are welcome to contribute your own.

krancour:
@brandonkal 

> This is why I brought this up.....

That's a whole lot of feedback and some (but not all) of it seems hyper-focused on your own use cases and preferences.

I see your point that care is being taken to decouple Brigade from k8s wherever possible, and yet the default worker oddly communicates directly with the Kubernetes API server.

I should correct what appears to be a misconception on your part however. The brigadier library does _not_ interact with k8s. The brigadier library is actually just a bunch of mocks and no-ops so that you can develop your scripts with the benefits of code completion and such. The default worker swaps in an alternative implementation of that package and _that_ talks to k8s.

In the end, you can view it as "here's a script that defines my workflow and this worker image is capable of executing that workflow within k8s." I'll again be clear that we're not even entertaining the possibility of an alternative worker/job execution substrate, but if we hypothetically were, a different image that swaps in a _different_ package in place of brigadier can accommodate that different substrate. You wouldn't need to make any changes to your script.

Yes... that would mean two separate worker images for "brigade.js + k8s" and another for "brigade.js + nomad" (for example). I think your point is that we could get away with just _one_ image if that one image always communicates with the brigade API server and we let the Brigade API server handle all of the interactions with the substrate. _That sounds great if we had any genuine interest in supporting an alternative substrate, but we don't._

Last, I'll point out that while we'd like Kubernetes expertise and direct cluster access to not be de facto prerequisites for success with Brigade 2.0, I really _don't_ want to cut off the avenues by which someone who is quite experienced with k8s still _can_ do some more complex things if they must. You presented it as if brigadier only lets you do a few simple things like create jobs, wait on jobs, get logs, etc., but you're forgetting that the full power of JS is at your fingertips and if you _want_ to do arbitrarily complex things in your worker-- including complex things with k8s-- you need only import the relevant libraries.

With that in mind, if we wanted the Brigade API server to offer some abstraction over k8s, it wouldn't be merely a few endpoints for the likes of "create a job". We'd have a _lot_ of work cut out for us and what would it buy us given that we're not interested in plugging anything other than k8s into it?

Your custom worker sounds interesting. I'll again point out we're trying to keep the core decoupled from the worker implementaton(s), so if you can create a _better_ worker, please do.

brandonkal:
I brought up abstracting the interaction with kubernetes to the API as an option to make developing custom workers less complex. It's not just about supporting nomad etc.

There is a file in the default worker called brigadier. This along with the k8s file handles the kubernetes implementations of the Job runs. I specifically mentioned that the npm package mocks this out outside of the worker. Second to last sentence if it was missed. So I don't see any misconception on my part.

If authors of custom workers must handle all the kubernetes work themselves, this proposal seems like a halfway measure. Now authors must speak to the kubernetes AND the brigade API servers.

krancour:
@brandonkal 

> If authors of custom workers must handle all the kubernetes work themselves, this proposal seems like a halfway measure. Now authors must speak to the kubernetes AND the brigade API servers.

This simply isn't true. Note the proposal says very little about changes to the default worker and that's actually because not many are needed. Workers would _not_ talk to the Brigade API. If you are under the impression that the task of creating a custom worker becomes more onerous with this proposal, that isn't the case.

krancour:
@brandonkal I think I'm starting to understand a nuance of your suggestion about the Brigade API "abstracting Kubernetes" that either wasn't clearly communicated or was and I somehow missed it. Regardless of which of those it may have been, see if I have this right...

You're _not_ suggesting that we make the Brigade API capable of abstracting the full range of things that Kubernetes can do. You're suggesting making it capable of doing specific things. It's _not_ an API for creating pods, services, deployments, ingresses, etc., etc., etc.-- with a zillion options. It's an API for _creating a job_. So that, perhaps, Brigade can ensure all correct labels and such are added to those pods. This lowers the barrier for entry to building a custom worker _not_ because interacting with Kubernetes is onerous, but because it relieves some of the burden of getting all the _Brigade-specific_ details right.

_If this is what you meant, I clearly see the value in it._

I see numerous technical hurdles to this, but I am at least willing to look at what would be involved.

krancour:
@brandonkal I need to figure out where to fit it neatly into the proposal, but I'll amend it to add something about a "jobs API."

brandonkal:
@krancour 
Yes, when I mentioned abstracting kubernetes, I meant only as it relates to the operations the default worker now performs (and custom workers must replicate). While from the user's perspective the operations are abstracted into Job.run() and Job.logs() operations (because the implementation is mocked out), in reality it is actually intermingled in the true worker runtime and custom worker developers have to reimplement this work.

## Background

Before considering brigade, I used drone CI. It requires its own YAML format but it has the concept of a conversion extension. On a push event it will check out the source and post the CI config file to your conversion server, expecting a YAML response that it understands.
This meant all I had to do was render Deno TypeScript to Drone YAML and I had nice auto-completion for writing pipelines.

But if I was going to use TypeScript, brigade was much more flexible:
1. It is designed to respond to any event, not just repository events.
2. You can submit more jobs partway-though.
3. You can use the result of one job/step to determine what jobs to launch afterwords (i.e. conditional logic).

With brigade, instead of rendering the whole pipeline at once, you submit single jobs in order or parallel and you can act on the results.

That's the part that it makes sense to abstract. To make that operation as simple as possible.
Submit a JobRunRequest and brigade handles the details of:
1. validating request
2. creating and waiting for the PVC, secret, and pod to schedule
3. watching and waiting for pod to complete its work
4. and later... notify requestor of success/failure

## How

The simplest route would be a go/rust binary that acts as a watchdog and handles requests (i.e. the real brigadier library would call it). Users would then extend the Docker image to customize the logic. Not unlike how `kubectl watch` as a sidecar makes it easier for an application pod to interact with kubernetes. That would probably be compatible with 1.x. even.

Going beyond that, it could be part of the controller's API, though that would be more work and require implementing auth for the workers outside of the service-account auth mechanism. That would enable running the worker logic outside of kubernetes if desired (i.e. for local debugging or for a serverless function).

2.0-PROPOSAL/README.md
```diff
@@ -483,6 +494,11 @@ N.B.: In the case of projects, JSON schema can be reused by IDEs to validate
 project definitions as they are edited. The author proposes this UX to be
 superior to the current `brig project create` process.
 
+As with Kubernetes, Brigade 2.0's API surface is likely to be logically
+decomposed into several smaller APIs foucsed on specific  areas of concern to
```
bridgetkromhout, line 498: > foucsed on specific  areas of concern
focused on specific areas of concern

2.0-PROPOSAL/README.md
```diff
@@ -483,6 +494,11 @@ N.B.: In the case of projects, JSON schema can be reused by IDEs to validate
...
+facilitate ease of use and different authentication models to suit the needs of
+different clients. (Refer to the [authentication section](#authentication)).
```
bridgetkromhout, line 500: Lots of "to" and "of" which makes this slightly convoluted to follow. Perhaps break into multiple sentences?

krancour:
@brandonkal 

We're mostly on the same page. Please refer to commit https://github.com/brigadecore/brigade/pull/1101/commits/e73c6233718d4b3f7d9d74aa5c8473897504e33c

> The simplest route would be a go/rust binary that acts as a watchdog...

This level of detail is not necessary right now.

2.0-PROPOSAL/README.md
krancour: Fixed in 7dade9ed162f73f3659ad20c34844893a616da8f

2.0-PROPOSAL/README.md
krancour: Also fixed in 7dade9ed162f73f3659ad20c34844893a616da8f

brandonkal:
> This level of detail is not necessary right now.

That's fine. I'm just offering ideas. But it seems a bit arbitrary if other details such as "non-expiring but revokable bearer tokens" are included. I would say it is worth mentioning in the proposal that a "jobs api" could be either a process within the worker container and doesn't have to be part of a long-running controller.

2.0-PROPOSAL/README.md
```diff
@@ -0,0 +1,907 @@
...
+deployment / re-deployment / upgrade.
+
+N.B.: When any _new_ user authenticates with OpenID Connect, a new Brigade
+account is created automatically and federated with the user's identity from the
+trusted provider. Such a user initially has no roles assigned (see
+[authorization](#authorization) section). Roles may be granted by a user manager
+or project admin. It is proposed that users, once registered _cannot_ be deleted
+since nothing prevents a deleted user from re-registering. The benefits of
+_keeping_ but disabling the user account and thereby maintaining the relational
+integrity of the underlying data store seem to outweigh the finality of
```
thomastaylor312, line 654: Is there going to be any problems with data retention laws here? People might want to be "fully removed"

2.0-PROPOSAL/README.md
```diff
@@ -0,0 +1,907 @@
...
+shared secret that is mutually agreed upon at the time of deployment /
+re-deployment / upgrade.
+
+Lastly, every worker (another non-human user) should be uniquely authorized to
+create jobs associated with itself. (i.e. No one _but_ a given worker should
+ever create jobs _for_ that worker.) Given this, every worker will be granted a
+bearer token to facilitate sessionless authenthication to the jobs API only.
+
+To prevent session hijacking by means of root password theft or bearer token
+theft, Brigade API servers in remote clusters should secure API traffic using
```
thomastaylor312, line 682: For kicking the tires, will there be a process for provisioning self-signed certs?

2.0-PROPOSAL/README.md
```diff
@@ -0,0 +1,907 @@
...
+feedback from and ratification by the broader Brigade community.
+
+## TL;DR
+
+Brigade 2.0 is proposed to introduce a less Kubernetes-centric experience,
+wherein users having little or no Kubernetes experience, or even those lacking
+direct access to a cluster, can quickly become productive. Breaking changes are
+on the docket, but Brigade 2.0 should feel familiar to anyone with previous
+Brigade experience. In short, the author proposes Brigade's nuanced transition
+from "event-driven scripting for Kubernetes" to "event driven scripting (for
```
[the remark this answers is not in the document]
krancour: Until now, k8s has been at the very core of Brigade. Apart from being used as a makeshift doc store and makeshift message bus, it's also impossible to use Brigade without credentials for the k8s cluster and presumably no one gives you that sort of access unless you also have some k8s expertise. Some issues with the UX also prompt users to "drop down to k8s" for certain things. i.e. k8s is "in your face."

The main thrust of this proposal is about k8s fading into the background as an implementation detail-- albeit an important one.

Given that this appears in the tl;dr section, I think I might be ok with this statement being slightly less than crystal clear. Those who feel any confusion should read on, because all of the above is spelled out more clearly point by point.

2.0-PROPOSAL/README.md
```diff
@@ -0,0 +1,907 @@
...
+While it is easy to imagine how the potential for data loss might preclude
+Brigade's use by enterprises with strict data retention policies, all users
+would do well to consider the ramifications of the above before using Brigade to
+implement any critical process.
+
+Refer to:
+
+* [Issue #960](https://github.com/brigadecore/brigade/issues/960)
+* [Issue #1034](https://github.com/brigadecore/brigade/issues/1034)
+* [Issue #1035](brigade/issues/1035)
```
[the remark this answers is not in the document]
krancour: Yep... that looks wrong. Will fix.

2.0-PROPOSAL/README.md
```diff
@@ -0,0 +1,907 @@
...
+purely philosophical course corrections.
+
+First, Brigade currently refers to individual executions of user-defined
+workflows as "builds." This is a misnomer which the maintainers collectively
+regret, as it strongly suggests that Brigade is a CI platform, which it is _not_
+even if it _is_ often used to implement CI workflows. Equally regrettable is
+that the term "build" has too often been used synonymously with "event." The
+author proposes to strike "build" from Brigade's lexicon entirely and speak more
+generically, only in terms of events, each of which has a worker that _handles_
+it.
```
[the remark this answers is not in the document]
krancour: It's only a half-pivot, and only in terminology. "Build" is used throughtout the code and "build" and "event" are casually used interchangeably throughout the existing docs. We want to just stop saying "build" and only say "event" because that word is loaded with fewer assumptions.

2.0-PROPOSAL/README.md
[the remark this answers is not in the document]
krancour: Fixed in 09197a073e0e90e7f710a3c5a1de3bef83a35fe8

2.0-PROPOSAL/README.md
```diff
@@ -0,0 +1,907 @@
...
+
+1. Neither Kubernetes expertise nor cluster credentials must be a prerequisite
+   for success.
+    1. Kubernetes is neither to be removed nor 100% abstracted. Realistic goals
+       are to minimize Kubernetes' use as anything other than a substrate for
+       workload execution and to isolate Brigade's touchpoints with Kubernetes
+       to specific components.
+    1. Some degree of freedom to "drop down" to Kubernetes may be retained to
+       enable implementation of advanced / fringe use cases, but the need to
+       utilize this should be _exceptional._
```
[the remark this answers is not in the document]
krancour: I was avoiding speculating on the reasons some might choose to do this, but now that you've raised this, I think there _is_ a specific prototype for such scenarios. It's "I want to do k8s-specific things _with_ Brigade. I know what I'm doing and I have cluster access."

The "escape hatch" for these scenarios is that information is available about what namespace the project uses and what service accounts are used for the workers and jobs. So if you have appropriate permissions on the cluster, you can "drop down" to give your project's workers or jobs more permissions or deploy additional things (like some kind of test fixture maybe) in your namespace.

I expect to cover this in an "advanced usage" section of docs, but for now, I think your advice about putting some kind of constraint on the end of this sentence seems about right.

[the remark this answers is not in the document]
krancour: Going with this:

> "Dropping down" to Kubernetes should be required only when implementing
      advanced use cases that call for modifications to a project's underlying
      Kubernetes namespace or service accounts.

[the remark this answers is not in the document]
thomastaylor312: That looks good to me!

2.0-PROPOSAL/README.md
krancour: Admittedly, I don't know. I've been looking at "data retention" from the perspective of an enterprise that is required to _keep_ records and not from the opposite perspective where the right to disappear into the ether might be mandated by law. My understanding is that such laws are for consumer protection, however, and I don't see Brigade as a consumer-facing product.

It remains a good question. If we learn later that there is any kind of issue with this, I am pretty confident we can fix it.

2.0-PROPOSAL/README.md
krancour: It's easy enough to make a Helm chart create a self-signed cert and poke it into a secret, so yes. This would be a more sensible default than skipping SSL by default.

2.0-PROPOSAL/README.md
krancour: @brandonkal fyi, I've been investigating NATS. I'm more optimistic about it now than I was initially. From a development perspective, the fair scheduling problem can possibly be solved more easily than with Kafka, although I haven't written any code to prove this _yet_.

From an ops perspective, I am a little more reluctant because I haven't yet found a Helm chart to set up NATS _with streaming support_ nor have I found an option to provision a _fully managed_ NATS in any (or preferably several) of the major public clouds. Ideally, we would want to be able to say "we're not experts on running NATS in production; here is some guidance, but here also are options for running it in production with zero fuss."

Stay tuned.

krancour: Helm chart for setting up NATS streaming: https://github.com/nats-io/k8s#helm-charts-for-nats

An option to deploy NATS as a managed service to the cloud of your choice: https://synadia.com/ngs/pricing

I plan to try these both today.

2.0-PROPOSAL/README.md
krancour: Objections are noted. I've spoken with two other maintainers offline and we agree that unless a majority of maintainers (who _all_ are reviewing this proposal and community feedback carefully) voice opposition to the term "project," the term will stand.

2.0-PROPOSAL/README.md
krancour: @brandonkal

Update on Kafka v NATS:

First, thank you for encouraging me to give NATS a closer look. My initial assessment that it was "too k8s centric" was in error and after spending some time getting to know NATS and writing some sample code, I found it an absolute joy to work with. It handled our use case better than anything else I looked at. I'll be frank in saying that _from a developer's perspective_, I now have a strong bias toward NATS.

However, from an operations perspective, I cannot endorse it. NATS can easily account for the "batteries included" story since I found Helm charts that would have worked just fine with a few tweaks, but should anyone decide that they do not care for the burden of hosting / operating NATS at scale in production and wish to provision a managed service instead, the situation is bleak, at present. Synadia's offering looks compelling, but it's the _only_ option for managed NATS that I could find and on top of that, it doesn't support NATS streaming (aka "STAN") which is a feature we absolutely need in order to implement our use case.

I look forward to this situation improving in the future and NATS becoming a more tenable option for use with future projects.

Now, for a surprising turn of events...

Having access, as I do, to folks from Azure's messaging team (responsible for products like Azure Service Bus and Azure Event Hubs; i.e. folks who live and breath messaging) I engaged them for some input. Their advice was that I may have been too quick to dismiss AMQP 1.0 and they offered guidance on using it to satisfy our use case. In the end, programmatically, it ended up being a better fit than either Kafka or NATS.

This left the question of how easily the "batteries included" and "managed service" narratives could be fulfilled on the ops side of the equation. ActiveMQ is probably the most popular and battle-tested OSS AMQP 1.0 implementation. There is no production quality Helm chart for this (though there are some interesting projects that are WIP). So, figuring out how to get Artemis (the next gen ActiveMQ) cranking inside a k8s cluster was non-trivial and cost me several days of effort, but in the end I am confident that it can be done _at least_ well enough to satsify the "batteries included" narrative. For those wishing to use a managed service in production, the story is much simpler, as AMQP is a Standard with a capital S (it's an OASIS standard) and I was easily able to locate numerous options for hosted/managed ActiveMQ and other AMQP-compatible services.

So, recapping this surprising twist-- the suggestion to take a closer look at NATS will be resulting in an amendment to the proposal to propose AMQP instead with ActiveMQ Artemis as the "included batteries."

krancour:
@mumoshu _great feedback and thank you for the LGTM._

> I wanted to emphasize that Brigade 2 shouldn't be "tied" to specific message queue/database implementations, if possible.
>
> From an operator's perspective, I'd imagine I would love choosing the best message queue/database that satisfies both Brigade 2 requirements AND I and my colleagues has more expertise with.
>
> For example, I may potentially choose AWS SQS and DynamoDB for companies that mainly uses AWS, if I could implement the alternative mq/db implementations for Brig 2.
>
> So, would you mind if we explicitly stated in this proposal that Brigade's controller/apiserver code MUST NOT have tight-coupling to specific queue/database APIs?
>
> In Brigade v1, we have the Store interface for this purpose. Something like in Brigade v2 would be great.

Yes. This was the plan, but it wasn't explicitly stated in the proposal and probably should be. The question of which message bus and database we will use has really been a question of _defaults_ that we will develop ourselves and support. I will amend the proposal to make this more clear.

(@brandonkal the above statement may be of interest to you.)

Also, it was my intention to be pluggable at _two_ levels-- the implementation level (e.g. at deployment time, choose a "connector" for you DB or message bus of choice) and _at the protocol level_. i.e. Keep the defaults and select a compatible service from your vendor of choice. The proposed defaults reflect this because you can provision something compatible from multiple vendors.

> I DO believe having the "default" mq/db implementations for Kafka and MongoDB in Brigade v2 is a great idea.

As an fyi, I'm amending the proposal today to suggest AMQP 1.0 as a default instead of Kafka. See See [this comment](https://github.com/brigadecore/brigade/pull/1101#discussion_r457395634) for the rationale. __Please do let me know if this amendment changes your LGTM.__

2.0-PROPOSAL/README.md
krancour: @brandonkal after reviewing Postgres' JSON support, I am still of the opinion that MongoDB is a better fit. I only got as far as writing some relatively simple queries because I encountered a problematic detail early on-- Postgres isn't schema-less. i.e. You still need to create tables to hold your JSON blobs. That's fine, except that in our case that would introduce a new complication in terms of how we would automate database migrations at the time of Helm chart install/upgrade-- which is a thing that last I attempted it was doable, but not without its own complications.

I'd prefer to avoid unnecessary complications at this juncture, so my initial recommendation of MongoDB stands.

However, I _am_ updating the proposal to be much more explicit about loose coupling to the underlying data store to preserve the possibility of support for alternative data stores being contributed by the community in the future.

mumoshu:
@krancour Thanks for your effort!

> As an fyi, I'm amending the proposal today to suggest AMQP 1.0 as a default instead of Kafka. See this comment for the rationale. Please do let me know if this amendment changes your LGTM.

Honestly saying, I have no experience with AMQP 1.0 and its implementations.

But I do see famous clouds like Azure and AWS have their own managed MQ services that support it. I have also read client code for that like seen in the example of [Azure/go-amqp](https://github.com/Azure/go-amqp#example-usage), and it looks good enough. So I really don't have any specific reason to say no to AMQP 1.0 at this point.

With that being said, I'm fine as long as we use only basic operations supported by it. For example, if we are to only use basic APIs like `fun Send(Message)` and `func Receive() Message`, it would be great as we can easily add/switch other MQ impls.

krancour:
@mumoshu 

> With that being said, I'm fine as long as we use only basic operations supported by it. For example, if we are to only use basic APIs like fun Send(Message) and func Receive() Message, it would be great as we can easily add/switch other MQ impls.

Funny you bring this up. The devil's always in the details. Even though there are many services that support this protocol, they have varying capabilities. (So much for protocols, eh?) By way of example, "fair scheduling" is likely a big concern for us. You don't want one event for project X to be blocked for a long time by a million events for project Y. How you deal with that in ActiveMQ v Azure Service Bus (for instance) are different. ActiveMQ can be configured to dynamically create new queues for each project. Azure Service Bus cannot do that, but it _does_ have an option where you can use a single queue and use session identifiers to logically group messages and not permit delivery of messages in one group to be blocked by messages in another.

Bottom line is we _will_ only use the basic functionality, but even then, there will need to be some pluggable "strategies" for dealing with the idiosyncrasies (differing capabilities) of whatever services you might plug in. This, however, shouldn't be too hard to account for.

mumoshu:
@krancour Thanks! So...

> ActiveMQ can be configured to dynamically create new queues for each project. Azure Service Bus cannot do that, but it does have an option where you can use a single queue and use session identifiers to logically group messages and not have delivery of messages in one group blocked by messages in another.

My assumption was that those settings would be configurable (per session | on queue creation), which means it won't be exposed via/opaque from Brigade-internal Send/Receive APIs, at least. (Am I correct?

We may still end up having implementation-specific mq session "factories", to make those settings opaque to the basic APIs. IMHO that's unavoidable. If that's what you mean by "pluggable strategies", that would be good.

krancour:
@mumoshu 

> We may still end up having implementation-specific mq session "factories", to make those settings opaque to the basic APIs. IMHO that's unavoidable. If that's what you mean by "pluggable strategies", that would be good.

Essentially correct.

technosophos:
I just re-reviewed with the AMPQ 1 changes, and am good with the plan. Is there anything you want to add @lukepatrick?

krancour:
This meets its self-defined acceptance criteria. I'm merging and we can consider this the official (tho not immutable) plan. Further feedback is welcome in the form of issues/PRs per usual.

PR #1113, opened 2020-08-05
Update installation command to Helm 3

radu-matei:
This commit updates the instructions in the installation guide to use
the correct Helm 3 command for deploying Brigade - with the name passed
as the first argument, rather than with the `--name` flag.

Signed-off-by: Radu M <root@radu.sh>


radu-matei:
Ooh, I missed that there were more, good catch.
Updating now.

krancour:
> Updating now.

@radu-matei did you make some new commits that you forgot to push?

PR #1118, opened 2020-09-17
First official bits of Brigade 2: An SDK for Go

krancour:
This PR contains the first official bits of Brigade 2-- namely an SDK for Go.

Why start here?

* The SDK gives a succinct overview of Brigade 2's domain model and interfaces without bogging reviewers down with implementation details.

* Making this available sooner rather than later unblocks any developers who wish to work on event gateways or tooling in parallel.

Of course, there's no official Brigade 2 API server for this SDK to interact with yet, but in the interim, it _is_ compatible with the Brigade 2 prototype at [krancour/brignext](https://github.com/krancour/brignext).

This PR is decomposed into two commits. The first removes any v1 things I don't anticipate us needing. The second actually adds new code. Since this is a large PR, it may benefit reviewers to focus on the latter _or_ since this PR is tantamount to a fresh start, it could make sense to review the result rather than the diffs.

Please note this PR targets a `v2` branch and _not_ the `master` branch.


sdk/v2/core/jobs_test.go
```diff
@@ -0,0 +1,240 @@
...
+	requireAPIVersionAndType(t, JobSpec{}, "JobSpec")
+}
+
+func TestJobStatusMarshalJSON(t *testing.T) {
+	requireAPIVersionAndType(t, JobStatus{}, "JobStatus")
+}
+
+func TestJobsClientCreate(t *testing.T) {
+	const testEventID = "12345"
+	const testJobName = "Italian"
```
vdice: 🇮🇹 🚗 💨

sdk/v2/core/projects.go
```diff
@@ -0,0 +1,327 @@
...
+	"github.com/brigadecore/brigade/sdk/v2/meta"
+)
+
+// ProjectsSelector represents useful filter criteria when selecting multiple
+// Projects for API group operations like list. It currently has no fields, but
+// exists to preserve the possibility of future expansion without having to
+// change client function signatures.
+type ProjectsSelector struct{}
+
+// ProjectList is an ordered and pageable list of ProjectS.
```
vdice:
```suggestion
// ProjectList is an ordered and pageable list of Projects.
```

sdk/v2/core/projects.go
```diff
@@ -0,0 +1,327 @@
...
+			Alias: (Alias)(p),
+		},
+	)
+}
+
+// ProjectSpec is the technical component of a Project. It pairs
+// EventSubscriptions with a prototypical WorkerSpec that is used as a template
+// for creating new Workers.
+type ProjectSpec struct {
+	// EventSubscription defines a set of trigger conditions under which a new
```
vdice:
```suggestion
	// EventSubscriptions defines a set of trigger conditions under which a new
```

sdk/v2/core/projects.go
```diff
@@ -0,0 +1,327 @@
...
+}
+
+// EventSubscription defines a set of Events of interest. ProjectSpecs utilize
+// these in defining the Events that should trigger the execution of a new
+// Worker. An Event matches a subscription if it meets ALL of the specified
+// criteria.
+type EventSubscription struct {
+	// Source specifies the origin of an Event (e.g. a gateway).
+	Source string `json:"source,omitempty"`
+	// Types enumerates specific Events of interest from the specified source.
```
vdice: For instance, should we mention the behavior if `Types` is empty?  Are any/all event types (meeting the other criteria) subscribed to in this case?

sdk/v2/core/projects.go
```diff
@@ -0,0 +1,327 @@
...
+	// WorkerTemplate is a prototypical WorkerSpec.
+	WorkerTemplate WorkerSpec `json:"workerTemplate"`
+}
+
+// EventSubscription defines a set of Events of interest. ProjectSpecs utilize
+// these in defining the Events that should trigger the execution of a new
+// Worker. An Event matches a subscription if it meets ALL of the specified
+// criteria.
+type EventSubscription struct {
+	// Source specifies the origin of an Event (e.g. a gateway).
```
vdice: I'm wondering if it is worth spelling out the behavior of each of these filters when they are left empty/uninitialized.

sdk/v2/core/secrets.go
```diff
@@ -0,0 +1,170 @@
...
+			},
+			Alias: (Alias)(s),
+		},
+	)
+}
+
+// Secret represents Project-level sensitive information.
+type Secret struct {
+	// Key is a key by which the secret can referred.
+	Key string `json:"key,omitempty"`
```
vdice: Wondering if both the `Key` and `Value` fields should be deemed required via removing the `omitempty`s?

sdk/v2/core/workers.go
```diff
@@ -0,0 +1,381 @@
...
+	// the Event.
+	Jobs map[string]Job `json:"jobs,omitempty"`
+}
+
+// WorkerSpec is the technical blueprint for a Worker.
+type WorkerSpec struct {
+	// Container specifies the details of an OCI container that forms the
+	// cornerstone of the Worker.
+	Container    *ContainerSpec `json:"container,omitempty"`
+	UseWorkspace bool           `json:"useWorkspace"`
```
vdice: Missing godocs for `UseWorkspace`

sdk/v2/internal/restmachinery/client.go
```diff
@@ -0,0 +1,185 @@
...
+	"net/http"
+	"strconv"
+
+	"github.com/brigadecore/brigade/sdk/v2/meta"
+	"github.com/pkg/errors"
+)
+
+// BaseClient provides "API machinery" used by all the specialized API clients.
+// Its various functions remove the tedium from common API-related operations
+// like managing authentication headers, encoding request bodies, interpretting
```
vdice:
```suggestion
// like managing authentication headers, encoding request bodies, interpreting
```

sdk/v2/internal/restmachinery/client_test.go
```diff
@@ -0,0 +1,290 @@
...
+					},
+				},
+			}
+			resp, err := client.SubmitRequest(context.Background(), testCase.req)
+			testCase.assertions(t, resp, err)
+		})
+	}
+}
+
+// 	testProject := Project{
```
vdice: Are these commented-out lines still intended to be included?

sdk/v2/internal/restmachinery/outbound_request.go
```diff
@@ -0,0 +1,24 @@
...
+
+// OutboundRequest models of an outbound API call.
+type OutboundRequest struct {
+	// Method specifies the HTTP method to be used.
+	Method string
+	// Path specifies a path (relative to the root of the API) to be used.
+	Path string
+	// QueryParams optionally specifies any URL query parameters to be used.
+	QueryParams map[string]string
+	// authHeaders optionally specifies any authentication headers to be used.
```
vdice: Starting w/ this field and down, the godoc appears to need updating (field needs capitalizing)

sdk/v2/meta/errors.go
```diff
@@ -0,0 +1,97 @@
+package meta
+
+import "fmt"
+
+// ErrAuthentication represents an error asserting an principal's identity.
```
vdice:
```suggestion
// ErrAuthentication represents an error asserting a principal's identity.
```

sdk/v2/meta/errors.go
```diff
@@ -0,0 +1,97 @@
...
+// instance, if a client attempts to authenticate to the API server using an
+// authentication mechanism that is explicitly disabled by the API server.
+type ErrNotSupported struct {
+	// Details is a natural language explanation of why the request was is not
+	// supported by the API server.
+	Details string `json:"reason,omitempty"`
+}
+
+func (e *ErrNotSupported) Error() string {
+	return e.Details
```
vdice: Do we want to prepend the details with a generic "Request not supported:" or similar?

sdk/v2/meta/meta.go
```diff
@@ -0,0 +1,56 @@
+package meta
+
+import "time"
+
+// APIVersion represents the API and major version thereof with which this
+// version of the Brigade SDK is compatible.
+const APIVersion = "brigade.sh/v2"
+
+// TypeMeta represents metadata about a resource type to help clients and
+// servers mutually head off potential confusion over types (and versions of
```
vdice:
```suggestion
// servers mutually head off potential confusion over types (and versions
```

sdk/v2/meta/meta.go
```diff
@@ -0,0 +1,56 @@
...
+// thereof) sent over the wire.
+type TypeMeta struct {
+	// Kind specifies the type of a serialized resource.
+	Kind string `json:"kind,omitempty"`
+	// APIVersion specifiesthe major version of the Brigade API with which the
```
vdice:
```suggestion
	// APIVersion specifies the major version of the Brigade API with which the
```

sdk/v2/meta/meta.go
```diff
@@ -0,0 +1,56 @@
...
+	ID string `json:"id,omitempty"`
+	// Created indicates the time at which a resource was created. This is
+	// recorded by the system. Clients must leave the value of this field set to
+	// nil when using the API to create or update resources.
+	Created *time.Time `json:"created,omitempty"`
+}
+
+// ListMeta is metadata for ordered collections of resources.
+type ListMeta struct {
+	// Continue, when non-empty is an opaque value created by and understood by an
```
vdice:
```suggestion
	// Continue, when non-empty, is an opaque value created by and understood by an
```

sdk/v2/meta/meta.go
```diff
@@ -0,0 +1,56 @@
...
+	// API operation that returns partial (pageable) results. Submitting this
+	// value with subsequent requests to the same operation specifies to that
+	// operation which page to return next.
+	Continue string `json:"continue,omitempty"`
+	// RemainingItemCount, when non-nil indicates that an API operation returned
```
vdice:
```suggestion
	// RemainingItemCount, when non-nil, indicates that an API operation returned
```

vdice (commented):
Fantastic SDK documentation 👏 .  Most comments are suggestions or very minor proposed typo/grammar amendments.  Will revisit this PR from a higher-level perspective next.

vdice:
Not sure why most/all of the review comments appear to show as 'outdated'... I think I issued them from a specific commit.  Maybe it was from a commit later amended/force-pushed?  And/or, if any are truly outdated, we can definitely just mark as resolved.

krancour:
Thanks for all the suggestions @vdice. These will keep me busy Monday morning. Enjoy your weekend!

sdk/v2/api_client.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+type apiClient struct {
+	authxClient  authx.APIClient
+	coreClient   core.APIClient
+	systemClient system.APIClient
+}
+
+// NewAPIClient returns a Brigade client.
+func NewAPIClient(apiAddress, apiToken string, allowInsecure bool) APIClient {
```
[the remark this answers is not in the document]
krancour: I also don't like unwieldy argument lists, but a benefit to the functions signatures as they stand are that you cannot instantiate a client without passing in the _required_ info `apiAddress` and `apiToken`. If those were fields in a config struct of some sort, the possibility would exist that they are accidentally left unset. I like the approach you're suggesting for _optional_ fields.

So... I think we split the difference on this maybe and say that the `allowInsecure` field moves into a new `APIClientOptions` struct, because it's not invalid to casually leave that unset-- the default behavior would be what you'd expect: that insecure connections are _not_ ignored.

Function signature would then look like this:

```
func NewAPIClient(apiAddress, apiToken string, opts restmachinery.APIClientOptions) APIClient
```

And I like this much better than what I originally had, too, because it allows for future expansion of the options without breaking the SDK.

Great suggestion!

golangci.yaml
```diff
@@ -0,0 +1,106 @@
+run:
+  concurrency: 1
+  deadline: 10m
+  skip-files:
+
+linters:
+  disable-all: true
+  enable:
+  - depguard
+  # - dupl
```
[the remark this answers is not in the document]
krancour: I'll review everything that's disabled. I know `dupl` was disabled because I was finding it a little too aggressive about finding duplicated code.

sdk/README.md
```diff
@@ -0,0 +1,103 @@
...
+would make sense for gateways (whose job is simply to broker events from
+upstream systems):
+
+```golang
+import "github.com/brigadecore/brigade/sdk/v2/core"
+
+// ...
+
+client, err := core.NewEventsClient(
+  apiAddress, // The address of the Brigade 2 Prototpye API server, beginning with http:// or https://
```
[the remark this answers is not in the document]
krancour: Weird side effect of pasting Go into markdown. I'll fix.

sdk/v2/api_client.go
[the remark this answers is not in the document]
vdice: Nice, I think that's a great compromise!

Makefile
```diff
@@ -0,0 +1,82 @@
...
+
+################################################################################
+# Containerized development environment-- or lack thereof                      #
+################################################################################
+
+ifneq ($(SKIP_DOCKER),true)
+	PROJECT_ROOT := $(dir $(realpath $(firstword $(MAKEFILE_LIST))))
+	# https://github.com/krancour/go-tools
+	# https://hub.docker.com/repository/docker/krancour/go-tools
+	GO_DEV_IMAGE := krancour/go-tools:v0.3.0
```
radu-matei, line 21: Definitely not a pressing matter, but it would be great if we had an org-level project that had all the tools we needed.

sdk/v2/core/logs.go
```diff
@@ -0,0 +1,152 @@
...
+	"time"
+
+	"github.com/brigadecore/brigade/sdk/v2/internal/restmachinery"
+)
+
+// LogsSelector represents useful criteria for selecting logs for streaming from
+// a specific container of a Worker or Job.
+type LogsSelector struct {
+	// Job specifies, by name, a Job spawned by the Worker. If this field is
+	// left blank, it is presumed logs are desired for the Worker itself.
```
radu-matei: Could the `LogsSelector` object have a field for explicitly retrieving the worker logs, rather than implicitly doing so when `Job` is empty?

sdk/v2/core/logs.go
```diff
@@ -0,0 +1,152 @@
...
+
+	logCh := make(chan LogEntry)
+	errCh := make(chan error)
+
+	go l.receiveStream(ctx, resp.Body, logCh, errCh)
+
+	return logCh, errCh, nil
+}
+
+// receiveStream is used to receive log messages as SSEs (server sent events),
```
radu-matei: I couldn't find any more references to using SSE - is this exposed to other clients as well? And if so, are SSEs the only streaming option?

sdk/v2/core/projects.go
```diff
@@ -0,0 +1,327 @@
...
+	// Roles returns a specialized client for Project Role management.
+	Roles() ProjectRolesClient
+
+	// Secrets returns a specialized client for Secret management.
+	Secrets() SecretsClient
+}
+
+type projectsClient struct {
+	*restmachinery.BaseClient
+	// rolesClient is a specialized client for Project Role managament.
```
radu-matei: typo: _managament_

sdk/v2/core/projects.go
```diff
@@ -0,0 +1,327 @@
...
+	rolesClient ProjectRolesClient
+	// secretsClient is a specialized client for Secret managament.
```
radu-matei: typo: _managament_

sdk/v2/core/logs_test.go
```diff
@@ -0,0 +1,81 @@
...
+	testSelector := LogsSelector{
+		Job:       "farpoint",
+		Container: "enterprise",
+	}
+	testOpts := LogStreamOptions{
+		Follow: true,
+	}
+	testLogEntry := LogEntry{
+		Message: "Captain's log, Stardate 41153.7. Our destination is Planet " +
+			"Deneb IV, beyond which lies the great unexplored mass of the galaxy...",
```
radu-matei: ❤️

Makefile
krancour: Agreed 100%. I will commit to having that done before 2.0 is GA. I just hadn't wanted to start there.

sdk/v2/core/workers.go
```diff
@@ -0,0 +1,381 @@
...
+	ConfigFilesDirectory string `json:"configFilesDirectory,omitempty"`
+	// DefaultConfigFiles is a map of configuration file names to configuration
+	// file content. This is useful for Workers that do not integrate with any
+	// source control system and would like to embed configuration (e.g.
+	// brigade.json) or scripts (e.g. brigade.js) directly within the WorkerSpec.
+	DefaultConfigFiles map[string]string `json:"defaultConfigFiles,omitempty"`
+}
+
+// WorkerGitConfig represents git-specific Worker details.
+type WorkerGitConfig struct {
```
radu-matei: Two questions related the worker git object:

- is there a need for it to be conceptually different from a job's git information?
- is `initSubmodules` a property that should only be available for a worker, or also for a job?

radu-matei (commented):
I have a few questions about replicated objects (particularly about the git information at the job and worker level), but otherwise, this looks great, thanks!

sdk/README.md
```diff
@@ -0,0 +1,103 @@
+# Brigade 2 SDK for Go
+
+The Brigade 2 SDK for Go provides Go language bindings for the new Brigade 2
+API.
+
+This SDK remains under development, but has been made publicly availably sooner
```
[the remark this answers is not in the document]
krancour: Thanks for all the typos you caught @radu-matei. Starting to wonder why the spell check linter isn't catching these....

sdk/v2/go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/brigadecore/brigade/sdk/v2
+
+go 1.14
```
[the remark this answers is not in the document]
krancour: That sounds good to me.

sdk/v2/core/events.go
```diff
@@ -0,0 +1,361 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/brigadecore/brigade/sdk/v2/internal/restmachinery"
+	"github.com/brigadecore/brigade/sdk/v2/meta"
+)
+
+// EventsSelector represents useful filter criteria when selecting multiple
+// Events for API group operations like list, cancel, or delete.
+type EventsSelector struct {
```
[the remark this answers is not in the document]
krancour: In my head, things were ordered according to a list retrieval workflow... so you have selector types and options types (request), followed by lists and core objects (response), and then finally the interface (last because the bulk of the code is the interface's implementation). That was admittedly pretty arbitrary, but was at least consistently applied.

But I can clearly see the rationale in ordering them by relative importance and I think that probably makes more sense, so I'm happy to rearrange.

sdk/v2/core/events.go
```diff
@@ -0,0 +1,361 @@
...
+				Kind:       "Event",
+			},
+			Alias: (Alias)(e),
+		},
+	)
+}
+
+// EventGitConfig represents git-specific Event details. These may override
+// similar details set at the Project level.
+type EventGitConfig struct {
```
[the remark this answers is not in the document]
krancour: I will take a closer look, but I know a few spots where similarly named types exist because in some contexts they need additional fields that aren't appropriate in some _other_ context. But I'll definitely revisit the question of whether it's appropriate to use composition to DRY those cases up a bit.

sdk/README.md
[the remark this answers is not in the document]
krancour: Scratch that... _of course_ Go linters don't catch misspellings in an MD. ::Grudgingly turns VS Code spell check plugin back on and retrains it that "k8s" isn't a typo.::

sdk/v2/core/jobs.go
```diff
@@ -0,0 +1,381 @@
...
+}
+
+// JobContainerSpec amends the ContainerSpec type with additional Job-specific
+// fields.
+type JobContainerSpec struct {
+	// ContainerSpec encapsulates generic specifications for an OCI container.
+	ContainerSpec `json:",inline"`
+	// UseWorkspace indicates whether the Job requires the Worker's shared
+	// workspace (if one exists) to be mounted into the OCI container.
+	UseWorkspace bool `json:"useWorkspace"`
```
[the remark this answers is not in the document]
krancour: Yes... that _is_ the shared storage. I renamed it to "workspace," because it felt cleaner and (not that we're a CI system), but "workspace" is familiar terminology for people coming from some other systems, like Jenkins, for instance.

Note one big difference between this and 1.x shared storage is that this storage is not created by the worker, but created _before_ the worker is started. The result is that the worker has access to the workspace (shared storage) which is something that's been requested a few times.

sdk/v2/core/logs.go
krancour: I had thought about that as well. There's a 1:1 between events and workers, so when you pass the eventID when retrieving logs, you have implicitly identified a worker. So it's not so much that the _absence_ of a value in the `job` field == a request for worker logs. It's more that `job` is an optional field for drilling down past what you've already asked for.

The function signature for getting logs might make that more apparent:

```
func (l *logsClient) Stream(
	ctx context.Context,
	eventID string,
	selector LogsSelector,
	opts LogStreamOptions,
) (<-chan LogEntry, <-chan error, error)
```

Thoughts? Maybe the comments on the `Job` field just need to be worded a bit better?

sdk/v2/core/logs.go
krancour: I should mention SSEs explicitly in the docs for the `steam` function of the `LogsClient` interface.

It is the only option I was thinking of providing currently-- but isn't the only one I built, initially. I had originally _also_ implemented a one-shot `GetLogs(...)`...

You can see with all resource types, I have added a `<Resource>List` object that has fields to support paging. (This, by the way, it modeled after how k8s does it.) The reason for doing that is to avoid OOMs if you try to retrieve a massive number of any sort of resource in one shot. Where logs where concerned, I started to implement paging for `GetLogs(...)` and then started asking myself why I was bothering to do that when log streaming was already a thing.

Let me know if you think there is rationale for still having something like `GetLogs(...)`. (Or let me know if I am misinterpreting your question.)

krancour: Coincidentally, our log stream works the same way k8's own log streams work.

sdk/v2/core/projects.go
krancour: Ok... _now_ I'm back to wondering if the spellcheck linter was on a coffee break.

sdk/v2/core/logs_test.go
krancour: It's a log. :)

sdk/v2/core/workers.go
krancour: Related to [this comment](https://github.com/brigadecore/brigade/pull/1118#discussion_r492276970), I'm going to double-check.

krancour: _At the very least_, it seems the docs need to differentiate more clearly between similarly named things.

golangci.yaml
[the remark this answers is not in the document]
krancour: `dupl` is accusing some bits of (test only) code that are _not_ identical of being identical. I'd encountered that on another project once before and I think that's why I had it disabled here from the get go.

`gconst` should be re-enabled, I think.

`gocyclo` has a problem with the complexity of one particular function that honestly isn't that complex. I'm going to re-enable it, but also configure it to be a little more tolerant. May _still_ have to drop a `nolint: gocyclo` on that function.

`maligned` wants to change the order of the fields on one particular struct and offers no explanation for why. The docs for that linter also do not explain what it's doing. I'm questioning its value and choosing to leave it disabled.

sdk/v2/api_client.go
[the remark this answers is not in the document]
vdice: Reviewed 2aaffbf3d69a46e6b2c4a323804a33787d4a7ef9 - LGTM, thank you!

sdk/v2/core/workers.go
krancour: @radu-matei, so...

`WorkerGitConfig` is used when defining the (`WorkerSpec`) "template" that a Project will use for creating new Workers in response to an Event. This says where to get source (possibly including things like `brigade.js`) from and, yes, indicates whether or not submodules should be initialized. (These are details that the interactive `brig project create` asks about for VCS projects in v1.x.)

What I think is definitely _not_ clear from the name of this type (or from the docs) is that this configuration is applied _not only_ to the init container of a Worker in need of source, but applies _equally_ to init containers of _Jobs_ (spawned by that Worker) if they too are in need of source. (And this, isn't different from 1.x either.)

`EventGitConfig` has a related, but different purpose. It gives individual Events the possibility of overriding the Project-level git config (described above). The `CloneURL`, for instance, may be overridden if the Event should prompt Brigade to clone source from a fork. `Commit` or `Ref` may similarly override Project-level fields of the same name, directing a specific commit, branch, tag, etc. to be pulled. (Again, I think no different from 1.x.) I cannot imagine the matter of whether submodules should or shouldn't be initialized to be something that the Event has any business overriding-- that seems more appropriately determined only by whoever set up the Project.

So there's a relationship between these two things, for sure. And for sure, one of them rightfully has one field that the other lacks and the rest of the fields are common... But it's not clear to me if there's an opportunity to DRY things up through composition here-- _at least not in a way that makes sense (to me)._

We _could_ do this, for instance:

```
type EventGitConfig struct {
  CloneURL string `json:"cloneURL,omitempty"`
  Commit string `json:"commit,omitempty"`
  Ref string `json:"ref,omitempty"`
}

type WorkerGitConfig struct {
  EventGitConfig `json:",inline"`
  InitSubmodules bool `json:"initSubmodules"`
}
```

This is DRYer, but it makes no sense that any Project-level configuration (the Worker template) would be composed of seemingly Event-specific details.

Then I thought about toying with the naming to make it make more sense:

```
type GitConfig struct {
  CloneURL string `json:"cloneURL,omitempty"`
  Commit string `json:"commit,omitempty"`
  Ref string `json:"ref,omitempty"`
}

type WorkerGitConfig struct {
  GitConfig `json:",inline"`
  InitSubmodules bool `json:"initSubmodules"`
}
```

This looks considerably better, I think, until you try to build out a Project and `GitConfig` "stutters" like so:

```
project := Project {
  Spec: ProjectSpec {
    WorkerTemplate: WorkerSpec {
      Git: &WorkerGitConfig {
        GitConfig: GitConfig { ... }, // This stutters, without any intuitive reason for doing so
      },
    },
  },
}
```

So there's the thought process. My conclusion was that clear was better than clever here, even if it was less DRY.

But I am totally open to suggestions for untangling this little knot because even if I think it's the least awkward of the possibilities I considered, I still think it could be improved.

krancour:
@radu-matei @vdice I believe everything related to spelling, grammar, or simple code fixes has been addressed. The only outstanding thing (unless I missed something) is [this thread](https://github.com/brigadecore/brigade/pull/1118#discussion_r492264327).

sdk/v2/core/workers.go
krancour: Slept on it and had a revelation...

Events don't have configuration; they have _details_.

I believe everything might be more clear if `WorkerSpec` has a field of type `GitConfig` whilst `Event` has a field of type `GitDetails`.

`GitConfig` and `GitDetails` may have fields in common (that repeat), but perhaps that's the small price we pay for favoring clarity over cleverness.

Would love thoughts on this.

sdk/v2/core/logs.go
radu-matei: Yeah, a comment describing this behavior would definitely help.

sdk/v2/core/logs.go
radu-matei: I am mostly asking from the perspective of someone implementing multiple dashboards for getting logs (web and CLI dashboards), where a plain `GET` request is always simpler compared to anything else.

I would explore other server push ways as well (directly using HTTP/2 streams, for example).

sdk/v2/core/workers.go
radu-matei: I am on board with the distinction between `GitConfig` and `GitDetails` you just made.

vdice:
Thanks @krancour !  I think there remain a few more comments to look at (they may have been hidden/compressed previously).  I believe I've resolved all that have been addressed (please feel free to do so as well).

sdk/v2/core/logs.go
krancour: > directly using HTTP/2 streams, for example

Isn't that what I've done? I modeled the code after the clockstream example on https://http2.golang.org/

sdk/v2/core/workers.go
krancour: > although it is definitely not blocking this PR.

I'll fix it now. I think it improves clarity substantially.

krancour:
> I think there remain a few more comments to look at (they may have been hidden/compressed previously). 

@vdice I'll double-check!

sdk/v2/go.mod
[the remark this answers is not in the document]
krancour: Done.

sdk/v2/core/events.go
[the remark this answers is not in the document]
krancour: Compromised on this... moved structs like `Event` before things like `EventList` and `EventsSelector`, but left interfaces like `EventsClient` and structs like `eventsClient` farther down, because the bulk of each file is implementation of those.

sdk/v2/core/workers.go
krancour: There was an opportunity to fix some similar confusion pertaining to `KubernetesConfig` (and other similarly named types). I fixed that as well.

sdk/v2/core/events.go
```diff
@@ -0,0 +1,361 @@
...
+}
+
+// EventsClient is the specialized client for managing Events with the Brigade
+// API.
+type EventsClient interface {
+	// Create creates a new Event.
+	Create(context.Context, Event) (EventList, error)
+	// List returns an EventList, with its Items (Events) ordered by age, newest
+	// first. Criteria for which Events should be retrieved can be specified using
+	// the EventListOptions parameter.
```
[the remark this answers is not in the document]
krancour: This was a case of something being renamed and the doc still having a reference to the old name. Fix incoming shortly.

sdk/v2/core/events.go
```diff
@@ -0,0 +1,361 @@
...
+type DeleteManyEventsResult struct {
+	// Count represents the number of Events deleted.
+	Count int64 `json:"count"`
+}
+
+// EventsClient is the specialized client for managing Events with the Brigade
+// API.
+type EventsClient interface {
+	// Create creates a new Event.
+	Create(context.Context, Event) (EventList, error)
```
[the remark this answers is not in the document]
krancour: Yes. That should be spelled out more clearly.

The reason has to do with event subscriptions. Since one might submit a new event that isn't for a specific project, but is matched to _subscribed_ projects, more than one event might actually be created. This returns all the events that were created. This way if you're a gateway, for instance, you end up with a discrete list of events that you can then do whatever with-- perhaps monitor for completion in order to report results upstream.

sdk/v2/core/jobs.go
```diff
@@ -0,0 +1,381 @@
...
+	// within a designated timeframe.
+	JobPhaseTimedOut JobPhase = "TIMED_OUT"
+	// JobPhaseUnknown represents the state wherein a Job's state is unknown. Note
+	// that this is possible if and only if the underlying Job execution substrate
+	// (Kubernetes), for some unanticipated, reason does not know the Job's
+	// (Pod's) state.
+	JobPhaseUnknown WorkerPhase = "UNKNOWN"
+)
+
+// Job represents a component spawned by a Worker to complete a single task
```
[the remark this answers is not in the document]
krancour: @vdice Good question. And I think this raises a slight design change that might warrant further discussion.

cc @radu-matei 

If you look at how jobs are now defined here (`JobSpec`s, actually), they no longer take an array of "tasks" as they once did. Instead they have a `Command` array and an `Arguments` array.

The primary reason I changed this is that the 1.x worker did a fair amount of magic to take an array of discrete "tasks" and turn it into a shell script-- and in doing so made fairly large assumptions that Linux containers all include `bash` and that Windows containers all include PowerShell. These assumptions aren't universally valid and they specifically preclude the possibility of using a lot of small, simple images that _don't_ include those shells. 2.0 workers will use the API to push Job creation responsibility to the API server, and rather than port those large assumptions to the backend, I made the call to change up how jobs are defined.

The `Command` array and an `Arguments` array I previously referred to _admittedly_ may end up feeling clumsier to work with, _but_ it's in 100% parity with how Docker CLI, Docker packages, Docker Compose configuration files, and Kubernetes pods ask one to define what is to be executed within the container. i.e., the new approach reuses a familiar and widespread precedent.

So... in the end, I think it's not misleading to say a Job represents a "single task."

With proper syntax, one can of course effect execution of multiple commands and there is always the option to use a single command to invoke a script-- which, not insignificantly, is made all the more easier by the fact that arbitrary files are so easily embedded into a project definition now-- a useful pattern that should absolutely be covered in end-user docs.

sdk/v2/core/jobs.go
```diff
@@ -0,0 +1,381 @@
...
+			},
+			Alias: (Alias)(j),
+		},
+	)
+}
+
+// JobsClient is the specialized client for managing Event Jobs with the
+// Brigade API.
+type JobsClient interface {
+	// Create, given an Event identifier and JobSpec, creates a new pending Job
```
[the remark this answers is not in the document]
krancour: This function used to take a JobSpec. I changed it recently and the docs didn't catch up. Fixing it. _Good eye._

sdk/v2/core/jobs_test.go
krancour: You can tell I had some fun with naming test fixtures.

sdk/v2/core/projects.go
krancour: I should spell that out, yes. Need to glance at the prototype backend to remember how it works-- I think it's a required field and should be set to "*" if you're interested in all types. I'll double-check.

sdk/v2/core/projects.go
krancour: They're all required. I'm fixing the docs to reflect that.

sdk/v2/core/jobs.go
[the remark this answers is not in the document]
vdice: Ah, thank you for the explanation.  I'm persuaded that moving to `Command` and `Arguments` is the right choice, thanks to the reasoning above.  Wording indeed seems okay to stay as is.

sdk/v2/core/secrets.go
krancour: `omitempty` doesn't have any real bearing on whether something is required or not. It really only affects the marshaled representation.

The question of validity is a server-side concern. The prototype API server would reject any secret that was lacking a value for either of those fields.

As for whether these fields should be marked "required" in the docs, I'm hesitant, because a secret, as with most other resources, can be sent in both directions. You can send a secret TO the API Server or retrieve one FROM it. When retrieving, it's perfectly plausible that the `Value` field would be empty because the prototype API server redacts the values on retrieval (and I expect our official API server will function the same way).

sdk/v2/core/workers.go
krancour: Ooops.

sdk/v2/internal/restmachinery/client_test.go
krancour: Nope... and already removed in a recent commit, iirc, so I'll mark this as resolved.

sdk/v2/internal/restmachinery/outbound_request.go
krancour: Nice catch. Fallout from a last-minute refactor.

sdk/v2/core/projects.go
krancour: lmk if the recent updates add clarity.

sdk/v2/core/events.go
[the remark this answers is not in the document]
krancour: lmk if the recent commits add sufficient clarity.

sdk/v2/core/projects.go
krancour: lmk if the recent commits add sufficient clarity.

sdk/v2/core/workers.go
krancour: Fixed.

sdk/v2/core/events.go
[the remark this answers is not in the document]
krancour: I think all of this was resolved in one of the other conversations.

sdk/v2/core/logs.go
krancour: Fix incoming shortly.

sdk/v2/core/logs.go
krancour: Err... maybe that isn't super apparent without seeing the server-side implementation.

Here's a link to relevant code in the prototype API server:

https://github.com/krancour/brignext/blob/master/v2/apiserver/internal/core/rest/logs_endpoints.go#L63-L73

sdk/v2/core/logs.go
krancour: @radu-matei please lmk if db586274cce1fb9f5beb07f37111d5d1eafe1413 clears things up.

krancour:
> I think there remain a few more comments to look at (they may have been hidden/compressed previously).

The things I missed were mostly things "folded over" with a "load more" link.

I've revisited everything and took the liberty of marking things resolved if they seem that they clearly are.

There're just a couple things left for @vdice and @radu-matei to revisit my explanations and/or changes.

sdk/v2/core/logs.go
radu-matei: Ah, that is excellent - this is definitely something we want to revisit with the API server, but this seems reasonable (HTTP/2 would have been my first choice).

radu-matei (approved):

sdk/v2/core/secrets.go
vdice: Thanks for the explanation, especially helpful to remember that this object may take different forms depending on transmission type.

sdk/v2/core/projects.go
vdice: Thanks!

sdk/v2/core/events.go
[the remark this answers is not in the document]
vdice: Yup, thank you for the updates.

sdk/v2/core/projects.go
```diff
@@ -103,15 +103,27 @@ type ProjectSpec struct {
...
+	// Source specifies the origin of an Event (e.g. a gateway). This is a
+	// required field.
 	Source string `json:"source,omitempty"`
-	// Types enumerates specific Events of interest from the specified source.
-	// This is useful in narrowing a subscription when a source also emits many
-	// events that are NOT of interest.
+	// Types enumerates specific Events of interest from the specified Source.
+	// This is useful in narrowing a subscription when a Source also emits many
+	// Event types that are NOT of interest. This is a required field. The value
+	// "*" may be utilized to denote that ALL events from originating from the
```
vdice:
```suggestion
	// "*" may be utilized to denote that ALL events originating from the
```

vdice (approved):
LGTM!  One last grammar fix spotted; otherwise good to go.  Thanks, @krancour!

sdk/v2/core/projects.go
vdice: Yes, updated godoc really helps; thank you.

sdk/v2/core/projects.go
krancour: Will fix this prior to merge.

PR #1122, opened 2020-10-08
implement login and logout

krancour:
This PR adds login / logout for users authenticating as "root" or using OpenID Connect.

This is the first PR to include any server-side functionality, so necessarily, a lot of infra bits included are in this PR as well.

Highlights:

- Server-side "REST machinery"
- CLI with login and logout commands
- API server, including Dockerfile
- Helm chart to stand up API server + MongoDB

The four constituent commits are broken up thematically, so reviewers may find it useful to review commit-by-commit.

Future PRs will be smaller since this one wrangles all the infra bits into place.


v2/apiserver/internal/authx/mongodb/common_test.go
```diff
@@ -0,0 +1,173 @@
+package mongodb
```
technosophos, line 1: WOW!

v2/cli/login_command.go
```diff
@@ -0,0 +1,121 @@
...
+
+import (
+	"fmt"
+
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/bacongobbler/browser"
+	"github.com/brigadecore/brigade/sdk/v2/authx"
+	"github.com/brigadecore/brigade/sdk/v2/restmachinery"
+	"github.com/pkg/errors"
+	"github.com/urfave/cli/v2"
```
technosophos, line 11: Ooooo! I had not seen cli/v2!

v2/internal/rand/seeded.go
```diff
@@ -0,0 +1,34 @@
...
+	Intn(max int) int
+}
+
+type seeded struct {
+	seededRand *mathrand.Rand
+	mut        *sync.Mutex
+}
+
+// NewSeeded returns a seeded, concurrency-safe random number generator.
+func NewSeeded() Seeded {
```
technosophos, line 22: This is a nice shim. I wish Go had this built-in.

technosophos (approved):
Wow! That was a lot of work!

LGTM

technosophos:
I looked at the CI error, but couldn't figure out what went wrong.

v2/cli/login_command.go
krancour: Awesome thing that v2 adds is ability to make certain flags required. Taking cues from the `az` CLI, I've taken advantage of that quite a bit. If a command takes several arguments, it's _way_ easier to use flags than to try and remember what _position_ each argument belongs in.

v2/internal/rand/seeded.go
krancour: :sigh: I opened an issue about that... 2+ years ago.

krancour:
> I looked at the CI error, but couldn't figure out what went wrong.

@technosophos there's one flaky test. It's a test that the server shuts down when it receives an interrupt and I think there's a race (in the test; not in the actual code) where the interrupt might be sent before we start listening for it.

I'm re-running the tests and plan to fix this flake later, if it's all the same to you.

v2/apiserver/internal/authx/rest/sessions_endpoints.go
```diff
@@ -0,0 +1,132 @@
...
+	// OIDC callback
+	router.HandleFunc(
+		"/v2/session/auth",
+		s.authenticate, // No filters applied to this request
+	).Methods(http.MethodGet)
+}
+
+func (s *SessionsEndpoints) create(w http.ResponseWriter, r *http.Request) {
+	// nolint: errcheck
+	rootSessionRequest, _ := strconv.ParseBool(r.URL.Query().Get("root"))
```
radu-matei, line 45: Is there no scenario where this could fail?

v2/apiserver/internal/lib/restmachinery/authn/token_auth.go
```diff
@@ -0,0 +1,226 @@
...
+// Decorate decorates one http.HandlerFunc with another.
+func (t *tokenAuthFilter) Decorate(handle http.HandlerFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		headerValue := r.Header.Get("Authorization")
+		if headerValue == "" {
+			t.writeResponse(
+				w,
+				http.StatusUnauthorized,
+				&meta.ErrAuthentication{
+					Reason: `Authorization" header is missing.`,
```
radu-matei, line 92: Missing `"`.

v2/apiserver/internal/authx/rest/sessions_endpoints.go
krancour: Arguably, there isn't.

```
// ParseBool returns the boolean value represented by the string.
// It accepts 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False.
// Any other value returns an error.
func ParseBool(str string) (bool, error) {
	switch str {
	case "1", "t", "T", "true", "TRUE", "True":
		return true, nil
	case "0", "f", "F", "false", "FALSE", "False":
		return false, nil
	}
	return false, syntaxError("ParseBool", str)
}
```

The only way that function call produces an error is if the input is not recognized as a string representation of `true` or `false`, but in that case, `false` is _also_ returned. So the function effectively returns `true` _if and only if_ the input is a recognized string representation of `true` and false for all other cases.

If we agree that anything that isn't "truthy" is false, then this works the way we'd expect without any need to check the error.

v2/apiserver/internal/lib/restmachinery/authn/token_auth.go
```diff
@@ -0,0 +1,226 @@
...
+			r.Header.Get("Authorization"),
+			" ",
+			2,
+		)
+		if len(headerValueParts) != 2 || headerValueParts[0] != "Bearer" {
+			t.writeResponse(
+				w,
+				http.StatusUnauthorized,
+				&meta.ErrAuthentication{
+					Reason: `Authorization" header is malformed.`,
```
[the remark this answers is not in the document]
krancour: Good eye.

v2/apiserver/internal/authx/common_test.go
```diff
@@ -0,0 +1,44 @@
+package authx
+
+import (
+	"encoding/json"
+	"reflect"
+	"testing"
+	"unsafe"
+
+	"github.com/brigadecore/brigade/v2/apiserver/internal/meta"
```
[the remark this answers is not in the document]
krancour: @radu-matei I am missing something here, possibly...

Is there something wrong with the import order that neither I or the linter are spotting?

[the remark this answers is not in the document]
krancour: Going to not let this block the PR since @technosophos has LGTM'ed. But please do open an issue if there is something outstanding here that I'm not spotting.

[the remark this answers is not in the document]
radu-matei: I should have been more explicit - IIRC, the previous linter settings for import order were enforcing something like:

https://github.com/brigadecore/brigade/blob/66dfceb3de22d5a735615d82d31077bf6a512a12/pkg/storage/kube/worker.go#L3-L15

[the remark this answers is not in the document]
krancour: Oh... new linter settings got copied over from the prototype. I'm not opposed to what you have above. It makes sense to me. Do you know if gofmt or goimports can be configured somehow to format imports that way automatically?

PR #1125, opened 2020-10-14
secret management features

krancour:
Manage project secrets...

### DIFF
--- a/v2/apiserver/internal/core/secrets.go
+++ b/v2/apiserver/internal/core/secrets.go
@@ -90,7 +90,7 @@ type SecretsService interface {
 		projectID string,
 		opts meta.ListOptions,
 	) (SecretList, error)
-	// Set set the value of a new Secret or updates the value of an existing
+	// Set sets the value of a new Secret or updates the value of an existing
 	// Secret. If the specified Project does not exist, implementations MUST
 	// return a *meta.ErrNotFound error. If the specified Key does not exist, it
 	// is created. If the specified Key does exist, its corresponding Value is
--- a/v2/apiserver/internal/core/secrets.go
+++ b/v2/apiserver/internal/core/secrets.go
@@ -1,0 +1,215 @@
+package core
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/brigadecore/brigade/v2/apiserver/internal/meta"
+	"github.com/pkg/errors"
+)
+
+// SecretList is an ordered and pageable list of Secrets.
+type SecretList struct {
+	// ListMeta contains list metadata.
+	meta.ListMeta `json:"metadata"`
+	// Items is a slice of Secrets.
+	Items []Secret `json:"items,omitempty"`
+}
+
+// All of the following functions are implemented on the SecretsList type to
+// facilitate sorting by each constituent Secret's Key field.
+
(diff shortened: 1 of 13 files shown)


v2/apiserver/main.go
```diff
@@ -41,26 +47,23 @@ func main() {
...
 	// Substrate
-	var substrate core.Substrate
-	{
-		kubeClient, err := kubernetes.Client()
-		if err != nil {
-			log.Fatal(err)
-		}
-		substrate = coreKubernetes.NewSubstrate(kubeClient)
-	}
+	substrate := coreKubernetes.NewSubstrate(kubeClient)
```
radu-matei, line 56: This is so much cleaner!

v2/cli/secrets_commands.go
```diff
@@ -0,0 +1,224 @@
...
+	return nil
+}
+
+func secretsSet(c *cli.Context) error {
+	projectID := c.String(flagProject)
+	kvPairsStr := c.StringSlice(flagSet)
+
+	// We'll make two passes-- we'll parse all the input into a map first,
+	// verifying as we go that the input looks good. Only after we know it's good
+	// will we iterate over the k/v pairs in the map to set secrets via the API.
```
radu-matei, line 159: Would it make sense to check that the key and value conform to the schema, before sending it to the API?

v2/cli/secrets_commands.go
```diff
@@ -0,0 +1,224 @@
...
+				&cli.StringFlag{
+					Name:     flagProject,
+					Aliases:  []string{"p"},
+					Usage:    "Clear secrets for the specified project",
+					Required: true,
+				},
+				&cli.StringSliceFlag{
+					Name:     flagUnset,
+					Aliases:  []string{"u"},
+					Usage:    "Clear a secret haveing the specified key (required)",
```
radu-matei, line 70: typo: _haveing_

v2/apiserver/internal/core/secrets.go
```diff
@@ -0,0 +1,215 @@
...
+type SecretsService interface {
+	// List returns a SecretList whose Items (Secrets) contain Keys only and
+	// not Values (all Value fields are empty). i.e. Once a secret is set, end
+	// clients are unable to retrieve values.
+	List(
+		ctx context.Context,
+		projectID string,
+		opts meta.ListOptions,
+	) (SecretList, error)
+	// Set set the value of a new Secret or updates the value of an existing
```
radu-matei, line 93: _"Set set[s] the value [...] or updates"_

radu-matei (approved):

v2/cli/secrets_commands.go
krancour: There are actually several spots where I debated whether it would be a good idea to do some client-side validation so that we could short-circuit and avoid a round-trip for something that was destined to fail, but I eventually came down on the side of keeping all validation server-side only because if you don't and you then need to relax or tighten the schema, you don't want to have to do that both client-side _and_ server-side-- or risk them disagreeing about what is valid. Opted to have the api server be the supremer arbiter of valid / invalid.

radu-matei: I didn't consider that different clients could try enforcing different validation rules - in this case, this makes sense.

PR #1126, opened 2020-10-15
fix job marshaling issue in sdk

krancour:
The API client was accidentally applying type metadata to outbound `JobSpec`s, but this is incorrect. `JobSpec`s are never sent directly from client to server, rather `Job`s _containing_ specs are sent. `Job` was lacking the correct application of that type metadata.

This mistake arose due to some refactoring that occurred as code was being imported from the prototype and cleaned up. It wasn't caught because job-related API server endpoints do not yet exist to integration test this part of the client against.

### DIFF
--- a/sdk/v2/core/jobs.go
+++ b/sdk/v2/core/jobs.go
@@ -51,6 +51,24 @@ type Job struct {
 	Status *JobStatus `json:"status"`
 }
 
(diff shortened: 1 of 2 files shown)


technosophos (approved):
LGTM

PR #1127, opened 2020-10-15
fix test schema uri

krancour:
This is just a vanity fix. Some code involved in testing the server-side `restmachinery` uses a fake JSON schema that was copied from another project of mine, and as a result of the careless copy/paste, the `$id` for that schema looks funky. This PR fixes that.


technosophos (approved):
LGTM

PR #1152, opened 2020-10-27
Add abstraction for writing to a messaging system

krancour:
This will ultimately be what enables the substrate component to communicate asynchronously with the scheduler.

This PR stops just short of actually dropping messages onto the queue because more extensive substrate additions are required before that can be done. Expect those changes in my next PR.

As always, I've taken care to organize the commits in a way that should make each of them a smaller / easier unit to review.

### DIFF
--- a/v2/apiserver/internal/lib/queue/amqp/writer.go
+++ b/v2/apiserver/internal/lib/queue/amqp/writer.go
@@ -1,0 +1,200 @@
+package amqp
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	amqp "github.com/Azure/go-amqp"
+	myamqp "github.com/brigadecore/brigade/v2/apiserver/internal/lib/amqp"
+	"github.com/brigadecore/brigade/v2/apiserver/internal/lib/queue"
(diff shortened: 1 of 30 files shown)


v2/apiserver/internal/lib/queue/amqp/writer.go
```diff
@@ -0,0 +1,200 @@
+package amqp
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	amqp "github.com/Azure/go-amqp"
+	myamqp "github.com/brigadecore/brigade/v2/apiserver/internal/lib/amqp"
```
radu-matei, line 9: Not particularly a fan of the `myamqp` import alias (but obviously, this is as nitty as is gets).

radu-matei (approved):

v2/apiserver/internal/lib/queue/amqp/writer.go
krancour: I don't like it either. I'm extremely open to alternatives.

PR #1153, opened 2020-10-28
on event creation, use queue to schedule worker execution

krancour:
This builds on top of #1152, which must be merged first.

### DIFF
--- a/v2/apiserver/internal/core/common_test.go
+++ b/v2/apiserver/internal/core/common_test.go
@@ -131,7 +131,7 @@ type mockSubstrate struct {
 		project Project,
 	) (Project, error)
 	DeleteProjectFn       func(context.Context, Project) error
-	CreateEventFn         func(context.Context, Project, Event) (Event, error)
+	ScheduleWorkerFn      func(context.Context, Project, Event) error
 	DeleteWorkerAndJobsFn func(context.Context, Project, Event) error
 }
 
@@ -149,11 +149,12 @@ func (m *mockSubstrate) DeleteProject(
 	return m.DeleteProjectFn(ctx, project)
 }
 
-func (m *mockSubstrate) CreateEvent(
+func (m *mockSubstrate) ScheduleWorker(
 	ctx context.Context,
 	project Project,
-	event Event) (Event, error) {
-	return m.CreateEventFn(ctx, project, event)
+	event Event,
+) error {
+	return m.ScheduleWorkerFn(ctx, project, event)
 }
 
 func (m *mockSubstrate) DeleteWorkerAndJobs(
--- a/v2/apiserver/internal/core/events_test.go
+++ b/v2/apiserver/internal/core/events_test.go
@@ -233,7 +233,7 @@ func TestEventsServiceCreateSingleEvent(t *testing.T) {
 		assertions func(Event, error)
 	}{
 		{
-			name: "error creating project in store",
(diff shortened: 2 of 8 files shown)


v2/apiserver/internal/core/common_test.go
```diff
@@ -131,7 +131,7 @@ type mockSubstrate struct {
 		project Project,
 	) (Project, error)
 	DeleteProjectFn       func(context.Context, Project) error
-	CreateEventFn         func(context.Context, Project, Event) (Event, error)
```
krancour, line 134: This got removed because the interface this mock fulfills no longer requires it-- in fact it hasn't for a while, so this is dead code.

v2/apiserver/internal/core/events_test.go
```diff
@@ -233,7 +233,7 @@ func TestEventsServiceCreateSingleEvent(t *testing.T) {
 		assertions func(Event, error)
 	}{
 		{
-			name: "error creating project in store",
+			name: "error creating event in store",
```
krancour, line 236: Old test case name contained a typo.

v2/apiserver/internal/core/kubernetes/substrate.go
```diff
@@ -251,8 +278,147 @@ func (s *substrate) DeleteProject(
...
+		APIAddress           string            `json:"apiAddress"`
+		APIToken             string            `json:"apiToken"`
+		LogLevel             core.LogLevel     `json:"logLevel"`
+		ConfigFilesDirectory string            `json:"configFilesDirectory"`
+		DefaultConfigFiles   map[string]string `json:"defaultConfigFiles" bson:"defaultConfigFiles"` // nolint: lll
+	}
+
+	// Create a secret with event details
+	eventJSON, err := json.MarshalIndent(
+		struct {
```
vdice, line 322: Q: why not use the `Event` struct here?

vdice (approved):
LGTM, just had one question...

v2/apiserver/internal/core/kubernetes/substrate.go
krancour: Good question. Initially I did, but the Event struct only had a reference to a project (projectID) and I want the worker to have access to (some) project details. Options are use a different struct that encapsualtes a copy of the (very few) relevant project details or else use a completely separate chunk of JSON, which seemed it would complicate things a bit on the brigade.js end.

krancour: @vdice Amending my previous answer... I was thinking about this more and I think I really need to justify why replicating some of the `Event` fields in a second struct does not constitute a maintenance burden going forward. i.e. Can I convince myself that every time a field gets (theoretically) added to the `Event` struct that it doesn't automatically _have_ to be replicated here as well.

I believe what we have here is a _worker-specific representation_ of an `Event`. Some `Event` fields are not of interest to a worker. A worker also needs to know certain _extra_ things that aren't part of the `Event` struct. The example I gave it the previous comment included project details (including secrets), but there are others as well. The `Worker` field here, for instance, has a field of its own-- `APIAddress`-- which isn't part of an `Event` but is information that a worker requires in order to _process_ the event...

Worker-specific event representation.

vdice: Definitely.  Thanks for the explanation.

PR #1154, opened 2020-10-29
add abstraction for reading messages from a messaging system

krancour:
This complements #1152 to add an abstraction for reading messages from an underlying messaging system-- and an amqp-based implementation of that abstraction. This will eventually be used by a scheduler component.

This duplicates a fair amount of boilerplate AMQP connection-management code from #1152. I intend to shortly follow up on this PR with another that merges relevant packages so that they can share more boilerplate code, _but_ I think the changes in this PR are more comprehensible as-is, which is why I'm tackling this as two PRs instead of one.


v2/scheduler/internal/lib/queue/amqp/reader.go
```diff
@@ -0,0 +1,204 @@
...
+		amqpSession:  amqpSession,
+		amqpReceiver: amqpReceiver,
+	}, nil
+}
+
+func (q *readerFactory) Close(context.Context) error {
+	if err := q.amqpClient.Close(); err != nil {
+		return errors.Wrapf(err, "error closing AMQP client")
+	}
+	log.Println("DEBUG: closed AMQP-based queue reader factory")
```
vdice, line 160: Is this statement intended to remain?  Will we eventually want to conditionally print based on a TBD log level?

v2/scheduler/internal/lib/queue/amqp/reader_test.go
```diff
@@ -0,0 +1,326 @@
...
+				os.Setenv("AMQP_USERNAME", "bar")
+			},
+			assertions: func(_ ReaderFactoryConfig, err error) {
+				require.Error(t, err)
+				require.Contains(t, err.Error(), "value not found for")
+				require.Contains(t, err.Error(), "AMQP_PASSWORD")
+			},
+		},
+		{
+			name: "SUCCESS not set",
```
vdice, line 57: nit: this test name is a bit confusing.  Maybe just "Success" or "GetReaderFactoryConfig Success"?

vdice (approved):
LGTM

v2/scheduler/internal/lib/queue/amqp/reader.go
krancour: Nice catch. I intended to remove that. I think introducing a leveled logger would be a great idea. Our obligations are v2-alpha by end of the year and we have _plenty_ of time after that to focus on stability and other non-functional requirements, so if you're ok with it, let's defer leveled logging until then.

I will, however, remove this line that I'd intended to remove prior to merging.

v2/scheduler/internal/lib/queue/amqp/reader_test.go
krancour: Wow. I don't even know what happened there. I will straighten that out prior to merging.

PR #1157, opened 2020-11-03
add scheduler component

krancour:
This listens for queued workers and jobs and uses the API to kick them off when available system capacity permits.

But don't get _too_ excited. Any worker that gets kicked off should fail because the new worker image isn't PR'ed yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,17 +99,17 @@ test-unit-go:
 build: build-images xbuild-cli
 
 .PHONY: build-images
-build-images: build-apiserver
+build-images: build-apiserver build-scheduler
 
-.PHONY: build-apiserver
-build-apiserver:
+.PHONY: build-%
+build-%:
 	docker build \
-		-f v2/apiserver/Dockerfile \
-		-t $(DOCKER_IMAGE_PREFIX)brigade-apiserver:$(IMMUTABLE_DOCKER_TAG) \
+		-f v2/$*/Dockerfile \
+		-t $(DOCKER_IMAGE_PREFIX)brigade-$*:$(IMMUTABLE_DOCKER_TAG) \
 		--build-arg VERSION='$(VERSION)' \
 		--build-arg COMMIT='$(GIT_VERSION)' \
 		.
-	docker tag $(DOCKER_IMAGE_PREFIX)brigade-apiserver:$(IMMUTABLE_DOCKER_TAG) $(DOCKER_IMAGE_PREFIX)brigade-apiserver:$(MUTABLE_DOCKER_TAG)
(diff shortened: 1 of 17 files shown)


Makefile
```diff
@@ -99,17 +99,17 @@ test-unit-go:
 build: build-images xbuild-cli
 
 .PHONY: build-images
-build-images: build-apiserver
+build-images: build-apiserver build-scheduler
```
radu-matei, line 102: Just a reminder that last time I tried building with `SKIP_DOCKER=true`, it was still trying to use Docker for some targets.

krancour: What targets? SKIP_DOCKER only prevents non-docker commands (like go test) from running in a container. Commands that build images will still use docker.

radu-matei: Sure, I think I was trying to run the Go tests outside a container (I will try again when properly reviewing this PR).

radu-matei: Ultimately, what I would very much like are targets to build code locally, without a container.

Sure, targets to build images have to use Docker, but while working on the project locally, it would be very helpful to be able to do this.

charts/brigade/values.yaml
```diff
@@ -125,6 +125,35 @@ apiserver:
     type: NodePort
     # nodePort: 31600
 
+scheduler:
+
+  image:
+    repository: krancour/brigade-scheduler
```
radu-matei, line 131: I assume this is a temporary image pointing to your registry?

krancour: For some reasons I recall having fixed this throughout `values.yaml` recently. Looking into that. Maybe that was in another branch.

In practice, it's only a placeholder at the moment. `make hack` builds and deploys the whole shooting match and it will override that value with the name of whatever image was just built.

This will become more relevant when we start actually building and releasing the chart (even as preview).

krancour: Ok... I wasn't imagining it. I have a s/krancour/brigadecore/ queued up in one of my other PRs.

radu-matei: Sure, this can be addressed later, just wanted to make sure we don't forget about it.

Makefile
krancour: > Ultimately, what I would very much like are targets to build code locally, without a container.

That should already be possible with `SKIP_DOCKER=true`. If that's not working for some reason, let's fix that rather than give up on the option to do things within a container.

fwiw, I understand not wanting to always use containers if they are slowing you down, but there's definitely a benefit to being able to execute things such as tests and CLI builds in an environment that's identical to where those things will run in CI, so I don't want to lose that.

radu-matei: I definitely wouldn't suggest stopping building in containers, but for example, right now it's impossible to build the API server locally without Docker.

radu-matei (approved):
LGTM (with one comment that is not really about this PR specifically)

PR #1159, opened 2020-11-05
move env vars --> config object functions to api server main package

### LINKED ISSUE #1158: move funcs that generate config from env var to main packages
There are a few functions scattered throughout the code that create configuration objects (used by various other components) based on env vars.

These functions are generally close (i.e. in the same package) as the configuration they generate.

This isn't _awful_ but it's not ideal and it's perhaps a bit short-sighted... I'm proposing that in an ideal world, only the `main` package binds env vars to any specific configuration. This way, other packages are less opinionated and more reusable.

This should be a pretty easy fix.

krancour:
Fixes #1158

There are no functional changes in this PR.

Besides moving a few functions, this PR also adds new test coverage for most of those.


technosophos:
REviewing this one now

technosophos (approved):
Nice refactoring. This feels cleaner.

PR #1162, opened 2020-11-11
add api endpoints and client for getting insight into the substrate

krancour:
This is another PR aimed at reducing the size of my forthcoming scheduler PR.

While hacking on the scheduler, I realized there was a good opportunity to abstract the scheduler away from the underlying substrate by adding API endpoints that can give the scheduler the insight it needs into how many workers and jobs are currently running on the substrate.

i.e. This will stop the forthcoming scheduler component from having to be Kubernetes-specific in any way.

### DIFF
--- a/v2/apiserver/internal/core/kubernetes/substrate.go
+++ b/v2/apiserver/internal/core/kubernetes/substrate.go
@@ -14,10 +14,41 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/client-go/kubernetes"
 )
+
+var workerPodsSelector = labels.Set(
+	map[string]string{
+		myk8s.LabelComponent: "worker",
+	},
+).AsSelector().String()
+
+var jobPodsSelector = labels.Set(
+	map[string]string{
+		myk8s.LabelComponent: "job",
+	},
+).AsSelector().String()
+
+var runningPodsSelector = fields.Set(
+	map[string]string{
+		"status.phase": string(corev1.PodRunning),
+	},
+).AsSelector().String()
+
+var pendingPodsSelector = fields.Set(
+	map[string]string{
+		"status.phase": string(corev1.PodPending),
+	},
+).AsSelector().String()
+
+var unknownPhasePodsSelector = fields.Set(
+	map[string]string{
+		"status.phase": string(corev1.PodUnknown),
+	},
+).AsSelector().String()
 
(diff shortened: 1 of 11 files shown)


v2/apiserver/internal/core/kubernetes/substrate.go
```diff
@@ -14,11 +14,42 @@ import (
...
+
+var jobPodsSelector = labels.Set(
+	map[string]string{
+		myk8s.LabelComponent: "job",
+	},
+).AsSelector().String()
+
+var runningPodsSelector = fields.Set(
+	map[string]string{
+		"status.phase": string(corev1.PodRunning),
```
radu-matei, line 37: Is there a typed object for `status.phase`?

v2/apiserver/internal/core/kubernetes/substrate.go
```diff
@@ -52,6 +83,84 @@ func NewSubstrate(
...
+	count := core.SubstrateJobCount{}
+	phaseSelectors := []string{
+		pendingPodsSelector,
+		runningPodsSelector,
+		unknownPhasePodsSelector,
+	}
+	// Use a set to assist in counting. This helps prevent us from double-counting
+	// a pod if its phase changes from pending to running while we're counting.
+	podsSet := map[string]struct{}{}
+	for _, phaseSelector := range phaseSelectors {
```
radu-matei, line 137: As far as I can tell, the two functions here share almost all the logic for actually counting pods - the only difference seems to be the pod selector passed to the pod client.

Would it make sense to avoid duplicating?

radu-matei (approved):
LGTM, with two minor comments.

v2/apiserver/internal/core/kubernetes/substrate.go
```diff
@@ -52,6 +83,84 @@ func NewSubstrate(
...
+) (core.SubstrateWorkerCount, error) {
+	count := core.SubstrateWorkerCount{}
+	phaseSelectors := []string{
+		pendingPodsSelector,
+		runningPodsSelector,
+		unknownPhasePodsSelector,
+	}
+	// Use a set to assist in counting. This helps prevent us from double-counting
+	// a pod if its phase changes from pending to running while we're counting.
+	podsSet := map[string]struct{}{}
```
radu-matei, line 97: Oh yes, the famous Go set..
(definitely ignore this comment)

v2/apiserver/internal/core/kubernetes/substrate.go
krancour: Good question. I had hoped so, but couldn't find one. A field selector allows arbitrary navigation of nested fields, so I suppose it also makes sense that there wouldn't be.

v2/apiserver/internal/core/kubernetes/substrate.go
krancour: Yes. I definitely overlooked that. Fixing it now.

krancour: Fixed in recent force push.

PR #1169, opened 2020-11-19
allow apiserver to list pods

krancour:
This is a follow-up to #1162

This permission for the apiserver to list k8s pods. It breaks apart the the permissions for pods and PVCs since listing pods is necessary and listing PVCs is not. Since this grows the number of sections in this file, I also reordered them alphabetically by resource type so that it doesn't become unwieldy.

A side-by-side comparison of changes might be more helpful than a line-by-line comparison.


radu-matei (approved):
LGTM

PR #1172, opened 2020-11-23
add support for CORS

krancour:
Students at BYU are working on code that invokes the API from within a browser and alerted me to the fact that CORS wasn't currently supported by our API server.

This PR adds support for CORS using a popular library.


radu-matei (approved):
LGTM.
This is fine for now, but I think it might make sense to allow for configurable allowed origins at some point.

PR #1173, opened 2020-11-23
prerequisite api changes for the forthcoming observer component

krancour:
The observer will trigger cleanup of Kubernetes resources related to completed workers and jobs, however, it seems that in keeping with a theme, that work should be delegated to the API server instead of the observer doing it directly.

This PR adds relevant endpoints to both the API server and the SDK.

### DIFF
--- a/v2/apiserver/internal/core/jobs.go
+++ b/v2/apiserver/internal/core/jobs.go
@@ -160,6 +160,9 @@ type JobsService interface {
 		jobName string,
 		status JobStatus,
 	) error
+	// Cleanup removes Job-related resources from the substrate, presumably
+	// upon completion, without deleting the Job from the data store.
+	Cleanup(ctx context.Context, eventID, jobName string) error
 }
 
 type jobsService struct {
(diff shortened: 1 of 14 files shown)


v2/apiserver/internal/core/jobs_test.go
```diff
@@ -250,6 +250,113 @@ func TestJobsServiceUpdateStatus(t *testing.T) {
 	}
 }
 
+func TestJobsServiceCleanup(t *testing.T) {
```
vdice, line 253: Should we add a success case here? (or intended to be covered by the corresponding sdk test?)

v2/apiserver/internal/core/workers_test.go
```diff
@@ -205,6 +205,82 @@ func TestWorkersServiceUpdateStatus(t *testing.T) {
 	}
 }
 
+func TestWorkersServiceCleanup(t *testing.T) {
```
vdice, line 208: Should we add a success case here? (or intended to be covered by the corresponding sdk test?)

v2/apiserver/internal/core/jobs.go
```diff
@@ -262,6 +265,41 @@ func (j *jobsService) UpdateStatus(
...
+	jobName string,
+) error {
+	event, err := j.eventsStore.Get(ctx, eventID)
+	if err != nil {
+		return errors.Wrapf(err, "error retrieving event %q from store", eventID)
+	}
+	_, ok := event.Worker.Jobs[jobName]
+	if !ok {
+		return &meta.ErrNotFound{
+			Type: "Job",
```
vdice, line 280: Not sure if I've asked about his previously... do we have plans to create consts for supported types?  `const JobType string = "Job"`, etc.?

vdice (approved):

v2/apiserver/internal/core/jobs.go
krancour: We should probably do that.

v2/apiserver/internal/core/workers_test.go
krancour: Oversight. Nice catch. Fixing.

v2/apiserver/internal/core/jobs_test.go
krancour: Oversight. Nice catch. Fixing.

v2/apiserver/internal/core/jobs.go
krancour: Created #1174 to track that.

v2/apiserver/internal/core/workers_test.go
krancour: Fixed.

v2/apiserver/internal/core/jobs_test.go
krancour: Fixed.

v2/apiserver/internal/core/jobs.go
vdice: Thank you!

PR #1176, opened 2020-11-24
add worker component

krancour:
This needs more tests, but I am removing it from WIP/draft status and requesting review because:

1. A (working pun intended) worker component is increasingly required to see things working e2e and...

1. There's a parallel effort underway to create a Brigade 2 TS/JS SDK, including an API client. Anticipating that we would eventually transition this worker to using that SDK, most of the missing tests from here _actually_ belong in the TS/JS SDK.

Bottom line, I feel like there's more benefit to getting this in _sooner_ with less coverage than to getting it in later with more.

With this merged, Brigade 2 WORKS end-to-end-- not to say that there are no known (and tracked) issues-- but this would be a _very_ good place to end the year.


examples/pipeline-demo.yaml
```diff
@@ -9,18 +9,14 @@ spec:
       brigade.js: |-
         const { events, Job } = require("brigadier");
 
-        events.on("github.com/brigadecore/brigade/cli:exec", (e) => {
-          let fooJob = new Job("foo", "debian:latest");
+        events.on("github.com/brigadecore/brigade/cli", "exec", async event => {
```
vdice, line 12: Non/blocking; just an idea: Should we add a line where we demonstrate calling a job's `logs()` method?

(Ignore if we already do this in another example, etc.)

vdice (approved):
LGTM! 🎉

examples/pipeline-demo.yaml
krancour: The prototype had some such examples. I haven't ported over all examples yet.

It does work, but I should note that I'm questioning whether that's a good/practical thing for people to do.

Jobs produce logs. Those logs are already accessible using the API or CLI. I can't figure out why anyone would need to access them from inside the worker. If it's just to dump them, that's kind of pointless, because, as I said, the job logs are already accessible...

I think there's another reason people do this and I think it was an unfortunate 1.x workaround turned 2.x anti-pattern. I think some people use job logs as a form of _output_. So something the job wrote to its logs might be retrieved and used as input (or as a conditional) as the worker continues doing whatever it does.

That feels dirty to me, but I know a lot of people did it because they _couldn't_ do something better-- they couldn't have a job write to a file that was accessible to the worker because in v1.x, shared storage was only mounted to the jobs and _not_ mounted to the worker itself. But that's fixed in 2.0, so I'm questioning whether there's any reason at all to allow workers to access job logs.

That said... I am not necessarily intent on removing `job.logs()` right now, but I'm holding back on documenting / demonstrating it for now.

PR #1177, opened 2020-11-24
add log agent

krancour:
This PR builds adds Docker builds of custom fluentd-based log agents for both Linux and Windows.

Why custom? In the case of Linux it's only because we bake in the necessary fluentd plugins-- which is a common thing to do. In the case of Windows, it's that _plus_ the necessity of producing an image that's compatible with Windows Server 2019. (Weirdly, fluentd's Windows image seems to be incompatible with that.)

The build of the Linux-based log agent is not much different from any of the other Docker image builds in this project.

For Windows, however, we have to resort to some shenanigans to get the image built-- we have to fall back on GitHub Actions  to accomplish it (instead of dogfooding Brigade) because "Docker-in-Docker" is not well documented for Windows and my efforts to make it work have been fruitless.

The Windows container build also take a _very long time_. And since most PRs won't touch it, I propose we _not_ configure the v2 branch protections in GitHub to require that particular check pass, and that we not allow pending or in-progress Windows log agent builds to block PRs that don't touch the logger.

### DIFF
--- a/.github/workflows/merge.yaml
+++ b/.github/workflows/merge.yaml
@@ -1,0 +1,17 @@
+name: Brigade v2 Merge
+
+on:
+  push:
+    branches:
+      - v2
+
+jobs:
+  build:
+    runs-on: windows-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build Windows logger agent
+        run: |
+          docker build -f v2/logger/Dockerfile.winserv-2019 v2/logger
--- a/.github/workflows/merge.yaml
+++ b/.github/workflows/merge.yaml
@@ -12,6 +12,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      # TODO: We should also publish the image
       - name: Build Windows logger agent
         run: |
           docker build -f v2/logger/Dockerfile.winserv-2019 v2/logger
--- a/charts/brigade/values.yaml
+++ b/charts/brigade/values.yaml
@@ -182,7 +182,7 @@ logger:
   linux:
 
     image:
-      repository: krancour/brigade-logger-linux
+      repository: brigadecore/brigade-logger-linux
       ## tag should only be specified if you want to override Chart.appVersion
       ## The default tag is the value of .Chart.AppVersion
       # tag:
@@ -209,7 +209,7 @@ logger:
   windows:
 
     image:
-      repository: krancour/brigade-logger-windows
+      repository: brigadecore/brigade-logger-windows
       ## tag should only be specified if you want to override Chart.appVersion
       ## The default tag is the value of .Chart.AppVersion
       # tag:
--- a/charts/brigade/values.yaml
+++ b/charts/brigade/values.yaml
@@ -177,6 +177,66 @@ worker:
 
   workspaceStorageClass: nfs
 
+logger:
+
+  linux:
+
+    image:
+      repository: krancour/brigade-logger-linux
+      ## tag should only be specified if you want to override Chart.appVersion
+      ## The default tag is the value of .Chart.AppVersion
+      # tag:
+      pullPolicy: IfNotPresent
+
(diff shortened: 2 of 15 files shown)


.github/workflows/merge.yaml
```diff
@@ -0,0 +1,17 @@
+name: Brigade v2 Merge
```
vdice, line 1: I'm wondering if we should add a comment to these new GitHub Action yaml files mentioning they are currently used for building Windows images and not yet intended for full CI (where it's the brigade.js that does the heavy lifting).  WDYT?

charts/brigade/values.yaml
```diff
@@ -177,6 +177,66 @@ worker:
 
   workspaceStorageClass: nfs
 
+logger:
+
+  linux:
+
+    image:
+      repository: krancour/brigade-logger-linux
```
vdice, line 185: This triggered the thought: We can/should add logger image publishing to brigade.js (for linux) and GH actions (for windows) to Brigadecore's `brigadecore` DH org (and/or whichever registry we finalize on for v2).  Maybe a ticket to add/update CI (and update these references), if not filed already?

.github/workflows/merge.yaml
vdice: Also maybe touching on some of the points mentioned in the PR description, namely:

>The Windows container build also take a very long time. And since most PRs won't touch it, I propose we not configure the v2 branch protections in GitHub to require that particular check pass, and that we not allow pending or in-progress Windows log agent builds to block PRs that don't touch the logger.


Could be good to preserve those in source...

vdice (commented):
Code LGTM.  Can/should we help verify logging functionality from this PR -- or is full functionality forthcoming along w/ other components?

.github/workflows/merge.yaml
krancour: Agreed. I will add a commit that adds a little bit more doc/context around why/how we're using GH Actions.

charts/brigade/values.yaml
krancour: This is actually a mistake-- that shouldn't say `krancour`, but `brigadecore`. I will fix that.

But stepping back, you're correct in observing that we're currently not publishing _any_ of the images we build. For the time being, with so few people actively working on this, building and deploying from source (see `make hack`) seems preferable to cluttering up whatever image repository we think this belongs in long-term.

I'd been thinking that we'd re-work our pipelines to publish "edge" images on merge to `v2` and semantically tagged images on release shortly before actually cutting the alpha release. Is that reasonable?

I will update that errant reference to `krancour`.

krancour:
> Can/should we help verify logging functionality from this PR -- or is full functionality forthcoming along w/ other components?

This PR only adds log aggregation. I'll be opening another in just a few minutes that will add API endpoints and corresponding CLI commands for streaming those logs... ymmv in trying it all end-to-end, however, since the new worker isn't merged yet.

(I've been validating a lot of this locally using a Franken-branch that has one unmerged PR piled on top of another.)

krancour:
@vdice at your convenience, please lmk if https://github.com/brigadecore/brigade/pull/1177/commits/fa8202fd40682f93ccf95e69fcca148dfb2104a3 addresses your feedback.

charts/brigade/values.yaml
vdice: > building and deploying from source (see make hack) seems preferable 

Indeed, during this period, this seems preferred.  Better to focus on v2 bread and butter.

> I'd been thinking that we'd re-work our pipelines to publish "edge" images on merge to v2 and semantically tagged images on release shortly before actually cutting the alpha release. Is that reasonable?

Sgtm, thanks!

vdice:
The added docs and updates in the latest commit are 💯 -- thanks @krancour !

vdice (approved):

PR #1179, opened 2020-11-25
transition from more complex docker-in-docker to less complex kaniko

krancour:
This PR transitions the official builds (those used in CI and, eventually, the release pipeline) to Kaniko. The main benefit to this is it ends our dependency on the comparatively complex and less secure "Docker in Docker" pattern (less secure because DinD requires a privileged container).

A secondary benefit is we get experience with a different build pattern that I think we might want to add to our playbook. (i.e. We should probably present this as an alternative to DinD in Brigade's own documentation whenever we get around to writing 2.0 docs.)

The custom Kaniko image this uses is in my own Docker Hub repository for now, with source in my own GitHub repository, but as with other "utility" images we use, we can more that over to the `brigadecore` organization in the near future.

Builds and release of my custom Kaniko image _are_ automated.

Note also that the `Makefile` got a bit of a makeover (pun intended). All "hack"-oriented targets (see bottom of the Makefile) shun Kaniko and use Docker to build images. Why? On a developer's device wherein Docker runs locally or in a VM (like Docker Desktop), DinD doesn't come into play, so using Docker instead of Kaniko speeds up development since it will allow developers to take advantage of the local image cache.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,18 @@ ifneq ($(SKIP_DOCKER),true)
 		-v $(PROJECT_ROOT):/workspaces/brigade \
 		-w /workspaces/brigade \
 		$(GO_DEV_IMAGE)
+
+	KANIKO_IMAGE := krancour/kaniko:v0.2.0
+
+	KANIKO_DOCKER_CMD := docker run \
+		-it \
+		--rm \
+		-e SKIP_DOCKER=true \
+		-e DOCKER_USERNAME=$${DOCKER_USERNAME} \
+		-e DOCKER_PASSWORD=$${DOCKER_PASSWORD} \
+		-v $(PROJECT_ROOT):/workspaces/brigade \
+		-w /workspaces/brigade \
+		$(KANIKO_IMAGE)
 endif
 
 # Allow for users to supply a different helm cli name,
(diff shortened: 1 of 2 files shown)


scripts/docker-login.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+auth=$(echo -n "${DOCKER_USERNAME}:${DOCKER_PASSWORD}" | base64)
+
+cat << EOF > /kaniko/.docker/config.json
+{
+  "auths": {
+    "https://index.docker.io/v1/": {
```
vdice, line 10: We'll probably want/need to parameterize this registry as well, right?  (Perhaps can use DOCKER_IMAGE_PREFIX, if defined; otherwise default to the value currently used?)

vdice (commented):
Neat stuff!  Kaniko indeed looks very handy (and more secure, as mentioned).

I did have the following thought as I tried this branch out:  Have we considered flipping the targets/approach such that the main build/publish targets are suited towards devs (using classic `docker build/run`, etc.) but then having separate `kaniko-*` targets for the others, which Brigade/GH actions/etc. would use?

I'm thinking they maybe shouldn't be the defaults since, 1. it sounds like builds often take longer and/or are less efficient (no image cache?) and 2. the login/auth flow for kaniko uses a script which may need further refinements to support different registries that a dev might use (and already be auth'd via docker's config, etc.)

scripts/docker-login.sh
krancour: I had been thinking we could just update this script if we want to push to somewhere else, as I don't anticipate that being something that changes often. But I'm also open to parameterizing it.

krancour: Thinking more... I suppose if one were to fork us, they might conceivably want to push the images they build from their fork somewhere else. I suppose this should indeed be parameterized.

krancour:
@vdice 

> I did have the following thought as I tried this branch out: Have we considered flipping the targets/approach such that the main build/publish targets are suited towards devs (using classic docker build/run, etc.) but then having separate kaniko-* targets for the others, which Brigade/GH actions/etc. would use?
>
> I'm thinking they maybe shouldn't be the defaults since, 1. it sounds like builds often take longer and/or are less efficient (no image cache?)

I see your point. If someone wants to build from source and hasn't read some new and improved `DEVELOPING.md` (that we haven't even written yet) they are probably going to assume that `make build` or `make push` do the trick. They're _not_ going to run the "special" `hack-` targets.

As a counterpoint, however, I would argue that as things currently stand in this branch, their assumption would actually be _correct_. `make build` and `make push` produce and push images using the same processes used by our CI (and eventually CD) pipelines. If someone wants to build from source, the output of those targets is arguably "official." imho, that's probably what most people would want when they find themselves building from source.

(For the sake of further argument, if any functional difference were ever to be discovered between images produced by Docker vs by Kaniko, we'd want to know people doing custom builds are using the same processes we're actually using in CD.)

From that perspective, the targets prefixed with `hack-` are unofficial optimizations for the few contributors who may find themselves building and deploying (typically locally) dozens of times a day and are willing to make a tradeoff that favors faster iterations over more official builds.

> and 2. the login/auth flow for kaniko uses a script which may need further refinements to support different registries that a dev might use (and already be auth'd via docker's config, etc.)

We should definitely solve for that.

krancour:
@vdice lmk if you think https://github.com/brigadecore/brigade/pull/1179/commits/ec22a26e5a966fd5255ce26c1fd6354efe19a213 improves things.

krancour:
GH is being weird... my last few comments were inline on commit ec22a26, but for some reason, they're not showing with code context here in the conversation tab. Be sure to look at ec22a26 directly.

Makefile
```diff
@@ -95,64 +107,84 @@ test-unit-go:
...
-		--build-arg COMMIT='$(GIT_VERSION)' \
-		.
-	docker tag $(DOCKER_IMAGE_PREFIX)brigade-$*:$(IMMUTABLE_DOCKER_TAG) $(DOCKER_IMAGE_PREFIX)brigade-$*:$(MUTABLE_DOCKER_TAG)
+	$(KANIKO_DOCKER_CMD) /kaniko/executor \
+		--dockerfile /workspaces/brigade/v2/$*/Dockerfile \
+		--context dir:///workspaces/brigade/ \
+		--no-push
+
+.PHONY: push-logger-linux
+push-logger-linux:
```
vdice, line 125: These push targets appear to re-build first.  Is that as intended (and/or how kaniko works)?

Makefile
```diff
@@ -95,64 +107,84 @@ test-unit-go:
...
 .PHONY: build-%
 build-%:
-	docker build \
-		-f v2/$*/Dockerfile \
-		-t $(DOCKER_IMAGE_PREFIX)brigade-$*:$(IMMUTABLE_DOCKER_TAG) \
-		--build-arg VERSION='$(VERSION)' \
-		--build-arg COMMIT='$(GIT_VERSION)' \
-		.
-	docker tag $(DOCKER_IMAGE_PREFIX)brigade-$*:$(IMMUTABLE_DOCKER_TAG) $(DOCKER_IMAGE_PREFIX)brigade-$*:$(MUTABLE_DOCKER_TAG)
+	$(KANIKO_DOCKER_CMD) /kaniko/executor \
```
vdice, line 119: Was there a reason the meta targets of `build` and `build-images` were removed?

vdice (commented):
I can see both points of view re: which direction to default the build/push targets.  Perhaps, as there's no rush to get this in, we wait a bit for @radu-matei and/or @technosophos to chime in?

Makefile
krancour: It's just how it works. Container comes up builds + pushes (unless you tell it not to push) and then goes down. When it goes down, the image you built within it is gone. So with the images we build being so ephemeral, build + push necessarily happen in one fell swoop.

Makefile
krancour: I found that they were never being used. As I hack away I use the `hack-` targets. Our CI/CD pipelines also never called those because they're very parallelized and thus use the more finely-grained targets.

But now that you've questioned it, perhaps these should remain for the benefit of anyone who just wants to manually kick off a build of the whole shootin' match.

I'm going to put them back.

krancour: I put 'em back.

technosophos (approved):
I think this is a fine change to make. I was never thrilled with DinD, but it was the best we had at the time.

Makefile
```diff
@@ -28,6 +28,18 @@ ifneq ($(SKIP_DOCKER),true)
 		-v $(PROJECT_ROOT):/workspaces/brigade \
 		-w /workspaces/brigade \
 		$(GO_DEV_IMAGE)
+
+	KANIKO_IMAGE := krancour/kaniko:v0.2.0
```
radu-matei, line 32: This should probably be pushed to `brigadecore`.

radu-matei (approved):
I think I agree with @krancour on having the outcome of `make build` be the same as CI.

LGTM

Makefile
krancour: I think I made a note of that already. I'm going to open an issue about that.

krancour: https://github.com/brigadecore/brigade/issues/1192

PR #1186, opened 2020-12-07
add support for new worker/job STARTING phase

### LINKED ISSUE #1185: Add a STARTING phase to both workers and jobs
Currently, every worker or job remains in a PENDING state until the observer observes the worker or job running on the substrate. This being the case, a worker or job that's PENDING could _either_ be 1. in a queue waiting to be scheduled _or_ 2. scheduled in k8s, but not running yet.

If we add a STARTING phase that workers and jobs enter when the scheduler uses the API to launch a pending worker or job on the substrate, we would be able to differentiate between workers and jobs that are simply waiting their turn and those who are currently _getting_ their turn, but haven't been observed as running yet.

The ability to differentiate between PENDING and STARTING may eventually prove useful. Supposing a worker or job were ever to get "stuck," knowing whether it's PENDING or STARTING would tell us more about _where_ it got stuck.

krancour:
Fixes #1185

### DIFF
--- a/v2/apiserver/internal/core/mongodb/events_store.go
+++ b/v2/apiserver/internal/core/mongodb/events_store.go
@@ -170,8 +170,13 @@ func (e *eventsStore) Cancel(ctx context.Context, id string) error {
 	res, err = e.collection.UpdateOne(
 		ctx,
 		bson.M{
-			"id":                  id,
-			"worker.status.phase": core.WorkerPhaseRunning,
+			"id": id,
+			"worker.status.phase": bson.M{
+				"$in": []core.WorkerPhase{
+					core.WorkerPhaseStarting,
+					core.WorkerPhaseRunning,
+				},
+			},
 		},
 		bson.M{
 			"$set": bson.M{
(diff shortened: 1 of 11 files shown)


v2/cli/event_commands.go
```diff
@@ -63,6 +63,12 @@ var eventCommand = &cli.Command{
 					Usage: "If set, will additionally abort and cancel events with " +
 						"their worker in a RUNNING phase",
 				},
+				&cli.BoolFlag{
+					Name:    flagStarting,
+					Aliases: []string{"r"},
```
vdice, line 68: Is this alias intended?  Noticing it isn't added to the other instances of this new flag.

v2/apiserver/internal/core/mongodb/events_store.go
```diff
@@ -184,6 +189,9 @@ func (e *eventsStore) Cancel(ctx context.Context, id string) error {
 		return errors.Wrapf(err, "error updating status of event %q worker", id)
 	}
 
+	// TODO: If the event was running when it was canceled, its jobs need to be
```
vdice, line 192: Do we need to create a ticket to track this TODO (and the add'l one further down)?

vdice (approved):
LGTM

v2/cli/event_commands.go
krancour: That's a copy/paste error. Good eye. Fixing.

krancour: Fixed in https://github.com/brigadecore/brigade/pull/1186/commits/65c74ab7cf9ac34afbf67d5871882751aefb9cae

v2/apiserver/internal/core/mongodb/events_store.go
krancour: I opened https://github.com/brigadecore/brigade/issues/1191

PR #1190, opened 2020-12-08
v2 compatible Brigadier library

krancour:
This is a new, v2-compatible Brigadier library.

I'll preemptively defend the choice to collapse brigadier into the main brigade repo. It's _not_ a philosophical preference for monorepos, or anything of the sort. Since the (forthcoming) worker component is itself _based_ on Brigadier, it seems most straightforward to keep both in the same repository so that any single logical change to the worker that requires a supporting change in Brigadier does not need to be coordinated across two repositories. It has the added benefit of putting Brigadier and Brigade on a single release train.

This is the same logic that was applied to the decision to keep the Go SDK in the same repo as the rest of Brigade, since non-API components like the scheduler and observer rely on it.

There shouldn't be much difficulty publishing Brigadier patches to the 1.x line from the existing brigadecore/brigadier repository while publishing 2.x releases from this one. We can also add a note to the brigadecore/brigadier `README.md` to indicate that repository contains source only for Brigadier 1.x and that 2.x code is kept with the rest of Brigade.


vdice (commented):
I don't have a strong opinion here, but would like to make sure @radu-matei and/or @technosophos have a chance to chime in.

An idea if we continue in this direction: Would it be possible for our own use of brigadier in CI (brigade.js) to reference the same brigadier code now co-located here?  That could be one big benefit of co-location: when brigadier code is changed, CI on the PR branch would run with these changes as well.

Also, a question: Were there any brigadier typescript file updates/mods when porting over (for v2 compatibility) to be aware of?  Or just directly copied over?

technosophos:
It's fine with me.

radu-matei:
I agree with @vdice's suggestion - it does mean that we would have to check in transpiled JavaScript code, and update it with every iteration of Brigadier (this is an example of doing so with `brigade-utils` - https://github.com/brigadecore/brigade-utils/blob/master/brigade.js)

v2/brigadier/src/groups.ts
```diff
@@ -0,0 +1,45 @@
...
+    const group = new Group(jobs)
+    return group.runAll()
+  }
+
+  public static async runEach(jobs: Job[]): Promise<void> {
+    const group = new Group(jobs)
+    return group.runEach()
+  }
+
+  private jobs: Job[] = []
```
radu-matei, line 15: Is this necessary if you initialize the jobs in the constructor?

krancour: Nope. Not necessary. I'll remove that.

krancour: Fixed in 72fbccf0eddfea14de7e8e0054d9da1a3aba31b1

krancour:
@vdice 

> Were there any brigadier typescript file updates/mods when porting over (for v2 compatibility) to be aware of? Or just directly copied over?

This is similar to Brigadier v1.x-- enough so that it should feel very familiar, but the model is different (Brigade 2.0 events, projects, etc. look a little different than their 1.x counterparts). Another slight change is that the `Job` constructor requires an `Event` as input now-- which makes total sense if you think about it. Jobs really don't _exist_ except in relation to an event. The technical reason for that change might not be immediately evident, but if you were to compare a 2.0 worker (forthcoming) based on this Brigadier to a v1.x worker based on the v1.x Brigadier, you'd see this change helped to avoid some icky use of global variable `currentEvent`.

Consider all of this to be new, but familiar code.

krancour:
> Would it be possible for our own use of brigadier in CI (brigade.js) to reference the same brigadier code now co-located here?

We can cross that bridge in another PR, but I'm not entirely in favor of it. I am pro-dogfooding, but I think if we involve code that should be _under test_ in executing those tests, it introduces too much volatility.

But either way, we don't need to decide that now.

vdice (approved):

PR #1193, opened 2020-12-11
misc. brigadier improvements

krancour:
1. Add a logger that is exposed to `brigade.js` authors.

    In 1.x, the only logger we had was entirely internal to the worker. We used it, but `brigade.js` authors could not. So although one could set worker log level when creating a project, it only ever controlled whatever logging the worker did itself behind the scenes. There was no access to a leveled logger within the script author's own code. This PR fixes that. The logger is based on the popular [winston](https://www.npmjs.com/package/winston) library.

1. Correct an oversight in the first commit of brigadier bits by adding a `logLevel` field to the `Worker` class.

1. Organize and cleanup imports.

    TIL that if one module is aggregating exports from other modules, this:

    ```
    import { Foo } from "foo"
    export { Foo }
    ```

    Is equivalent to this, which is cleaner syntax:

    ```
    export { Foo } from "foo"
    ```

1. Minor test code cleanup

1. Test the module's exports

### DIFF
--- a/v2/brigadier/package.json
+++ b/v2/brigadier/package.json
@@ -34,5 +34,8 @@
     "mocha": "^8.2.1",
     "ts-node": "^9.1.1",
     "typescript": "^4.1.2"
+  },
+  "dependencies": {
+    "winston": "^3.3.3"
   }
 }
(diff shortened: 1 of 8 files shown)


vdice (approved):
Nice!

PR #1201, opened 2020-12-14
use new, official dev images

### LINKED ISSUE #1192: move dev / utility images and their source into brigadecore org
There are, as of this writing, two custom images we use to facilitate v2 development-- `krancour/go-tools` and `krancour/kaniko`.

The images should be moved to the brigadecore Docker Hub org and the corresponding source should be moved to the brigadecore GitHub org.

krancour:
Fixes #1192

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,9 +16,7 @@ GIT_VERSION = $(shell git describe --always --abbrev=7 --dirty --match=NeVeRmAtC
 
 ifneq ($(SKIP_DOCKER),true)
 	PROJECT_ROOT := $(dir $(realpath $(firstword $(MAKEFILE_LIST))))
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -141,14 +139,14 @@ build-images: build-apiserver build-scheduler build-observer build-logger-linux
 
 .PHONY: build-logger-linux
 build-logger-linux:
-	$(KANIKO_DOCKER_CMD) /kaniko/executor \
+	$(KANIKO_DOCKER_CMD) kaniko \
```
krancour, line 142: New image puts kaniko on the path.

vdice (approved):
🎉

PR #1206, opened 2020-12-22
secure job environment storage

### LINKED ISSUE #1200: Job container env vars shouldn't be stored in the db
Currently, we're storing a job's environment variables in the database at the time the job is created. Since we assume there may be sensitive data contained in a job's environment variables, this means potential secrets can be sitting, unencrypted, in the database, which is obviously very bad.

We already do create a k8s secret to store these environment variables, but we do it _later_-- immediately before the job is started.

If we move creation of this same k8s secret _forward_ to the time the job is created, then the environment variables do not need to be persisted in the data base.

krancour:
Fixes #1200

### DIFF
--- a/v2/apiserver/internal/core/jobs_test.go
+++ b/v2/apiserver/internal/core/jobs_test.go
@@ -211,6 +211,52 @@ func TestJobsServiceCreate(t *testing.T) {
 			},
 		},
 		{
+			name: "error storing job environment in substrate",
+			service: &jobsService{
+				eventsStore: &mockEventsStore{
+					GetFn: func(context.Context, string) (Event, error) {
+						return Event{
+							Worker: Worker{
+								Spec: WorkerSpec{
+									UseWorkspace: true,
+									JobPolicies: &JobPolicies{
+										AllowPrivileged:        true,
+										AllowDockerSocketMount: true,
+									},
+								},
+							},
+						}, nil
+					},
+				},
+				projectsStore: &mockProjectsStore{
+					GetFn: func(context.Context, string) (Project, error) {
+						return Project{}, nil
+					},
+				},
+				jobsStore: &mockJobsStore{
+					CreateFn: func(context.Context, string, string, Job) error {
+						return nil
+					},
+				},
+				substrate: &mockSubstrate{
+					StoreJobEnvironmentFn: func(
+						context.Context,
+						Project,
+						string,
+						string,
+						JobSpec,
+					) error {
+						return errors.New("something went wrong")
+					},
+				},
+			},
+			assertions: func(err error) {
+				require.Error(t, err)
+				require.Contains(t, err.Error(), "something went wrong")
+				require.Contains(t, err.Error(), "error storing event")
+			},
+		},
+		{
 			name: "error scheduling job on substrate",
 			service: &jobsService{
 				eventsStore: &mockEventsStore{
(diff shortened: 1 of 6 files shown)


v2/apiserver/internal/core/jobs_test.go
```diff
@@ -210,6 +210,52 @@ func TestJobsServiceCreate(t *testing.T) {
 				require.Contains(t, err.Error(), "error saving event")
 			},
 		},
+		{
+			name: "error storing job environment in substrate",
+			service: &jobsService{
+				eventsStore: &mockEventsStore{
+					GetFn: func(context.Context, string) (Event, error) {
+						return Event{
```
vdice, line 218: nit: wondering if this extra worker spec config is relevant to the tests at hand?   If not, perhaps we remove in the name of brevity?

v2/apiserver/internal/core/kubernetes/substrate_test.go
```diff
@@ -788,6 +788,65 @@ func TestSubstrateStartWorker(t *testing.T) {
 	}
 }
 
+func TestSubstrateStoreJobEnvironment(t *testing.T) {
```
vdice, line 791: Wondering where we could/should put a test vetting the redaction of the environment for a job (and any sidecar containers)?  Although the logic hits the jobs store as well as the substrate, would it be possible to add unit test coverage using mocks?

vdice (commented):
Looking good; couple of Qs.

v2/apiserver/internal/core/kubernetes/substrate_test.go
krancour: Good question. If the tests don't cover it already, it's an oversight. I'll fix that.

v2/apiserver/internal/core/jobs_test.go
krancour: At a glance, it doesn't look like it needs to be there. I am scratching my head as to why I ever had it there to begin with. I'll shorten this as suggested.

radu-matei (approved):
Once @vdice's comments are addressed, LGTM.

v2/apiserver/internal/core/jobs_test.go
krancour: Didn't hit me until I went to change it, but there _is_ a reason this was there.

Starting on line 363, this hardcoded `Job` is the used for _all_ the test cases:

```
Job{
	Spec: JobSpec{
		PrimaryContainer: JobContainerSpec{
			Privileged:          true,
			UseHostDockerSocket: true,
			WorkspaceMountPath:  "/var/workspace",
		},
	},
}
```

This job has all the bells and whistles so that various test cases have the opportunity to trip over "the Worker's spec doesn't allow jobs to do x."

So the Worker spec that you were thinking may not be needed here _is_ needed, because it allows the code under test to clear all of those hurdles and reach the point in the code where redaction actually happens.

If we wanted to get by with a more minimal Event for this test case (or others), we'd have to also define a Job for each _individual_ test case instead of always using one that has all the bells and whistles. That would work, but my gut says in the end it's a similar amount of code, but more complexity.

v2/apiserver/internal/core/kubernetes/substrate_test.go
krancour: It shouldn't be in this test suite. It should be in the tests for the JobsService. We can add assertions to the mock substrate to verify that redaction had taken place as expected on the Job that the JobsService passed to it.

Working on it.

krancour: Fixed in c96370e090be8986e782f216eeda73c5b8709efc

v2/apiserver/internal/core/jobs_test.go
vdice: Ah, ok, thanks for the explanation.  Current approach sgtm!

vdice (approved):

PR #1214, opened 2021-01-05
feat(secrets_commands.go): add secret command alias

vdice:
I found myself typing `brig project secret set ...` often, especially as the current mode of this command is to set one secret at a time.  Therefore, I thought this might be an appropriate alias to add.

### DIFF
--- a/v2/cli/secrets_commands.go
+++ b/v2/cli/secrets_commands.go
@@ -16,8 +16,9 @@ import (
 )
 
 var secretsCommand = &cli.Command{
-	Name:  "secrets",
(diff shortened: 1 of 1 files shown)


krancour (approved):
LGTM. Note you can use the `--set` flag multiple times, which is why this was plural to begin with.

I may actually be in favor of adding aliases for all resource-oriented commands so that the CLI is not at all picky about singular/plural. I'd want us to be consistent if we do that-- official commands singular across the board and aliases plural across the board. We can tackle that in a follow up. This still LGTM for now.

PR #1215, opened 2021-01-05
fix(project.json): add commit and ref fields to gitConfig

vdice:
Adds missing `commit` and `ref` fields to the `workerSpec`'s `git` section of the project config, to match the similar config on an event.

Actually, since the two `gitConfig` definitions are now equivalent between the project and event schema, we could entertain housing common schema definitions in a sibling file that both could then pull from... thoughts?

### DIFF
--- a/v2/apiserver/schemas/event.json
+++ b/v2/apiserver/schemas/event.json
@@ -62,6 +62,9 @@
 				},
 				"commit": {
 					"type": "string",
+					"pattern": "^[a-f0-9]*$",
+					"minLength": 8,
+					"maxLength": 40,
 					"description": "A git commit sha"
 				},
 				"ref": {
--- a/v2/apiserver/schemas/event.json
+++ b/v2/apiserver/schemas/event.json
@@ -62,7 +62,7 @@
 				},
 				"commit": {
 					"type": "string",
-					"pattern": "^[a-f0-9]*$",
+					"pattern": "^[a-fA-F0-9]*$",
 					"minLength": 8,
 					"maxLength": 40,
 					"description": "A git commit sha"
--- a/v2/apiserver/schemas/project.json
+++ b/v2/apiserver/schemas/project.json
@@ -226,7 +226,7 @@
         },
 				"commit": {
 					"type": "string",
-					"pattern": "^[a-f0-9]*$",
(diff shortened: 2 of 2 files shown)


krancour:
> Actually, since the two gitConfig definitions are now equivalent between the project and event schema, we could entertain housing common schema definitions in a sibling file that both could then pull from... thoughts?

Seems very reasonable. There are probably a lot of similar opportunities to clean things up. The only reason it wasn't like that to begin with is because I didn't start out feeling particularly confident in my JSON schema skills.

v2/apiserver/schemas/project.json
```diff
@@ -223,6 +223,14 @@
 						}
 					],
 					"description": "The URL for cloning a git project"
+        },
+				"commit": {
+					"type": "string",
+					"description": "A git commit sha"
```
krancour, line 229: Can we restrict this field to hexadecimal characters only?

`event.json` probably deserves the same change... and as you noted, common types can probably be consolidated into some kind of common / library file.

vdice: Updated.  Could you check my work on the pattern?  Also, I added a min and max length according to git short/full sha conventions -- overkill?

krancour: I think that's reasonable.

v2/apiserver/schemas/event.json
```diff
@@ -62,6 +62,9 @@
 				},
 				"commit": {
 					"type": "string",
+					"pattern": "^[a-f0-9]*$",
```
krancour, line 65: Are git shas _always_ lowercase?

I sincerely do not know the answer to this.

vdice: All signs point to yes -- however, I just played around with taking a legit commit sha, updating many of the alphabetical characters to uppercase, and git had no qualms checking out the commit.  With that in mind, I'd be comfortable allowing `A-F` as well.  What do you think?

krancour: By Postel's principle... yes... let's accept both.

vdice: Updated.

krancour (approved):
LGTM

PR #1217, opened 2021-01-07
Add support for server sent event streams

### LINKED ISSUE #1212: Server side CORS policy needs to be more relaxed
It should allow requests from all origins and needs to explicitly indicate that cross origin requests are permitted to include the auth header.

krancour:
I was recently tooling around with creating a TS/JS Brigade 2 API client that works in both Node and any modern browser. In doing so, I stumbled across the fact that no analog to Node's `http2` library exists in the browser world. (There is no polyfill for it either.) This is not to say that browsers can't do http/2-- that would be absurd. But they _don't_ (yet?) expose any http/2 client that can be used to _programmatically_ consume a stream. This means any hypothetical TS/JS Brigade 2 API client used _within a browser_ cannot consume our log streams and status streams as they were currently implemented on the backend.

However, I learned that there's an entire [specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events) for _server sent events_. Modern browsers have decent support for it. Node does not, but the capability can be polyfilled into Node. Server sent events are a quick and easy way to implement a _one way_ stream (not full duplex like http/2) from server to client. _They are ideal for our purposes._ (And of course, they do still work over http/2 if that's how the client and server happen to communicate under the hood, but they don't _require_ it.)

Initially, I tried to transition our status and log streams to using SSE _exclusively_, but found that there were only a couple Go clients available that I could use in the Go SDK for Brigade 2. One was riddled with bugs. The other did not faithfully implement the specification. It seemed best for Go-based clients to stick to the original approach...

In the end, this PR preserves existing ability to do http/2 streams, but adds the _option_ for clients (like a hypothetical TS/JS client) to _request_ a stream of SSEs.

One nice consequence to this is that worker code can be dramatically simplified (also included in this PR) by using SSE instead of the `http2` library, since the `EventSource` API (a polyfill) automatically handles reconnects and stuff-- because that's actually part of the SSE spec.

Out of necessity, also fixes #1212

Unit tests for the new SSE stream options are lacking, because as noted, there isn't a good way to consume those streams using Go. All of this works, e2e. Since workers depend on this, when we get around to creating an e2e test suite, it will implicitly cover this new functionality.

### DIFF
--- a/v2/apiserver/internal/core/rest/logs_endpoints.go
+++ b/v2/apiserver/internal/core/rest/logs_endpoints.go
@@ -69,7 +69,7 @@ func (l *LogsEndpoints) stream(
 			if err != nil {
 				restmachinery.WriteAPIResponse(
 					w,
-					http.StatusNotFound,
+					http.StatusBadRequest,
 					&meta.ErrBadRequest{
 						Reason: "Value of \"Last-Event-ID\" header was not parseable as " +
 							"an int",
--- a/v2/apiserver/internal/core/rest/logs_endpoints.go
+++ b/v2/apiserver/internal/core/rest/logs_endpoints.go
@@ -39,12 +39,45 @@ func (l *LogsEndpoints) stream(
 	// nolint: errcheck
 	follow, _ := strconv.ParseBool(r.URL.Query().Get("follow"))
 
+	// Clients can request use of the SSE protocol instead of HTTP/2 streaming.
+	// Not every potential client language has equally good support for both of
+	// those, so allowing clients to pick is useful.
+	sse, _ := strconv.ParseBool(r.URL.Query().Get("sse")) // nolint: errcheck
+
 	selector := core.LogsSelector{
 		Job:       r.URL.Query().Get("job"),
 		Container: r.URL.Query().Get("container"),
 	}
 	opts := core.LogStreamOptions{
 		Follow: follow,
+	}
+
+	var lastEventID int64
+	// SSE has support for resuming where you left off after a
+	// disconnect/reconnect. Formally, I (krancour) believe the specification says
+	// clients can echo the ID of the last message received in a Last-Event-ID
+	// header, but I've encountered clients that use a lastEventId query parameter
+	// instead. Invoking Postel's principle, we'll support both.
+	if sse {
+		lastEventIDStr := r.Header.Get("Last-Event-ID")
+		if lastEventIDStr == "" {
+			lastEventIDStr = r.URL.Query().Get("lastEventId")
+		}
+		if lastEventIDStr != "" {
+			var err error
+			lastEventID, err = strconv.ParseInt(lastEventIDStr, 10, 64)
+			if err != nil {
+				restmachinery.WriteAPIResponse(
+					w,
+					http.StatusNotFound,
+					&meta.ErrBadRequest{
+						Reason: "Value of \"Last-Event-ID\" header was not parseable as " +
+							"an int",
+					},
+				)
+				return
+			}
+		}
 	}
 
 	logEntryCh, err := l.Service.Stream(r.Context(), id, selector, opts)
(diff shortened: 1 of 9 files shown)


v2/apiserver/internal/core/rest/logs_endpoints.go
```diff
@@ -47,6 +52,34 @@ func (l *LogsEndpoints) stream(
...
+		if lastEventIDStr == "" {
+			lastEventIDStr = r.URL.Query().Get("lastEventId")
+		}
+		if lastEventIDStr != "" {
+			var err error
+			lastEventID, err = strconv.ParseInt(lastEventIDStr, 10, 64)
+			if err != nil {
+				restmachinery.WriteAPIResponse(
+					w,
+					http.StatusNotFound,
```
krancour, line 72: This is the wrong status code. I'll fix this.

krancour: Fixed in 193960356a6f2fe251d8b4621998bf6fbc55fe58

v2/apiserver/internal/core/rest/logs_endpoints.go
```diff
@@ -47,6 +52,34 @@ func (l *LogsEndpoints) stream(
...
+		}
+		if lastEventIDStr != "" {
+			var err error
+			lastEventID, err = strconv.ParseInt(lastEventIDStr, 10, 64)
+			if err != nil {
+				restmachinery.WriteAPIResponse(
+					w,
+					http.StatusBadRequest,
+					&meta.ErrBadRequest{
+						Reason: "Value of \"Last-Event-ID\" header was not parseable as " +
```
vdice, line 74: Technically, the value could also come from the corresponding url query... not sure if we should reword to be more generic ("Value of the last event ID was not parseable...") or other.  In any case, perhaps not a big deal.

v2/worker/src/jobs.ts
```diff
@@ -55,122 +57,116 @@ export class Job extends BrigadierJob {
...
-      startMonitorReq() // This starts the monitor for the first time.
+        }
+      )
+      eventSource.addEventListener("message", (event: any) => { // eslint-disable-line @typescript-eslint/no-explicit-any
+        let status: any // eslint-disable-line @typescript-eslint/no-explicit-any
+        try {
+          status = JSON.parse(event.data)
+        } catch(e) {
+          eventSource.close() 
+          reject(new Error("Error parsing job status"))
```
vdice, line 79: Do we want to include/log `e` as well?  (Would inform a few other similar spots, if so).

vdice (commented):
Nice, looks good! A few questions...

v2/apiserver/internal/core/rest/logs_endpoints.go
krancour: Good thought. I think I wrote that error message _before_ becoming more forgiving about how last id was sent to us.

I'll reword as you suggest.

v2/worker/src/jobs.ts
krancour: Yes... we should probably be _wrapping_ the original exception.

I'll find the other spots where that might be applicable as well.

v2/worker/src/jobs.ts
```diff
@@ -45,13 +45,11 @@ export class Job extends BrigadierJob {
         },
       })
       if (response.status != 201) {
-        console.log(response.data)
-        throw new Error(response.data)
+        throw new Error(`Received ${response.status} from the API server`)
```
krancour, line 48: Throwing `response.data` was probably a very dumb thing for me to have done in the first place.

vdice: Oh yea - good catch.

v2/worker/src/jobs.ts
vdice: https://github.com/brigadecore/brigade/pull/1217/commits/298693b3d29f4d5477cd03fd089b8f178e92168e looks great - thanks!

vdice (approved):

Issue #1195, opened 2020-12-11
k8s resource name length limits


krancour:
Because we incorporate UUIDs into things like namespace names, secret names, and pod name, and because UUIDs are rather long, we might be flirting with bumping up against resource name length limits (if they exist) in cases where users choose long project names, job names, etc.

Research is required.

Once the limits are understood, we can adjust the apiserver's JSON schemas to enforce length limits on thing like project names, job names, etc. such that we avoid any issues.

krancour:
For reference: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

It would seem even with the current (generous) length limits we have placed on projects, job names, etc., we're not in any danger of running over the limit (253 characters) by appending those to a UUID.

krancour:
This does mean, we can possibly entertain larger length limits.

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1392

PR #1226, opened 2021-01-13
Add experimental support for brigade.ts

### LINKED ISSUE #527: typescript hook definition and event-time compilation capabilities
New Feature Request:
 As a brigade user
In order to reap the benefits of typescript in my hooks / pipelines definition files
I use brigade's typescript interfaces and event-time hook compilation capabilities

---

I use es6 imports to bring in interfaces needed to define various brigade event hook definitions, and then when events happen that otherwise depend on the plain javascript definition of the above, the brigade runtime typescript compiles my stuff.

Perhaps brigade exposes some type of lifecycle hook that could be used to do arbitrary preprocessing, which in turn could be used to typescript compile the hooks into the expected output location for brigade framework to then process.

krancour:
Fixes #527

Details on each commit:

### limit project names to 18 characters

In the course of tinkering with the rest of this PR, I created a project with a name longer than 18 characters. The length of the "brigade-" prefix combined with the project name and a UUID (v4) suffix created a namespace name that exceeded the maximum length of 63 characters.

See #1195. I do not believe such stingy limits are broadly applicable.

This could have been its own PR, but it's a small change.

### add node_modules to .dockerignore

It never hurts to slim down the Docker build context.

### add experimental typescript support

This works by enhancing the pre-processing done in the prestart script. The goal is to take a `brigade.ts` and compile it down to a `brigade.js` that the rest of the Worker can then deal with as per usual. Resolving dependencies is, of course, a prerequisite for compiling a TS script. At a minimum, a `brigade.ts` script is going to have a dependency on the `@brigadecore/brigadier` package. Getting dependency resolution to work more smoothly is actually the bulk of the code in this commit.

Previously, dependencies (if any) were enumerated in `brigade.json`. The prestart script would spawn a `yarn add` child process to install those packages by name.  This can be done more idiomatically by providing a `package.json` as input to a `yarn install` child process instead. So, out with the optional `brigade.json` and in with an optional `package.json`. This hopefully feels more familiar to developers.

The first thing the prestart script actually does is take any and all files that have been embedded in the project definition (which is common for a project with no VCS) and write them out to the file system as actual files. This is primarily because our `yarn install` child process requires an actual `package.json` file on the file system-- a "package.json" that's only a string in memory isn't of any use. A desirable side effect of writing all of the files embedded in the project out to the file system is that embedded scripts can seamlessly reference other embedded scripts and resources, just as if they were actual files (because they eventually will be). This makes it possible, for the first time, for developers of _non-VCS_ projects to break up an unwieldy `brigade.js` (or `brigade.ts`) into smaller, more organized scripts. Previously, this could only have been done in a VCS project.

Next, if no `package.json` exists on disk at this point, the prestart script creates a minimal one, so in the basic case of a project whose `brigade.js` (or `brigade.ts`) doesn't have any dependencies apart from `@brigadecore/brigadier`, providing a `package.json` is completely optional.

Next, the script _patches_ the `package.json` to amend it with (or override) _local_ (i.e. baked into the Worker image) coordinates for the `@brigadecore/brigadier` package. This ensures any `brigade.ts` is compiled against the _same_ version of `@brigadecore/brigadier` on which the Worker itself is based. This seems more sensible than to potentially compile against any random, possibly very old version of `@brigadecore/brigadier` that the project's _original_ `project.json` (if it had one) might have referenced. Note that any other dependencies enumerated in `package.json` remain unaltered.

Next, the script creates the `yarn install` child process and lets it do all of the real work.

Last, but only if applicable, the script spawns a `tsc` process to compile `brigade.ts` to `brigade.js`.

### change default config directory to .brigade instead of project root

This is partly inspired by innumerable other tools that keep their configuration in a dot-folder, but much more importantly, not expecting `package.json` to live in the project root means:

1. Your brigade script can have a _different_ `package.json`, and therefore different set of dependencies, from the rest of the project. This is of particular importance to JavaScript or TypeScript projects that are likely to have a `project.json` already.

1. When the prestart script runs `yarn install` inside the `.brigade` directory, it will not pollute the rest of the project. Again, this is of particular importance to JavaScript or TypeScript projects.

### major reworking of examples

The examples are updated to exercise various scenarios using  both JavaScript and TypeScript.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,4 @@
 .git
 .gocache
+v2/brigadier/node_modules
+v2/worker/node_modules
--- a/examples/git/github-demo.yaml
+++ b/examples/git/github-demo.yaml
@@ -1,0 +1,19 @@
+apiVersion: brigade.sh/v2
+kind: Project
+metadata:
+  id: github-demo
+description: A project that demonstrates integration with GitHub
+spec:
+  eventSubscriptions:
+  - source: github.com/brigadecore/brigade-github-gateway/v2
+    types:
+    - pull_request:opened
+    - pull_request:synchronize
+    - pull_request:reopened
+    - push
+    labels:
+      repo: krancour/brigade2-pipeline-demo
+  workerTemplate:
+    logLevel: DEBUG
+    git:
+      cloneURL: https://github.com/krancour/brigade2-pipeline-demo.git
--- a/examples/javascript/dependencies-demo.yaml
+++ b/examples/javascript/dependencies-demo.yaml
@@ -1,0 +1,23 @@
+apiVersion: brigade.sh/v2
+kind: Project
+metadata:
+  id: dependencies-demo
+description: A project that defines dependencies in package.json
+spec:
+  workerTemplate:
+    logLevel: DEBUG
+    defaultConfigFiles:
+      package.json: |-
+        {
+          "private": true,
+          "dependencies": {
+            "colors": "1.4.0"
+          }
+        }
+      brigade.js: |-
+        const { events } = require("@brigadecore/brigadier");
+        const colors = require("colors/safe");
+
+        events.on("github.com/brigadecore/brigade/cli", "exec", () => {
+          console.log(colors.green("It's not easy being green."));
+        });
--- a/examples/javascript/hello-world.yaml
+++ b/examples/javascript/hello-world.yaml
@@ -1,0 +1,15 @@
+apiVersion: brigade.sh/v2
+kind: Project
+metadata:
+  id: hello-world
+description: The simplest possible demonstration
+spec:
+  workerTemplate:
+    logLevel: DEBUG
+    defaultConfigFiles:
+      brigade.js: |-
+        const { events } = require("@brigadecore/brigadier");
+
+        events.on("github.com/brigadecore/brigade/cli", "exec", () => {
+          console.log("Hello, World!");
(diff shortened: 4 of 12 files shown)


radu-matei:
A few notes:

- you don't actually need a `package.json` to perform a `yarn add <package>`, which the previous version did with custom worker dependencies.
- all JavaScript / TypeScript projects require a `package.json` file - what happens when such a project needs to add a Brigade script? Is there any way to differentiate between the project's dependencies and Brigade's? This was the main reason a `brigade.json` file was added.

krancour:
> you don't actually need a package.json to perform a yarn add <package>, which the previous version did with custom worker dependencies.

Correct, but we were reading a non-canonical file (`brigade.json`) to get the names and versions of required dependencies to pass to `yarn add`. That's unnecessary if we just observe the conventional approach.

radu-matei:
Right that was mainly to avoid confusion with potentially existing JS/TS `package.json` files, and to maybe reuse the JSON file for other things (such as environment variables).

As long as the behavior os documented in this scenario, everything should be fine.

krancour:
> Right that was mainly to avoid confusion with potentially existing JS/TS `package.json` files

Yup. That made perfect sense at the time. Now, with Brigade-related files moved into the `.brigade` folder (by default), that confusion isn't an issue.

radu-matei (approved):

Issue #1134, opened 2020-10-20
Implement new git init image


krancour:
The Brigade 2.0 prototype reused the Brigade 1.x git init image. (Erroneously known previously as the "vcs sidecar".)

That image is configured by setting a _lot_ of environment variables. That's not entirely ideal because it makes it somewhat onerous to implement a custom image. More to the point, a lot of those env vars are pretty specific to git. Supposing we someday wanted to add support for a different VCS, like Mercurial, many of those env vars wouldn't be used, and new ones would be needed. Documenting all of this would be somewhat of a nightmare.

I'd prefer to take a cue from how the 2.0 (prototype) worker works. Instead of receiving configuration as environment variables, the entire event JSON is mounted into the container and the worker (or VCS init container in this case) can extract whatever information it needs from the event-- simple.

The existing git init image ("vcs sidecar") is implemented as a bash script that uses the git cli. Unmarshaling the event JSON in a bash script might be doable with something like `jq`, but it might be easier to switch from using bash to implement this image to using Go, TypeScript, or JavaScript instead. (Go would produce the smallest image, which might be desirable here.)

vdice:
Implemented in https://github.com/brigadecore/brigade/pull/1238; closing.

PR #1008, opened 2019-11-04
Adding support for ssh certificates.

vmasutin:
Tested with github, please see
https://github.blog/2019-08-14-ssh-certificate-authentication-for-github-enterprise-cloud/
for more details.

Signed-off-by: Valery Masiutsin <val.masutin@gmail.com>

**What this PR does / why we need it**:

**Special notes for your reviewer**:

**If applicable**:
- [x] this PR contains documentation
- [x] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #1229, opened 2021-01-14
Add git initializer (formerly vcs sidecar)

vdice:
- Updates the git initializer binary to be golang-based, for easier parsing of incoming event data and general SDK use
- ~~Adds bash-based test script (ported from v1) to test go binary directly~~
- Adds golang-based e2e test script to test functionality against a running Brigade server

Closes https://github.com/brigadecore/brigade/issues/1134

Apologies for the massive PR.  There still remains functionality to implement, most mentioned in TODO comments, which we can create follow-up issues for:

- Add retry logic to the git initializer
- Add support for SSH Certs  (should be as simple as grabbing the project secret and placing it in the appropriate file in the git initializer image)
- Other?

As for adding to CI, I've only added the go binary tests and am holding off adding the e2e tests.  I figure we should discuss e2e for v2 in general beforehand.

**If applicable**:


vdice:
Calling on git guru @adamreese for special attention if cycles permit!

brigade.js
```diff
@@ -90,6 +90,11 @@ function buildLoggerLinux() {
   return buildImage("logger-linux");
 }
 
+// Build the worker
+function buildWorker() {
+  return buildImage("worker");
+}
+
```
krancour, line 111: Let's remove this entire commit from this PR. Everything in here is already merged into the v2 branch.

It's not that it's going to cause a merge conflict-- it won't. It will just be easier to review if we can eliminate some noise.

Makefile
```diff
@@ -45,6 +45,8 @@ ifneq ($(SKIP_DOCKER),true)
 		-e SKIP_DOCKER=true \
 		-e DOCKER_USERNAME=$${DOCKER_USERNAME} \
 		-e DOCKER_PASSWORD=$${DOCKER_PASSWORD} \
+		-e VERSION="$(VERSION)" \
+		-e COMMIT="$(GIT_VERSION)" \
```
krancour, line 49: I know I was probably the one to put these two lines here, but they aren't needed.

The container that Kaniko runs in doesn't need a `VERSION` or `COMMIT` env var because, thanks to the changes on lines 159 and 160, the Kaniko container is passed the entire build command with version/commit info already set appropriately.

Again-- just trying to cut through some noise.

Makefile
```diff
@@ -253,4 +260,7 @@ hack: hack-push-images hack-build-cli
 		--set worker.image.pullPolicy=Always \
 		--set logger.linux.image.repository=$(DOCKER_IMAGE_PREFIX)logger-linux \
 		--set logger.linux.image.tag=$(IMMUTABLE_DOCKER_TAG) \
-		--set logger.linux.image.pullPolicy=Always
+		--set logger.linux.image.pullPolicy=Always \
+		--set gitInitializer.image.repository=$(DOCKER_IMAGE_PREFIX)git-initializer \
+		--set gitInitializer.image.tag=$(IMMUTABLE_DOCKER_TAG) \
+		--set gitInitializer.image.pullPolicy=Always
```
krancour, line 290: Nit: If we move the `--set gitInitializer.image...` lines above the `--set worker.image...` lines, it keeps two related things closer together.

brigade.js
krancour: I don't always understand how/why GitHub sometimes shows me commit-specific comments in different contexts...

So to clarify, I am talking about commit bc3808da46021780a261c855eea5990e28711d42.

examples/github-demo.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: brigade.sh/v2
+kind: Project
+metadata:
+  id: github-demo
+description: A project that demonstrates integration with GitHub
+spec:
+  workerTemplate:
+    git:
+      cloneURL: https://github.com/krancour/brigade2-pipeline-demo.git
```
krancour, line 9: Let's remove this new example.

In another recent commit already merged into `v2`, the `examples/` directory was restructured.

This example already exists in the `v2` branch, albeit at a different path.

Again-- just trying to cut through some noise.

v2/git-initializer/main.go
```diff
@@ -0,0 +1,265 @@
...
+
+	"github.com/urfave/cli/v2"
+
+	"github.com/brigadecore/brigade/sdk/v2/core"
+	"github.com/brigadecore/brigade/v2/internal/signals"
+	"github.com/brigadecore/brigade/v2/internal/version"
+)
+
+// Building w/ CLI frontend such that various opts can be easily overridden
+// for testing, e.g. payload file location, workspace location, etc. (see test.sh)
```
krancour, line 30: I'm still not quite in favor of building this as a CLI. I understand that made it easier to pass options, etc. when invoking this from a suite of bash-based tests, but I don't understand why the tests are bash-based to begin with.

If we were to stick to go-based tests then this program really doesn't need to be a CLI because we can directly test various function invocations per usual _and_ we'd spare ourselves from having to juggle an additional test suite that's implemented in another language.

Makefile
```diff
@@ -135,6 +135,23 @@ test-unit-js:
...
+			-ldflags "-w \
+				-X github.com/brigadecore/brigade/v2/internal/version.version=${VERSION} \
+				-X github.com/brigadecore/brigade/v2/internal/version.commit=${GIT_VERSION}" \
+			.
+
+.PHONY: test-git-initializer
+test-git-initializer: go-build-git-initializer
+	@cd v2/git-initializer && \
+		./test.sh
+
```
krancour, line 149: See [this comment]( https://github.com/brigadecore/brigade/pull/1229/commits/b591990ca5dff04b1bfbd2cbee45cb775954da54#r557473549)

I'd prefer not introducing a suite of bash-based tests if we can avoid it.

Makefile
```diff
@@ -135,6 +135,23 @@ test-unit-js:
...
+# can remove once we have a test harness around the Docker image
+.PHONY: go-build-git-initializer
+go-build-git-initializer:
+	@cd v2/git-initializer && \
+		go build \
+			-o ../../bin/git-initializer \
+			-ldflags "-w \
+				-X github.com/brigadecore/brigade/v2/internal/version.version=${VERSION} \
+				-X github.com/brigadecore/brigade/v2/internal/version.commit=${GIT_VERSION}" \
+			.
```
krancour: Is this meant to still be here?

sdk/v2/core/workers.go
```diff
@@ -98,6 +98,17 @@ func WorkerPhasesTerminal() []WorkerPhase {
...
+// is a terminal phase
+func IsTerminal(phase WorkerPhase) bool {
+	for _, terminalPhase := range WorkerPhasesTerminal() {
+		if phase == terminalPhase {
+			return true
+		}
+	}
+	return false
+}
+
```
krancour, line 111: Something like this is already in the `v2` branch. You can use `phase.IsTerminal()`

v2/git-initializer/e2e/e2e.go
```diff
@@ -0,0 +1,153 @@
...
+						return fmt.Errorf("worker's terminal phase: %s", phase)
+					}
+					return nil
+				}
+			}
+			return nil
+		}
+	}
+	return nil
+}
```
krancour, line 182: Since e2e tests aren't exclusively to the benefit of the new and improved git initializer, do you have any thoughts about adding an e2e test suite in a separate PR?

krancour:
Welp-- my usual strategy of reviewing PRs commit by commit (which I often encourage others to do also) may have backfired on me this time...

I'm realizing that I might have made comments on specific lines in specific commits that were resolved by _later_ commits.

Please feel free to ignore anything that seems like it's already resolved. Sorry for any confusion.

brigade.js
krancour: Disregard this comment. It looks like this was all straightened out by a recent merge commit.

Makefile
krancour: Nevermind. I see now that a subsequent commit removed it.

sdk/v2/core/workers.go
vdice: Oh great!  Not sure how I missed this :)

krancour: It was a very recent addition.

v2/git-initializer/main.go
vdice: Ok, I've removed the CLI portion and the ported bash tests.  We can work on a golang port of these if we'd like to see a lower, functional level of coverage of this binary.  For now, I've vetted continued working behavior via the e2e tests.

Makefile
vdice: Removed.

sdk/v2/core/workers.go
vdice: The original `IsTerminal()` method exists in the internal pkg, so I copied the same over into the external sdk pkg -- or should I also remove the original instance in internal and update references?

v2/git-initializer/e2e/e2e.go
vdice: Either way; indeed, they touch many parts of Brigade, despite the tested project files being GitHub-centric.

We can include the tests as is for now and file a ticket for follow-up/expansion -- or remove them from this PR and work on a more general e2e strategy instead?

examples/github-demo.yaml
vdice: Removed.

Makefile
vdice: Moved.

Makefile
vdice: Removed.

vdice:
I've never been able to figure out the fossa UI (failing check)... any help would be appreciated.  I ran a `go mod tidy` to clean up go.sum in v2... not sure what else we should do.

v2/git-initializer/main.go
krancour: That seems like a very fair middle ground.

Want to open an issue to improve unit test coverage on this component? I'll be happy to contribute.

sdk/v2/core/workers.go
krancour: Ooooooh. I forgot that this component would obviously be using the SDK and not types internal to the apiserver.

The copy is good. I know it doesn't seem DRY, but I like the share-nothing approach between SDK and apiserver. Since we anticipate SDKs being implemented in many different languages (a TS/JS one is already under construction), the share-nothing approach seems even more appropriate because then the Go SDK isn't special somehow.

👍

v2/git-initializer/e2e/e2e.go
krancour: Keep 'em because with the removal of the bash-based tests for the git initializer, the e2e tests are the only validation we currently have for that component.

Maybe still open an issue to _expand_ the e2e tests?

krancour:
> I've never been able to figure out the fossa UI (failing check)

Where is that even coming from?? I've never seen that check before.

v2/git-initializer/main.go
vdice: 👍 https://github.com/brigadecore/brigade/issues/1230

vdice:
The FOSSA check was intro'd/required when Brigade was adopted as a CNCF project... but again, since it rarely seems to complain, I haven't invested to much time in learning how to deal with it when it does!  I'll see if I can figure it out...

vdice:
Oops, accidentally hit the close button!  Re-opened.

vdice:
So, I was able to 'untrack' this branch from the FOSSA check.  I think it picked it up b/c this is a branch on origin rather than one of our forks.  Still don't know how to navigate their webpage/UI to drill down to the actual issues it's complaining about... we should do that before 2.0.0 but I'd say we are fine to not let it block this PR.

v2/git-initializer/e2e/e2e.go
vdice: 👍 added in https://github.com/brigadecore/brigade/issues/1231

vdice:
@radu-matei helped sort out the FOSSA check.  @krancour can you help with the pending brigade.js checks?  Not sure why they aren't reporting back.

radu-matei:
FOSSA seems to have flagged a dependency we don't actually use - https://github.com/shurcooL/gofontwoff
I haven't been able to find any reference whatsoever to that project in this PR or in Brigade, so I manually resolved the issue.

krancour:
I'll figure in the morning why the rest of the checks look to be stuck.

Makefile
```diff
@@ -133,6 +133,18 @@ test-unit-js:
...
 	'
 
+# Currently the e2e script is git initializer-centric
+#
+# Relies on `make hack` being run and/or the local kube context pointing
+# to a k8s cluster with Brigade running
+.PHONY: test-e2e
+test-e2e:
+	@kubectl --namespace brigade port-forward service/brigade-apiserver 7000:443 &>/dev/null & \
+		echo $$! > /tmp/brigade-apiserver.PID
```
krancour, line 143: Something looks suspicious here. The previous line doesn't end with `&&`, so how does this line run?

Makefile
```diff
@@ -153,12 +166,14 @@ build-logger-linux:
...
 	$(KANIKO_DOCKER_CMD) kaniko \
+		--build-arg VERSION="$(VERSION)" \
+		--build-arg COMMIT="$(GIT_VERSION)" \
 		--dockerfile /workspaces/brigade/v2/$*/Dockerfile \
 		--context dir:///workspaces/brigade/ \
 		--no-push
 
 .PHONY: push-images
-push-images: push-apiserver push-scheduler push-observer push-logger-linux
+push-images: push-apiserver push-scheduler push-observer push-logger-linux push-git-initializer
```
krancour, line 176: Push of the worker image is missing here. It's not a problem introduced by this PR, but would you mind fixing it right here?

v2/git-initializer/testdata/github-init-submodules.yaml
```diff
@@ -0,0 +1,17 @@
...
+    git:
+      cloneURL: https://github.com/sgoings/makeup.git
+      initSubmodules: true
+    defaultConfigFiles:
+      brigade.js: |-
+        const { events } = require("brigadier");
+
+        events.on("github.com/brigadecore/brigade/cli", "exec", () => {
+          console.log("Hello, World!");
+        });
```
krancour, line 17: Total nit: Add a trailing newline to avoid the red mark.

v2/git-initializer/testdata/github-malformed-url.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: brigade.sh/v2
+kind: Project
+metadata:
+  id: malformed-url
+description: A GitHub-based project with a purposely malformed url
+spec:
+  workerTemplate:
+    git:
+      cloneURL: https://github.com/brigadecore/brigadee.git
```
krancour, line 9: Total nit: Add a trailing newline to avoid the red mark.

v2/git-initializer/testdata/github-init-submodules.yaml
krancour: If you want.

v2/git-initializer/testdata/github-private-ssh.yaml
```diff
@@ -0,0 +1,17 @@
...
+    git:
+      cloneURL: git@github.com:vdice/testrepo.git
+      commit: a292c2ffe0226f702519ec2833054953e82a816e
+    defaultConfigFiles:
+      brigade.js: |-
+        const { events } = require("brigadier");
+
+        events.on("github.com/brigadecore/brigade/cli", "exec", () => {
+          console.log("Hello, World!");
+        });
```
krancour, line 17: Total nit: Add a trailing newline to avoid the red mark.

v2/git-initializer/testdata/github-public-https.yaml
```diff
@@ -0,0 +1,16 @@
...
+  workerTemplate:
+    git:
+      cloneURL: https://github.com/krancour/brigade2-pipeline-demo.git
+    defaultConfigFiles:
+      brigade.js: |-
+        const { events } = require("brigadier");
+
+        events.on("github.com/brigadecore/brigade/cli", "exec", () => {
+          console.log("Hello, World!");
+        });
```
krancour, line 16: Total nit: Add a trailing newline to avoid the red mark.

v2/git-initializer/e2e/e2e.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	if projectBytes, err = yaml.YAMLToJSON(projectBytes); err != nil {
+		return errors.Wrapf(err, "error converting file %s to JSON", filename)
+	}
+
+	project := core.Project{}
+	if err = json.Unmarshal(projectBytes, &project); err != nil {
+		return errors.Wrapf(err, "error unmarshaling project file %s", filename)
+	}
```
krancour, line 87: I have two mutually exclusive thoughts for how to improve these tests a smidge.

1. These tests are falling just short of being truly e2e because the CLI isn't being used to drive the tests. Executing `brig project create -f <path to test case>` would spare you from reading the file and converting it to JSON because the CLI would handle all of that for you. This would be more truly e2e, but it would also make these tests way uglier, so I don't entirely mind if we _don't_ do this right now, which takes me to option 2...

2. Since you're already using the SDK, consider not bothering with files for the test cases. You can enumerate the test cases _directly_ as `core.Project` objects at the top of this file (or in a separate file). It would spare you from messing with reading files and converting the yaml to json.

v2/git-initializer/e2e/e2e.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		return err
+	}
+	tokenStr := token.Value
+
+	client := sdk.NewAPIClient(
+		apiServerAddress,
+		tokenStr,
+		apiClientOpts,
+	)
```
krancour, line 105: Excellent use of the SDK. This belongs in the textbook. 😉

krancour:
@vdice knowing we intend to expand e2e tests, can I suggest moving `v2/git-initializer/e2e/e2e.go` to a less git-initializer-specific path? Something like `v2/e2e/`?

krancour:
@vdice another thought... instead of `e2e.go` being a main program, can we maybe use the traditional Go testing framework and a table-driven test approach so that the tests can be triggered with `go test yada yada`? Then the tests could include testify assertions and we'd end up with reports that pinpoint failure cases with greater precision than I think we have now.

v2/git-initializer/main.go
```diff
@@ -0,0 +1,238 @@
...
+
+	// Setup Auth
+	var auth transport.AuthMethod
+	// TODO: What about token-based auth?  (see v1 askpass.sh/BRIGADE_REPO_AUTH_TOKEN)
+
+	// TODO: Check for SSH Cert
+	// https://github.blog/2019-08-14-ssh-certificate-authentication-for-github-enterprise-cloud/
+	//
+	// (BRIGADE_REPO_SSH_CERT in v1)
+	// I think all we need to do is to make sure the cert file exists at /id_dsa-cert.pub
```
krancour, line 68: Remind me... when is this needed? It's just when the root CA for the git server isn't already trusted?

Makefile
vdice: Good eye.  It works b/c the background operator `&` can chain commands:
```
 $ echo "foo" & echo "bar"
[1] 71188
bar
foo
[1]+  Done                    echo "foo"
```
Here the second command just consumes the process ID created from the first (backgrounded) command and echos it to a file.

v2/git-initializer/e2e/e2e.go
vdice: Yea, I like 2.

Considering 1. and the comment below (to use the go testing framework), perhaps we move this to `main_test.go` and relegate proper e2e (using the cli as mentioned, etc.) to an effort down the road.  WDYT?

v2/git-initializer/e2e/e2e.go
vdice: 🍏 📖

v2/git-initializer/main.go
vdice: It's an optional feature and support for it exists in the v1 git-sidecar, thus I figured for parity we'd want to continue support.

Ref https://github.com/brigadecore/brigade/pull/1008

Makefile
krancour: TIL that the background operator can also chain commands.

v2/git-initializer/e2e/e2e.go
krancour: Yes. I like. Call this "integration" instead of e2e and we'll kick the e2e can down the road.

v2/git-initializer/main.go
krancour: I know v1 accommodated this. I'm just experiencing some kind of mental block in understanding it.

krancour: This is non-blocking... it was for my own edification.

v2/git-initializer/e2e/e2e.go
vdice: Ok, pushed a commit addressing the other comments; will proceed to working on this change next.

vdice:
@krancour this should be ready to revisit.  

My only lingering reservation remains in categorizing the git-initializer tests (now in `main_test.go`) as integration.  Although it isn't *full* e2e, I wonder if we tag/categorize as e2e anyways, since is does require a running brigade server.  Regardless, I haven't added the new tests to CI b/c of that latter fact: we need to stand up a server before running tests.  Not sure on the priority of plumbing this in to CI...

Makefile
```diff
@@ -251,6 +266,18 @@ hack: hack-push-images hack-build-cli
...
 		--set logger.linux.image.pullPolicy=Always
+
+.PHONY: hack-expose-apiserver
+hack-expose-apiserver:
+	@kubectl --namespace brigade port-forward service/brigade-apiserver 7000:443 &>/dev/null & \
+		echo $$! > /tmp/brigade-apiserver.PID
+
+.PHONY: hack-unexpose-apiserver
+hack-unexpose-apiserver:
+	@kill -TERM $$(cat /tmp/brigade-apiserver.PID)
```
krancour, line 283: This is interesting. I've been relying on a tweak I made to my personal `new-kind-cluster` script to handle this, but this is much nicer because it gives this ability to everyone who hacks on this and isn't limited to working with kind.

I like it.

v2/git-initializer/main.go
```diff
@@ -0,0 +1,238 @@
...
+				continue
+			}
+
+			// TODO: update this, as it is faulty... there may be multiple matches
+			// e.g. "main" might match /refs/heads/main and /refs/heads/main2
+			if strings.Contains(ref.Name().String(), fullRef.Name().String()) ||
+				strings.Contains(ref.Hash().String(), fullRef.Hash().String()) {
+				fullRef = ref
+				found = true
+			}
```
krancour, line 149: This is the only thing I can spot that I don't think can reasonably be deferred to a follow-up.

v2/git-initializer/main_test.go
```diff
@@ -0,0 +1,211 @@
...
+	"github.com/brigadecore/brigade/sdk/v2/core"
+	"github.com/brigadecore/brigade/sdk/v2/meta"
+	"github.com/brigadecore/brigade/sdk/v2/restmachinery"
+)
+
+type testcase struct {
+	name          string
+	project       core.Project
+	terminalPhase core.WorkerPhase
+}
```
krancour, line 24: Total nit: This is only used in one place, so you can make it an anonymous type. Down on line 34 you can do:

```
testCases := []struct{
  // define fields
} {
  {
    // Case 1
  },
  ...
  {
    // Case n
  },
}
```

v2/git-initializer/main_test.go
```diff
@@ -0,0 +1,211 @@
...
+				WorkerTemplate: core.WorkerSpec{
+					Git: &core.GitConfig{
+						CloneURL: "https://github.com/krancour/brigade2-pipeline-demo.git",
+					},
+					DefaultConfigFiles: map[string]string{
+						"brigade.js": defaultBrigadeJS,
+					},
+				},
+			},
+		},
```
krancour, line 51: I like this a lot better than pulling it in from files. 👍

v2/git-initializer/main_test.go
```diff
@@ -0,0 +1,211 @@
...
+		terminalPhase: core.WorkerPhaseSucceeded,
+	},
+	// TODO: once we have mech for injecting private ssh key in secure manner
```
krancour, line 54: Can it be a normal secret?

I think the missing mech you're talking about is the ability to define secrets declaratively?

In the meantime, we could add an optional "post project create" (terrible name) function to each test case. It can receive an API client as an argument and complete any setup-- which could include setting a secret.

Example:

```
testCases := []struct{
  project sdk.Project,
  postProjectCreate func(sdk.APIClient),
} {
  {
    project: ...,
    postProjectCreate: func(client sdk.APIClient) {
      err := client.Core().Project().Secrets().Set(...)
      require.NoErr(t, err)
    },
  },
  // ...
}
// ...
for _, testCase := range testCases {
  _, err := client.Core().Projects().Create(testCase.project)
  require.NoErr(t, err)
  if testCase.postProjectCreate != nil {
    testCase.postProjectCreate(client)
  }
}
```

Note example above was off the top of my head-- I didn't try it.

krancour:
@vdice this looks great!

I added a few comments-- almost entirely nits and/or things that can reasonably be deferred.

The only thing I would say I'm really concerned about is the logic for finding a full ref.

v2/git-initializer/main.go
vdice: Ok, logic improved/made more precise.  I think we're looking good here now.  (Passes integration tests and ye olde bash tests)

v2/git-initializer/main_test.go
vdice: Ah, great idea!  Indeed, I'm looking for a mech like that -- but the trickier bit is: What's a good approach for adding an ssh key that is valid for the private repo used in this test case.  There's setup involved: repo needs to exist, the public key needs to be added, the private key needs to be wired in to the test project as sketched above.  We'd want to support any dev running these locally and also in CI... 🤔

krancour: Ah..... I completely forgot that such a server would need to be set up. 🤦 

Follow up PR then?

v2/git-initializer/main.go
krancour: This is a major improvement!

I am wondering two things:

1. Is there still a problem due to `/refs/heads/foo` and `/refs/tags/foo` having the same short name?

1. It looks like this wouldn't allow the ref field to contain the entire long ref and still work. I feel like that has to be supported because "upstream event sources" (a GitHub web cook received by the vNext GitHub gateway, for instance) are likely to wheel and deal in long refs.

Maybe we can bypass a lot of this is we receive what's determined to be a ref in the form `/refs/heads/<branch>` or `/refs/heads/<tag>` and otherwise fall back on the above to make a "best effort" at matching?

vdice: Ooh good point re: 1.  Yes, let me look at what we could add around cases like these.

Re: 2, full/long refs would still be supported.  Since we convert whatever is supplied to us via `Ref` into a legit `plumbing.Reference` value via `fullRef := plumbing.NewReferenceFromStrings(commitRef, commitRef)`, we then have access to the full suite of methods such as `Short()`, etc.  Unless I'm not understanding the issue correctly?

krancour: > So, Ref comes in as /refs/heads/main and we gather the list of remote refs, where /refs/heads/main is the one we are after (exactly the same); the short names will match (main == main) and we're good.

Good up to that point.

> Well, unless there is a main tag 😄 , hence 1 above.

Right. That's where things fall apart.

I could give `/refs/heads/main` and end up pulling `/refs/tags/main`.

If I had _only_ given `main` I could understand how I ended up `/refs/tags/main`, but it would be especially worrisome to have started out being so precise and fully qualifying it as `/refs/heads/main` and still end up with the wrong thing.

vdice: Ah, gotcha.  Ok, I've improved logic in https://github.com/brigadecore/brigade/pull/1229/commits/f9f5348d0f202e6ad12b113be821810c5aee8907 thusly:

1. Don't list remotes/look for matches if the provided ref is already a full reference
2. If we find more than one match, error out with an appropriate message

WDYT?

v2/git-initializer/main_test.go
```diff
@@ -0,0 +1,40 @@
...
+		require.True(t, isFullReference(ref))
+	})
+
+	t.Run("short reference", func(t *testing.T) {
+		ref := plumbing.ReferenceName("foo")
+		require.False(t, isFullReference(ref))
+	})
+
+	t.Run("tasty reference", func(t *testing.T) {
+		ref := plumbing.ReferenceName("refs/tacos/carnitas")
```
krancour, line 37: 😂

v2/tests/git_initializer_test.go
```diff
@@ -0,0 +1,201 @@
...
+				WorkerTemplate: core.WorkerSpec{
+					Git: &core.GitConfig{
+						CloneURL: "https://github.com/brigadecore/empty-testbed.git",
+						Ref:      "non-existent",
+					},
+				},
+			},
+		},
+		terminalPhase: core.WorkerPhaseFailed,
+	},
```
krancour, line 101: I just noticed that we're missing a test case that covers using a commit sha instead of a ref.

vdice: Good call; added.

krancour (approved):
LGTM

vdice:
Closing and replacing with a PR from my fork (instead of origin) in attempts to get CI working properly; see https://github.com/brigadecore/brigade/pull/1238

PR #1232, opened 2021-01-16
Reorganize authz

krancour:
In a separate branch, I've been working on:

* Adding endpoints, services, database code, etc. to the API server for managing principals' (users and service accounts) roles.
* Enforcing access control.

Through that effort, it's become clear that some cleanup and refactoring of existing authz-related things is necessary in order to proceed. In an effort to avoid gigantic PRs, this PR includes said refactoring, while follow-up PRs will bring management for principals' roles and enforcement of access control.

This PR improves documentation (comments), moves things around so that (for instance) the `authx` package ceases to know anything about special project-scoped roles (which are a `core` concern), and generally keeps the definition of each role closer to where it is used. Some de-duping has occurred here as well-- no doubt because of a previously botched refactor.

_Note there are no functional changes in this PR._

This should leave me with a better foundation upon which to finish building out authz.


sdk/v2/api_client.go
```diff
@@ -13,13 +12,11 @@ import (
 type APIClient interface {
 	Authx() authx.APIClient
 	Core() core.APIClient
-	System() system.APIClient
 }
 
 type apiClient struct {
-	authxClient  authx.APIClient
-	coreClient   core.APIClient
-	systemClient system.APIClient
```
krancour, line 22: The `system` package and the API clients in it were the result of a previous failure to find the right home for stuff. The `system` package goes away on both the SDK/API client side and API server side and roles, logic, etc. previously found therein all moves to a better home. This is a consistent theme throughout this PR.

sdk/v2/authx/api_client.go
```diff
@@ -5,6 +5,8 @@ import "github.com/brigadecore/brigade/sdk/v2/restmachinery"
 // APIClient is the root of a tree of more specialized API clients within the
 // authx package.
 type APIClient interface {
+	// RoleAssignments returns a specialized client for managing RoleAssignments.
+	RoleAssignments() RoleAssignmentsClient
```
krancour, line 9: This is for managing system-level role assignments-- e.g. for roles like `ADMIN` or `PROJECT_CREATOR`.

sdk/v2/authx/principals.go
```diff
@@ -13,3 +13,14 @@ const (
...
+// PrincipalReference is a reference to any sort of security principal (human
+// user, service account, etc.)
+type PrincipalReference struct {
+	// Type qualifies what kind of principal is referenced by the ID field-- for
+	// instance, a User or a ServiceAccount.
+	Type PrincipalType `json:"type,omitempty"`
+	// ID references a principal. The Type qualifies what type of principal that
+	// is-- for instance, a User or a ServiceAccount.
+	ID string `json:"id,omitempty"`
+}
```
krancour, line 26: Using a `PrincipalReference ` makes `RoleAssignments` a little cleaner.

sdk/v2/authx/roles.go
```diff
@@ -59,3 +60,71 @@ func (r RoleAssignment) MarshalJSON() ([]byte, error) {
...
+	// principal also specified by the RoleAssignment.
+	Grant(context.Context, RoleAssignment) error
+	// Revoke revokes the system-level Role specified by the RoleAssignment for
+	// principal also specified by the RoleAssignment.
+	Revoke(context.Context, RoleAssignment) error
+}
+
+type roleAssignmentsClient struct {
+	*rm.BaseClient
+}
```
krancour, line 77: This is the client for system-level role assignments. Project-level role assignments are special, so they have their own separate client that hangs off the projects client.

sdk/v2/core/projects.go
```diff
@@ -172,17 +173,19 @@ type ProjectsClient interface {
 	// Delete deletes a single Project specified by its identifier.
 	Delete(context.Context, string) error
 
-	// Roles returns a specialized client for Project Role management.
-	Roles() ProjectRolesClient
+	// RoleAssignments returns a specialized client for managing project-level
+	// RoleAssignments.
+	RoleAssignments() authx.RoleAssignmentsClient
```
krancour, line 178: Project-level roles are "special," but minimally so. The scope of a project-level role == the ID of the project. On the server side, when you assign a principal to a project-level role (as opposed to a system-level role), we need to make sure the project exists! That's a concern that doesn't exist for system-level roles. Owing to this, there are (or will be, actually) separate endpoints on the API server for managing system-level and project-level role assignments. They'll actually share most of their code, but two separate endpoints == two separate clients.

v2/apiserver/internal/authx/named_principals.go
```diff
@@ -12,40 +12,16 @@ var (
...
-		RoleAdmin(),
-		RoleEventCreator(RoleScopeGlobal),
-		RoleProjectAdmin(RoleScopeGlobal),
-		RoleProjectCreator(),
-		RoleProjectDeveloper(RoleScopeGlobal),
-		RoleProjectUser(RoleScopeGlobal),
-		RoleReader(),
-	}
-}
-
```
krancour, line 26: All these named principals were deleted for now and will eventually reappear elsewhere. This wasn't the right place for them.

v2/apiserver/internal/authx/named_roles.go
```diff
@@ -1,161 +1,19 @@
...
-}
-
-// RoleProjectUser is the name of a project-level Role that enables principals
-// to create and manage Events for the specified Project.
-func RoleProjectUser(projectID string) Role {
-	return Role{
-		Type:  RoleTypeProject,
-		Name:  RoleNameProjectUser,
-		Scope: projectID,
+		Name: "ADMIN",
```
krancour, line 9: I killed off a lot of constants in favor of string literals because (on the server side, at least) these role names are each only used in one place.

PR #1236, opened 2021-01-22
work with different mongodb deployment topologies

krancour:
We were previously making an assumption that mongo was deployed in a replicated topology (even if the number of replicas was one).

This change makes the API server and the log agent (the only two components that talk to mongodb) tolerant of _either_ a standalone or replicated topology.


vdice (approved):
LGTM, pending passing CI

Issue #1220, opened 2021-01-08
Common schema definitions in standalone file?


vdice:
As mentioned in https://github.com/brigadecore/brigade/pull/1215, we have at least one instance where more than one schema references the same (but currently duplicated) definition.  (In this case, both the event and project schemas share an identical gitConfig definition).

Let's see if it is viable to create a common definitions schema file that can be used by more than one schema.  Or, do each need to be entirely standalone?

vdice:
Fixed/closed by #1240

PR #1240, opened 2021-01-22
Consolidate shared schema definitions

vdice:
* Consolidates common/shared schema definitions in a separate schema file.

Note: here I've settled on using the full file reference when referencing this common file.  When I attempt to use a relative reference, e.g. `"common.json#/definitions/url"`, we get:

```
open /brigade/schemas/github.com/brigadecore/brigade/v2/common.json: no such file or directory
```

But I can't tell where/how the full reference is resolved as this.  I would expect it to be resolved as `/brigade/schemas/common.json`, since that's where we place all of them in the apiserver image.

For what it's worth, we _could_ use the relative reference as shown above if we added the following line to the apiserver Dockerfile:

`COPY v2/apiserver/schemas/common.json /brigade/schemas/github.com/brigadecore/brigade/v2/common.json`

(I've tried placing *all* schemas there, but then we hit the same error, just doubled, b/c the same `github.com/brigadecore/brigade/v2` is added in no matter what.)

Closes https://github.com/brigadecore/brigade/issues/1220

### DIFF
--- a/v2/apiserver/schemas/common.json
+++ b/v2/apiserver/schemas/common.json
@@ -44,10 +44,6 @@
 				"ref": {
 					"type": "string",
 					"description": "A reference to a git branch or tag"
-				},
-				"initSubmodules": {
-					"type": "boolean",
-					"description": "Whether to initialize git submodules"
 				}
 			}
 		},
--- a/v2/apiserver/schemas/common.json
+++ b/v2/apiserver/schemas/common.json
@@ -25,7 +25,6 @@
 		"gitConfig": {
 			"type": "object",
 			"description": "Worker configuration pertaining specifically to git",
-			"additionalProperties": false,
 			"properties": {
 				"cloneURL": {
 					"oneOf": [
--- a/v2/apiserver/schemas/common.json
+++ b/v2/apiserver/schemas/common.json
@@ -1,0 +1,76 @@
+{
+	"$schema": "http://json-schema.org/draft-07/schema#",
+	"$id": "github.com/brigadecore/brigade/common.schema.json",
+	"description": "A common set of schema definitions shared by other Brigade schemas",
+
+	"definitions": {
+
+		"apiVersion": {
+			"type": "string",
+			"description": "The major version of the Brigade API with which this object conforms",
+			"enum": ["brigade.sh/v2"]
+		},
+
+		"description": {
+			"type": "string",
+			"minLength": 3,
+			"maxLength": 80
+		},
+
+		"empty": {
+			"type": "string",
+			"enum": [ "" ]
+		},
+
+		"gitConfig": {
+			"type": "object",
+			"description": "Worker configuration pertaining specifically to git",
+			"additionalProperties": false,
+			"properties": {
+				"cloneURL": {
+					"oneOf": [
+						{ "$ref": "#/definitions/empty" },
+						{ "$ref": "#/definitions/url" }
+					],
+					"description": "The URL for cloning a git project"
+				},
+				"commit": {
+					"type": "string",
+					"pattern": "^[a-fA-F0-9]*$",
+					"minLength": 8,
+					"maxLength": 40,
+					"description": "A git commit sha"
+				},
+				"ref": {
+					"type": "string",
+					"description": "A reference to a git branch or tag"
+				},
+				"initSubmodules": {
+					"type": "boolean",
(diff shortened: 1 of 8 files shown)


v2/apiserver/schemas/common.json
```diff
@@ -0,0 +1,76 @@
...
+				"ref": {
+					"type": "string",
+					"description": "A reference to a git branch or tag"
+				},
+				"initSubmodules": {
+					"type": "boolean",
+					"description": "Whether to initialize git submodules"
+				}
+			}
+		},
```
krancour, line 53: This shouldn't be common. When using this in a project definition, there is a an `initSubmodules` field that isn't supposed to be included when using this in an event....

But I just looked, and it _was_ identical in both places-- my fault; not yours.

Can this one section be removed from common can have it be defined differently for projects (include `initSubmodules` field) and events (exclude `initSubmodules`) field?

vdice: I bet we could reconfigure things to support that.

However, I'm wondering if there might be utility in allowing an event to toggle `initSubmodules` and override a project's default?  We'd need to update the event struct as well, of course.  But wanted to pose that possibility.

krancour: I don't recall v1 supporting it.

I wouldn't discount the possibility that there's utility in it for some edge case, but as things stand now, this is in the realm of things where the "project admin" (once roles are enforced) can specify permissible behavior. I think it's similar in many respects to the project admin setting the policy for whether jobs are permitted to run as privileged or mount the host's Docker socket-- i.e. a security concern. After all, our support for initializing submodules is an avenue for bringing third-party code that isn't part of a PR into the environment for compilation.

krancour: Ok... my thinking on this has changed, slightly.

If we don't initialize the submodules, there's nothing that actually stops someone from creating a job that does it, so allowing `initSubmodules` to be specified at the event-level really open a new avenue of attack. Consider my concerns over security moot.

That said, I can't find the utility of `initSubmodules` at the event-level. The practical use for _any_ event-level git details is to enabled CI/CD workflows by identifying specific coordinates (i.e. commit, ref, etc.) for code that is to be tested, built, deployed, etc. It's job is _not_ to say _how_ to build, but only _what_ to build.

Defining _how_ to build is the responsibility of the project definition.

For this reason, I stand by the notion that the `initSubmodules` field only makes sense in the context of the project-level git configuration.

vdice: I appreciate the thorough consideration, here!  I'm good with keeping things as is (no `initSubmodules` in the event schema's gitConfig) and will see what it looks like to consolidate the rest of the git bits.

krancour: sgtm

vdice: Updated, PTAL when convenient.

v2/apiserver/schemas/common.json
```diff
@@ -44,10 +44,6 @@
 				"ref": {
 					"type": "string",
 					"description": "A reference to a git branch or tag"
-				},
-				"initSubmodules": {
-					"type": "boolean",
-					"description": "Whether to initialize git submodules"
 				}
 			}
 		},
```
krancour, line 49: This whole type isn't common anymore, should this def move closer to where it's used in event.json?

vdice: The whole gitConfig type defined here _is_ common (event references it and only it, project references all of its properties and adds another)... I couldn't quite figure out how to consume the entire object in the latter case (project) via `additionalProperties: {...}` so I went the route of iterating through the common type's properties... lemme keep looking a bit as to whether it can be made more elegant.

krancour: > project references all of its properties and adds another

I think I overlooked that.

krancour: > lemme keep looking a bit as to whether it can be made more elegant.

Now that I'm reading it correctly, I happy as is.

But if you find something more elegant, that's good too.

krancour (approved):
LGTM

v2/apiserver/schemas/common.json
vdice: Updated to use `allOf` in tandem w/ `additionalProperties`.

PR #1241, opened 2021-01-22
Add role assignments

krancour:
Despite its size, this PR is pretty narrowly scoped:

Adds endpoints, services, and store for managing the relationships between principals (users, service accounts, etc.) and roles. The association between the two is made via a "role assignment." When a role is granted to a principal, it's a new role assignment. When a role is revoked from a principal, it's a deletion of a role assignment.

From a _logical_ perspective, there are two types of roles in Brigade v2. There are "system-level" roles and "project-level" roles. In case the distinction between the two isn't clear, project-level roles are in relation to a specific project. A principal could be the "ADMIN" of one project, but have no rights at all on other projects.

There is _one_ common set of store code that accommodates both of these concerns in order to be nice and DRY. There are, however, two sets of services, and two sets of endpoints to accommodate these concerns simply because there is additional logic that is applicable when granting or revoking a project-level role-- namely we need to make sure the project in question exists.

A note on the CLI: At both the system-level and project-level, there aren't very many roles. And roles are not user-definable. With the roles being finite, few, and well defined, I made a judgement call that the optimal UX has a dedicated command for each role.

So, for instance:

```
brig role grant EVENT_CREATOR <options>
```

In the above command, `EVENT_CREATOR` is technically a subcommand. By doing this, we're able to support a meaningful and well-documented set of options for each role.

I am very happy to give a walkthrough or demo of all of this if needed.

A follow-up PR will add functionality for listing/searching role assignments.


v2/apiserver/internal/authz/role_assignments.go
```diff
@@ -0,0 +1,177 @@
...
+				"error retrieving service account %q from store",
+				roleAssignment.Principal.ID,
+			)
+		}
+	default:
+		return nil
+	}
+
+	// Give them the Role
+	if err := r.roleAssignmentsStore.Grant(ctx, roleAssignment); err != nil {
```
vdice, line 111: If we wanted, we could simplify lines 111-122 to:
```
return errors.Wrapf(r.roleAssignmentsStore.Grant(ctx, roleAssignment), "error granting role..."...
```

(Similarly for Revoke below.)

No opinion either way; disregard if you'd already thought of it but prefer this version.

v2/apiserver/internal/core/rest/role_assignments_endpoints.go
```diff
@@ -0,0 +1,81 @@
+package rest
```
vdice, line 1: Ditto note here re: potentially updating file names for this and similar sibling files to `project_role_*.go`

v2/apiserver/schemas/project-role-assignment.json
```diff
@@ -0,0 +1,83 @@
+{
+	"$schema": "http://json-schema.org/draft-07/schema#",
+	"$id": "github.com/brigadecore/brigade/v2/project-role-assignment.schema.json",
+
+	"definitions": {
+
+    "apiVersion": {
```
vdice, line 7: nit: for `apiVersion`, we can reference the canonical def in common.json.  So, we can remove this from definitions and set up the reference on line 27 (currently) below.

v2/apiserver/schemas/project-role-assignment.json
```diff
@@ -0,0 +1,83 @@
...
+			"type": "string",
+			"description": "The major version of the Brigade API with which this object conforms",
+			"enum": ["brigade.sh/v2"]
+		},
+
+		"kind": {
```
vdice, line 13: Spacing looks off here... this definitions section looks to have 4 space indents as opposed to the section below?  Or maybe just GitHub display weirdness?

v2/apiserver/schemas/project-role-assignment.json
```diff
@@ -0,0 +1,83 @@
...
+	"required": ["apiVersion", "kind", "principal", "role"],
+	"additionalProperties": false,
+	"properties": {
+    "apiVersion": {
+      "$ref": "#/definitions/apiVersion"
+    },
+    "kind": {
+      "$ref": "#/definitions/kind"
+    },
+    "principal": {
```
vdice, line 32: Q: is this `principal` property the same between this and the RoleAssignment schema?  If so, we could potentially place in common.json

v2/cli/project_roles_commands.go
```diff
@@ -0,0 +1,238 @@
...
+
+var (
+	projectRoleGrantFlags = []cli.Flag{
+		&cli.StringFlag{
+			Name:     flagID,
+			Aliases:  []string{"i", flagProject, "p"},
+			Usage:    "Grant the role for the specified project (required)",
+			Required: true,
+		},
+		&cli.StringSliceFlag{
```
vdice, line 21: For this and similar commands, had we considered a single `--principal` flag to provide the principal ID (which could be a user or service account) instead of the two `-u` and `-s` flags?

vdice (commented):
Looking great!  Only had minor comments/Q's, none blocking.  But will hold off on final approval pending potential updates...

sdk/v2/core/role_assignments.go
```diff
@@ -0,0 +1,64 @@
+package core
```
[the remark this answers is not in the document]
krancour: Yes. We discussed that earlier and I said I would. Thank you for reminding me!

v2/apiserver/internal/authz/role_assignments.go
krancour: I always forget that there's no harm wrapping a nil error. Can do.

v2/apiserver/schemas/project-role-assignment.json
krancour: Good catch. I wrote this before your recent (and excellent) work DRYing up the schemas. When I rebased, I forgot to take advantage of the common type definitions you established. I will fix accordingly.

v2/apiserver/schemas/project-role-assignment.json
krancour: I will check.

v2/apiserver/schemas/project-role-assignment.json
krancour: Yes. Will extract. This another case of the common defs not having been established before I wrote this.

v2/cli/project_roles_commands.go
krancour: If we did that, we'd need some way (another flag) to disambiguate whether the principal in question is a user or a service account. If we keep the two separate flags, that context is intrinsic.

sdk/v2/authz/role_assignments.go
```diff
@@ -0,0 +1,132 @@
...
+	PrincipalTypeUser PrincipalType = "USER"
+)
+
+// PrincipalReference is a reference to any sort of security principal (human
+// user, service account, etc.)
+type PrincipalReference struct {
+	// Type qualifies what kind of principal is referenced by the ID field-- for
+	// instance, a User or a ServiceAccount.
+	Type PrincipalType `json:"type,omitempty"`
+	// ID references a principal. The Type qualifies what type of principal that
```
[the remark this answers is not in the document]
krancour: There's no question that if you're looking directly at this code, the repeated description of the `Type` field in the description of the `ID` field seems redundant.

Since the SDK is intended for broad consumption, consider someone who isn't looking directly at this code or even the godocs generated from it, but is instead only seeing dynamic context help in their IDE, as in the attached screen cap. I think the repeated description of the `Type` field provides critical context that such a user needs to know about.

Screen Shot 2021-01-27 at 12 51 22 PM

v2/apiserver/internal/authz/role_assignments.go
krancour: Went to apply this and it looks like this:

```
// Give them the Role
return errors.Wrapf(
  r.roleAssignmentsStore.Grant(ctx, roleAssignment),
  "error granting role %q with scope %q to %s %q in store",
  roleAssignment.Role.Name,
  roleAssignment.Role.Scope,
  roleAssignment.Principal.Type,
  roleAssignment.Principal.ID,
)
```

Invoking "clear is better than clever," I do think I prefer how it was before.

vdice: Ah yea, that's quite a return statement!  Okay, I think I do, too!  Thanks for taking a look.

v2/apiserver/schemas/project-role-assignment.json
krancour: Our JSON schemas have consistently used tabs. Some spaces snuck in here instead.

To avoid the religious was over tabs v spaces, I'll state that I have no preference for one over the other and instead usually opt for whatever is considered conventional for any given language or file format. (Go, for instance, says to use tabs.)

I'm not aware their being any established convention for tabs v spaces in JSON, so I'm going to change these spaces to tabs to be consistent with the rest of this file and with all our other JSON files.

sdk/v2/core/role_assignments.go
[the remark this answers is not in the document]
krancour: Renamed

v2/apiserver/internal/authz/role_assignments.go
krancour: I'm less concerned with it being big and more concerned that a very important function call is buried inside what might be dismissed as "just error-handling code."

v2/apiserver/internal/core/rest/role_assignments_endpoints.go
krancour: Renamed

v2/apiserver/schemas/project-role-assignment.json
krancour: Fixed

v2/apiserver/schemas/project-role-assignment.json
krancour: Fixed

krancour:
There's one thing I know needs to be fixed. In the common JSON schema, the `ID` field in the `PrincipalReference` type uses a regex that isn't tolerant of email addresses, which is the likely ID format for a user (depends on what third party identity providers use as ID, but it's usually an email address).

v2/apiserver/schemas/project-role-assignment.json
vdice: Ah ok; Yea, no pref here either.  This sounds good.  I think I have my vs code setting at "intuit from file whether to insert tabs v spaces"

v2/apiserver/internal/authz/role_assignments.go
```diff
@@ -85,14 +85,20 @@ func (r *roleAssignmentsService) Grant(
...
 				"error retrieving user %q from store",
 				roleAssignment.Principal.ID,
 			)
 		}
+		// From an end-user's perspective, User IDs are case insensitive, but when
+		// creating a role assignment, we'd like to respect case. So we DON'T use
+		// the ID from the inbound RoleAssignment-- which may have incorrect case.
+		// Instead we replace it with the ID (with correct case) from the User we
+		// found.
+		roleAssignment.Principal.ID = user.ID
```
vdice, line 101: Ah, good thinking.

vdice (approved):
LGTM!

PR #1246, opened 2021-01-25
make examples reference project schema

krancour:
Provided you have the popular Red Hat YAML plugin for VS Code, this change provides context help when tinkering with examples.


vdice (approved):
Nice!

Issue #1224, opened 2021-01-13
bug: observer doesn't notice vcs init container failures


krancour:
If the worker's vcs init container fails, the worker should be transitioned to a failed phase, but it isn't currently.

Similarly, if a job's vcs init container fails, the job should be transitioned to a failed phase, but it isn't currently.

vdice:
Let's check this again soon.  At least with sad case e2e test added in https://github.com/brigadecore/brigade/pull/1229 (malformed git repo url), the worker appears to fail correctly.

krancour:
@vdice can I reassign this one to you? As you indicated, this bears double-checking.

vdice:
I have a branch with updated vcs tests and finer-grained assertions -- that I'll PR regardless -- but I was curious around the following:

> Similarly, if a job's vcs init container fails, the job should be transitioned to a failed phase, but it isn't currently.

I can't quite figure out how to test the scenario where a worker's vcs init container succeeds -- as it must if any jobs are to be started -- but then setting things up for the job's vcs init container to fail.  This seems currently impossible, as the job's vcs details are always inherited from the worker and don't appear to be overrideable.  Or did you have an approach in mind for this scenario?

krancour:
Good point. Realistically, if a worker's git initializer succeeded, any and all jobs that also use a git initializer should have that initializer succees, unless.... unless there's a network failure or some such, but I'm not sure how we'd inject such a fault.

Let's just make sure the observer gets everything right when a worker's git initializer fails and then trust tuat the similar logic works for jobs with failed initializers also.

vdice:
Nothing to do here; https://github.com/brigadecore/brigade/pull/1252 added test verification.

PR #1252, opened 2021-01-28
ref(git_initializer_test.go): add finer-grained assertions and more test scenarios

vdice:

* Adds finer-grained assertions (worker and job logs/terminal phases, vcs logs, etc.) for the git initializer integration tests
* Adds a test scenario around a job failure (which should provoke both job and worker terminal phases of failed)

Closes https://github.com/brigadecore/brigade/issues/1224

**If applicable**:

### DIFF
--- a/v2/tests/git_initializer_test.go
+++ b/v2/tests/git_initializer_test.go
@@ -3,8 +3,8 @@
 package tests
 
 import (
+	"bytes"
 	"context"
-	"log"
 	"testing"
 	"time"
 
@@ -16,20 +16,18 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-var defaultConfigFiles = map[string]string{
-	"brigade.js": `
-		const { events } = require("brigadier");
+type testcase struct {
+	name                       string
+	project                    core.Project
+	configFiles                map[string]string
+	expectedJobLogsContains    string
+	expectedJobPhase           core.JobPhase
+	expectedWorkerLogsContains string
+	expectedWorkerPhase        core.WorkerPhase
+	expectedVCSLogsContains    string
+}
 
-		events.on("github.com/brigadecore/brigade/cli", "exec", () => {
-			console.log("Hello, World!");
-		});
-	`}
-
-var testcases = []struct {
-	name          string
-	project       core.Project
-	terminalPhase core.WorkerPhase
-}{
+var testcases = []testcase{
 	{
 		name: "GitHub - no ref",
 		project: core.Project{
@@ -41,7 +39,9 @@ var testcases = []struct {
 				},
 			},
 		},
-		terminalPhase: core.WorkerPhaseSucceeded,
+		expectedJobLogsContains: "README.md",
+		expectedJobPhase:        core.JobPhaseSucceeded,
+		expectedWorkerPhase:     core.WorkerPhaseSucceeded,
 	},
 	{
 		name: "GitHub - full ref",
@@ -55,7 +55,9 @@ var testcases = []struct {
 				},
 			},
 		},
-		terminalPhase: core.WorkerPhaseSucceeded,
+		expectedJobLogsContains: "README.md",
+		expectedJobPhase:        core.JobPhaseSucceeded,
+		expectedWorkerPhase:     core.WorkerPhaseSucceeded,
 	},
 	{
 		name: "GitHub - casual ref",
@@ -69,7 +71,9 @@ var testcases = []struct {
 				},
 			},
 		},
-		terminalPhase: core.WorkerPhaseSucceeded,
+		expectedJobLogsContains: "README.md",
+		expectedJobPhase:        core.JobPhaseSucceeded,
+		expectedWorkerPhase:     core.WorkerPhaseSucceeded,
 	},
 	{
 		name: "GitHub - commit sha",
@@ -83,21 +87,9 @@ var testcases = []struct {
 				},
 			},
 		},
-		terminalPhase: core.WorkerPhaseSucceeded,
-	},
-	{
-		name: "GitHub - non-existent ref",
-		project: core.Project{
-			Spec: core.ProjectSpec{
-				WorkerTemplate: core.WorkerSpec{
-					Git: &core.GitConfig{
-						CloneURL: "https://github.com/brigadecore/empty-testbed.git",
-						Ref:      "non-existent",
-					},
-				},
-			},
-		},
-		terminalPhase: core.WorkerPhaseFailed,
+		expectedJobLogsContains: "README.md",
+		expectedJobPhase:        core.JobPhaseSucceeded,
+		expectedWorkerPhase:     core.WorkerPhaseSucceeded,
 	},
 	{
 		name: "GitHub - submodules",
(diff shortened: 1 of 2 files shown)


krancour:
Just started looking at this and will finish tomorrow, but I'm very excited about this because we're approaching a point where I can reliably replace all the manual things I do to test e2e by just running this instead.

v2/tests/git_initializer_test.go
```diff
@@ -179,46 +231,150 @@ func TestMain(t *testing.T) {
...
+	for {
+		event, err := client.Core().Events().Get(ctx, e.ID)
+		require.NoError(t, err, "error getting event")
+
+		phase := event.Worker.Status.Phase
+		if phase.IsTerminal() {
+			require.Equal(t, tc.expectedWorkerPhase, phase, "worker's terminal phase does not match expected")
+			break
+		}
+	}
```
krancour, line 265: If a worker ever gets stuck in a non-terminal phase, this will loop infinitely.

If I may suggest, you can use `client.Core().Events().Workers().WatchStatus(ctx, e.ID)`. It will give you back two channels-- one for status updates and one for errors.

(Tangent, this is more efficient also because it only makes a single connection to the API server and the underlying goroutine sleeps when there aren't updates.)

With those channels in hand, you can do a select like this to implement a timeout so we never get stuck even if a worker gets stuck:

```
statucCh, errCh, err := client.Core().Events().Workers().WatchStatus(ctx, e.ID)
require.NoError(t, err)
timer := time.NewTimer( <some timeout duration> )
defer timer.Stop()
for {
  select {
    case status := <-statusCh:
      // If terming, assert expected phase
    case err := <- errCh:
      // Deal with error
    case <- timer.C:
      // Deal with timeout
  }
}
```

v2/tests/git_initializer_test.go
```diff
@@ -179,46 +231,150 @@ func TestMain(t *testing.T) {
...
+
+		job := jobs["ls"]
+		require.NotNil(t, job, "expected job does not exist")
+
+		phase := job.Status.Phase
+		if phase.IsTerminal() {
+			require.Equal(t, tc.expectedJobPhase, phase, "job's terminal phase does not match expected")
+			break
+		}
+	}
```
krancour, line 323: See [this comment](https://github.com/brigadecore/brigade/pull/1252/files#r569511298).

There is a function for streaming job status updates that could be helpful here.

v2/tests/git_initializer_test.go
```diff
@@ -179,46 +231,150 @@ func TestMain(t *testing.T) {
...
+	tc testcase) {
+
+	for {
+		event, err := client.Core().Events().Get(ctx, e.ID)
+		require.NoError(t, err, "error getting event")
+
+		jobs := event.Worker.Jobs
+		require.Equal(t, 1, len(jobs), "expected job count is not 1")
+
+		job := jobs["ls"]
```
krancour, line 315: Maybe the job name can be passed in instead of being hardcoded here? As it stands now, `assertJobTerminalPhase ` isn't useful for hypothetical test cases that are constructed differently and do not have a job named `ls`.

v2/tests/git_initializer_test.go
```diff
@@ -179,46 +231,150 @@ func TestMain(t *testing.T) {
...
+}
+
+func assertJobLogs(
+	t *testing.T,
+	ctx context.Context,
+	client sdk.APIClient,
+	e core.Event,
+	tc testcase) {
+
+	selector := &core.LogsSelector{Job: "ls"}
```
krancour, line 333: Same comment as [here](https://github.com/brigadecore/brigade/pull/1252/files#r569513877)

v2/tests/git_initializer_test.go
```diff
@@ -179,46 +231,150 @@ func TestMain(t *testing.T) {
...
+		case <-ctx.Done():
+			break
+		}
+
+		// log and err channels empty, let's return logs
+		if logEntryCh == nil && errCh == nil {
+			return b.String()
+		}
+	}
+}
```
krancour, line 380: If the container you're capturing logs from produces enough lines of output, it's possible to blow up the heap here.

fwiw, this is why the API always streams logs to begin with and doesn't expose any option to retrieve them as a single chunk-- risk of blowing up the heap on either the client side _or_ server side.

Instead of capturing a potentially large chunk of logs, which I believe you are then asserting contain some text, what if you pass in the text you're expecting to find, ditch the buffer, and scan the log stream for the specified text until you either find it or reach the end of the stream?

krancour: > If the container you're capturing logs from produces enough lines of output, it's possible to blow up the heap here.

And I acknowledge that as written, these tests probably won't do that.

I'm more concerned about:

1. What happens when something goes wrong. We don't want an OOM obscuring the actual issue.

1. Keeping this flexible enough to accommodate hypothetical test cases that do produce larger amounts of logs.

v2/tests/git_initializer_test.go
vdice: This is much nicer; I had been sneakily using the context timeout as a lazyman's test timeout.  Updated with this suggestion.

v2/tests/git_initializer_test.go
vdice: Updated, thank you.

v2/tests/git_initializer_test.go
vdice: Now tracked via a const.

v2/tests/git_initializer_test.go
vdice: Good call; updated!

v2/tests/git_initializer_test.go
```diff
@@ -16,20 +17,23 @@ import (
...
+type testcase struct {
+	name                       string
+	project                    core.Project
+	configFiles                map[string]string
+	expectedJobLogsContains    string
+	expectedJobPhase           core.JobPhase
+	expectedWorkerLogsContains string
+	expectedWorkerPhase        core.WorkerPhase
+	expectedVCSLogsContains    string
+}
```
krancour, line 34: I'm a little worried that test cases constructed this way might be too rigid for future expansion.

Every case is following the same formula: take this projects, create an event for it, check that _these specific things are true_ (i.e. the worker ended up in this phase, the job ended up in this phase, the git init logs contain this phrase, the worker logs contain this other phrase, etc.)

Things get messy if we want to deviate from that formula. So, for instance, suppose we wanted to add a test case with no git initializer? (I know these are the git initializer tests, but I think we both agreed that this was also the seed for a more extensive suite of integration tests.) What about test cases that involve multiple jobs? I already see at least one spot in the code where we conditionally ignore some assertions based on test case name... you can see how those sorts of conditionals could grow over time...

I'm wondering if we can't make this more flexible and sustainable over time (and pave the way for a full battery of int tests) by structuring the test cases like this:

```
type testCase struct {
  name string
  project core.Project
  event core.Event
  assertions func(t *testing.T, eventID string, eventClient core.EventClient)
}
```

This would allow us to, case by case, say here's the project, here's the event to create for that project, and here's a function that expresses all the things I expect to be true. In each case, that function can quite easily utilize all the great "assertion" helpers you've already written here, which do a fantastic job of making complex assertions (like scanning the log stream for a specific phrase).

In the end, this might admittedly make our definition of individual test cases more verbose, but imho, it would pay dividends by making the int test "framework" more flexible and cleaner (no conditionals to accommodate specific cases).

vdice: Thanks, definitely great foresight here.  I'll work on refactoring as illustrated with `assertions` built into the test case.  This is the predominant pattern that's been established in other tests, too, so it will be great to have alignment.

vdice: Ok, updated.  PTAL when convenient.

v2/tests/git_initializer_test.go
```diff
@@ -179,46 +235,201 @@ func TestMain(t *testing.T) {
...
+	t *testing.T,
+	ctx context.Context,
+	client sdk.APIClient,
+	e core.Event,
+	tc testcase,
+) {
+	statusCh, errCh, err := client.Core().Events().Workers().Jobs().WatchStatus(
+		ctx,
+		e.ID,
+		testJobName,
```
krancour, line 355: I'd still rather see job name passed in so that not all test cases are constrained to having only one job with one specific name. Similar for the `assertJobLogs` func.

After this, this PR is spectacular.

vdice: Oops!  Now added.

krancour (approved):
LGTM!

PR #1253, opened 2021-01-28
enforce access controls

krancour:
Finally... the PR the _enforces_ the access controls.

### DIFF
--- a/v2/apiserver/internal/authz/mongodb/role_assignments_store.go
+++ b/v2/apiserver/internal/authz/mongodb/role_assignments_store.go
@@ -6,6 +6,7 @@ import (
 	"github.com/brigadecore/brigade/v2/apiserver/internal/authz"
 	"github.com/brigadecore/brigade/v2/apiserver/internal/lib/mongodb"
 	"github.com/pkg/errors"
+	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
@@ -61,4 +62,33 @@ func (r *roleAssignmentsStore) Revoke(
 		)
 	}
 	return nil
+}
+
+func (r *roleAssignmentsStore) Exists(
+	ctx context.Context,
+	roleAssignment authz.RoleAssignment,
+) (bool, error) {
+	criteria := bson.M{
+		"role.type":      roleAssignment.Role.Type,
+		"role.name":      roleAssignment.Role.Name,
+		"principal.type": roleAssignment.Principal.Type,
+		"principal.id":   roleAssignment.Principal.ID,
+	}
+	if roleAssignment.Role.Scope == "" {
+		criteria["role.scope"] = bson.M{
+			"$exists": false,
+		}
+	} else {
+		criteria["role.scope"] = bson.M{
+			"$in": []string{roleAssignment.Role.Scope, "*"},
+		}
+	}
+	err := r.collection.FindOne(ctx, criteria).Err()
+	if err == mongo.ErrNoDocuments {
+		return false, nil
+	}
+	if err != nil {
+		return false, errors.Wrap(err, "error finding role assignment")
+	}
+	return true, nil
 }
--- a/v2/apiserver/internal/authz/mongodb/role_assignments_store.go
+++ b/v2/apiserver/internal/authz/mongodb/role_assignments_store.go
@@ -40,7 +40,7 @@ func (r *roleAssignmentsStore) Grant(
 		&options.FindOneAndReplaceOptions{
 			Upsert: &tru,
 		},
-	); res.Err() != nil {
+	); res.Err() != nil && res.Err() != mongo.ErrNoDocuments {
 		return errors.Wrapf(
 			res.Err(),
 			"error upserting role assignment %v",
--- a/v2/apiserver/internal/core/projects.go
+++ b/v2/apiserver/internal/core/projects.go
@@ -5,6 +5,9 @@ import (
 	"encoding/json"
 	"time"
 
+	"github.com/brigadecore/brigade/v2/apiserver/internal/authn"
+	"github.com/brigadecore/brigade/v2/apiserver/internal/authz"
+	libAuthn "github.com/brigadecore/brigade/v2/apiserver/internal/lib/authn"
 	libAuthz "github.com/brigadecore/brigade/v2/apiserver/internal/lib/authz"
 	"github.com/brigadecore/brigade/v2/apiserver/internal/meta"
 	"github.com/brigadecore/brigade/v2/apiserver/internal/system"
@@ -156,10 +159,11 @@ type ProjectsService interface {
 }
 
 type projectsService struct {
-	authorize     libAuthz.AuthorizeFn
-	projectsStore ProjectsStore
-	eventsStore   EventsStore
-	substrate     Substrate
+	authorize            libAuthz.AuthorizeFn
+	projectsStore        ProjectsStore
+	eventsStore          EventsStore
+	roleAssignmentsStore authz.RoleAssignmentsStore
+	substrate            Substrate
 }
 
 // NewProjectsService returns a specialized interface for managing Projects.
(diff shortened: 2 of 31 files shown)


krancour:
Test failures are legitimate. Will fix shortly.

v2/apiserver/internal/core/substrate.go
```diff
@@ -68,19 +70,33 @@ type SubstrateService interface {
...
 
 func (s *substrateService) CountRunningWorkers(
 	ctx context.Context,
 ) (SubstrateWorkerCount, error) {
+	// At present, only the scheduler ever needs to know how many Workers
+	// are currently executing, but it's very easy to imagine new tooling, like
+	// a hypothetical new dashboard, wanting to grant users some summary-level
+	// insight into what's going on with the substrate, so we'll only require
+	// system.RoleReader() to authorize this operation.
+	if err := s.authorize(ctx, system.RoleReader()); err != nil {
```
vdice, line 96: Good thinking here; makes sense.

v2/apiserver/internal/lib/authz/authorizers.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"context"
+
+	"github.com/brigadecore/brigade/v2/apiserver/internal/meta"
+)
+
+// AuthorizeFn is the signature for any function that can, presumably, retrieve
+// a principal from the provided Context and make an access control decision
+// based on the principal having (or not having) at least on of the provided
```
vdice, line 11:
```suggestion
// based on the principal having (or not having) at least one of the provided
```

vdice (commented):
This looks great to me!  Looks like CI may need attention... will delay final approval on the chance that new changes come in.

krancour:
@vdice I have another commit I want to tack onto this. Will amend shortly.

It's to make the creator of a new project its admin.

v2/apiserver/internal/authz/mongodb/role_assignments_store.go
```diff
@@ -40,7 +40,7 @@ func (r *roleAssignmentsStore) Grant(
 		&options.FindOneAndReplaceOptions{
 			Upsert: &tru,
 		},
-	); res.Err() != nil {
+	); res.Err() != nil && res.Err() != mongo.ErrNoDocuments {
```
krancour, line 43: TIL that an "upsert" returns a `mongo.ErrNoDocuments` error on a successful _insert_ (but not on an _update_). I want to suppress that because it's not actually any kind of a problem.

v2/apiserver/internal/core/projects.go
```diff
@@ -200,6 +206,71 @@ func (p *projectsService) Create(
 			errors.Wrapf(err, "error storing new project %q", project.ID)
 	}
 
+	// Make the current user an admin, developer, and user of the project
```
krancour, line 209: If we _don't_ do this, not only can no one update or use this project, but the project won't even have a user with sufficient permissions to change that. i.e. If we don't do this, every project that is created is completely useless unless you are the root user.

krancour:
@vdice access requirements, summarized for your convenience:

```
Service Accounts:

  Create: System Admin
  List: Reader
  Get: Reader
  Lock: System Admin
  Unlock: System Admin

Users:

  List: Reader
  Get: Reader
  Lock: System Admin
  Unlock: System Admin

System Role Assignments:

  Grant: System Admin
  Revoke: System Admin

Events:

  Create (project ID specified): Project User
  Create (project ID not specified): Event Creator
  List: Reader
  Get: Reader
  Cancel: Project User
  Cancel Many: Project User
  Delete: Project User
  Delete Many: Project User

Jobs:

  Create: Worker
  Start: Scheduler
  Get Status: Reader
  Watch Status (stream status changes): Reader
  Update Status: Observer
  Cleanup: Observer

Logs:

  Stream: Project User

Project Role Assignments:

  Grant: Project Admin
  Revoke: Project Admin

Projects:

  Create: Project Creator
  List: Reader
  Get: Reader
  Update: Project developer
  Delete: Project Admin

Secrets:

  List (values always redacted): Reader
  Set: Project Admin
  Unset: Project Admin

Substrate:

  Count Running Workers: Reader
  Count Running Jobs: Reader

Workers:

  Start: Scheduler
  Get Status: Reader
  Watch Status (stream status changes): Reader
  Update Status: Observer
  Cleanup: Observer
```

v2/apiserver/internal/core/projects.go
```diff
@@ -200,6 +206,71 @@ func (p *projectsService) Create(
...
+	case *authn.ServiceAccount:
+		principalRef = authz.PrincipalReference{
+			Type: authz.PrincipalTypeServiceAccount,
+			ID:   prin.ID,
+		}
+	default:
+		return project, nil
+	}
+
+	if err = p.roleAssignmentsStore.Grant(
```
vdice, line 228: I was thinking perhaps simply granting admin rights would be sufficient (it'd already come with authorization that the developer and user role assignments would bring.)  Can you elaborate on the need for explicitly adding all?

v2/apiserver/internal/core/projects.go
```diff
@@ -200,6 +206,71 @@ func (p *projectsService) Create(
...
+			Type: authz.PrincipalTypeUser,
+			ID:   prin.ID,
+		}
+	case *authn.ServiceAccount:
+		principalRef = authz.PrincipalReference{
+			Type: authz.PrincipalTypeServiceAccount,
+			ID:   prin.ID,
+		}
+	default:
+		return project, nil
```
vdice, line 225: So in this case do we indeed have a project that is "useless", i.e. no principals are authz'd to interact w/ it?

vdice (commented):
A couple Q's on the latest commit...

vdice:
> @vdice access requirements, summarized for your convenience:

Thank you for these!  Do you think it'd be handy to preserve this list (or something similar) in source somewhere?

v2/apiserver/internal/lib/authz/authorizers.go
krancour: Will fix.

v2/apiserver/internal/core/projects.go
krancour: > it'd already come with authorization that the developer and user role assignments would bring

Partially accurate.

As things stand currently, an admin can manage permissions, secrets, and delete the project.

A user can _use_ the project (like create events), but cannot modify it in any way.

A developer can update the definition of the project.

These are deliberately separate. You can easily imagine three different personas fulfilling these roles.

Imagine a team lead is the admin since he or she might have access to secrets that others are not privy to.

Imagine that a member of the team is a developer.

Imagine that some person in a business unit is a user.

Now..... all of that said, someone who is an admin _can_ grant themselves the other two roles, but they don't have them automatically.

My intent here was that when a project is created, the party who created it is automatically given the _full_ range of things that can be done with/to the project.

Happy to discuss more if you think this can be improved upon.

v2/apiserver/internal/core/projects.go
krancour: The only plausible and likely scenario here is that the principal is the root user. The root user already has the admin bit on all projects (scope `*`).

krancour:
> Do you think it'd be handy to preserve this list (or something similar) in source somewhere?

Let's make a point of adding these access requirements to the API docs when we eventually get around to something like that.

v2/apiserver/internal/core/projects.go
vdice: Thanks for laying out those scenarios; definitely makes sense.

What still hasn't explicitly been mentioned is that a project admin (or, strictly speaking, a principal whose only roleAssigment for a project is RoleProjectAdmin) is not automatically authorized to use or develop on the project itself (at least in regards to the authorizations that the associated roleAssignments for those two bring), right?  If I have this right, I think it is okay for now... it may just be somewhat unexpected for some Brigade users ('users' in a general sense here).  Or, it may not.  Therefore, it seems prudent to go with what you've laid out until (if) we get feedback in this area.

krancour: The help says this:

> ADMIN      Grant the ADMIN project role, which enables management of all apects of the project, including its secrets, as well as project-level permissions for other users and service accounts.

Similar comments exist in the code.

You're right that "enables management of all apects of the project" (there's a typo in there, isn't there?) could be misleading.

vdice (approved):
LGTM!  Awesome work.

Issue #1202, opened 2020-12-19
project deletes aren't cascading to include deletion of project events


krancour:

krancour:
@vdice would you mind taking this one?

I think you just need to make the project service invoke a delete-many on the event store when a project is deleted.

Preferably, delete the project's events _first_ and _then_ the project. If we did it the other way around and there were a failure, we'd be left with orphaned events, but if we delete events first, we can avoid that.

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1254

PR #1254, opened 2021-01-29
feat(projects.go): delete all project events on project delete

vdice:

* Adds logic to delete all events associated with a given project when said project is deleted

Closes https://github.com/brigadecore/brigade/issues/1202

**If applicable**:

### DIFF
--- a/v2/apiserver/internal/core/projects.go
+++ b/v2/apiserver/internal/core/projects.go
@@ -155,16 +155,19 @@ type ProjectsService interface {
 
 type projectsService struct {
 	projectsStore ProjectsStore
+	eventsStore   EventsStore
 	substrate     Substrate
 }
 
 // NewProjectsService returns a specialized interface for managing Projects.
 func NewProjectsService(
 	projectsStore ProjectsStore,
+	eventsStore EventsStore,
 	substrate Substrate,
 ) ProjectsService {
 	return &projectsService{
 		projectsStore: projectsStore,
+		eventsStore:   eventsStore,
 		substrate:     substrate,
 	}
 }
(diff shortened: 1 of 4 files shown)


v2/apiserver/internal/core/projects.go
```diff
@@ -240,6 +243,36 @@ func (p *projectsService) Delete(ctx context.Context, id string) error {
...
+			return errors.Wrapf(err, "error deleting event %q from store", id)
+		}
+
+		if err = p.substrate.DeleteWorkerAndJobs(ctx, project, event); err != nil {
+			return errors.Wrapf(
+				err,
+				"error deleting event %q worker and jobs from the substrate",
+				id,
+			)
+		}
```
krancour, line 272: I think the call to `p.substrate.DeleteProject` on line 279 covers this. It delete's the project's namespace-- and that would take everything with it.

v2/apiserver/internal/core/projects.go
```diff
@@ -240,6 +243,36 @@ func (p *projectsService) Delete(ctx context.Context, id string) error {
...
+			"error listing events associated with project %q from store",
+			id,
+		)
+	}
+
+	// Delete all events associated with this project
+	for _, event := range eventList.Items {
+		if err = p.eventsStore.Delete(ctx, event.ID); err != nil {
+			return errors.Wrapf(err, "error deleting event %q from store", id)
+		}
```
krancour, line 264: Instead of iterating over all the events and deleting them individually, can we make a single call to `p.eventsStore.DeleteMany(...)`?

v2/apiserver/internal/core/projects.go
vdice: Yes indeed, thank you for pointing this out.  Updated.

v2/apiserver/internal/core/projects.go
vdice: You are right, thank you!  My thinking cap was MIA this Friday :)  Updated.

krancour (approved):
LGTM

Issue #1191, opened 2020-12-08
pending and running "child" jobs should get appropriate status updates when a running worker is aborted


krancour:
When a running worker is canceled, its status changes to aborted.

A running worker could have associated jobs that are either pending or running.

We _do_ already kill these on the substrate when the worker is canceled, but we're not currently doing any status updates on those jobs in the db to reflect that they were canceled (if pending) or aborted (if running).

This should be fixed.

vdice:
This is closed by https://github.com/brigadecore/brigade/pull/1275

PR #1262, opened 2021-02-03
feat(*): add logic to cancel jobs on event cancellation

vdice:

* Adds logic to cancel any/all worker jobs when their corresponding event is cancelled

Close https://github.com/brigadecore/brigade/issues/1191

**If applicable**:

### DIFF
--- a/v2/apiserver/internal/core/events.go
+++ b/v2/apiserver/internal/core/events.go
@@ -564,11 +564,15 @@ func (e *eventsService) CancelMany(
 	go func() {
 		for _, event := range events.Items {
 			for job := range event.Worker.Jobs {
-				if err = e.jobsStore.Cancel(
+				err = e.jobsStore.Cancel(
 					context.Background(),
 					event.ID,
 					job,
-				); err != nil {
+				)
+				if _, ok := errors.Cause(err).(*meta.ErrNotFound); ok {
+					continue
+				}
+				if err != nil {
 					log.Println(
 						errors.Wrapf(
 							err,
--- a/v2/apiserver/internal/core/events.go
+++ b/v2/apiserver/internal/core/events.go
@@ -237,6 +237,7 @@ type eventsService struct {
 	authorize           libAuthz.AuthorizeFn
 	projectsStore       ProjectsStore
 	eventsStore         EventsStore
+	jobsStore           JobsStore
 	substrate           Substrate
 	createSingleEventFn func(context.Context, Project, Event) (Event, error)
 }
@@ -246,12 +247,14 @@ func NewEventsService(
 	authorizeFn libAuthz.AuthorizeFn,
 	projectsStore ProjectsStore,
 	eventsStore EventsStore,
+	jobsStore JobsStore,
 	substrate Substrate,
 ) EventsService {
 	e := &eventsService{
 		authorize:     authorizeFn,
 		projectsStore: projectsStore,
 		eventsStore:   eventsStore,
+		jobsStore:     jobsStore,
 		substrate:     substrate,
 	}
 	e.createSingleEventFn = e.createSingleEvent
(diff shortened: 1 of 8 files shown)


v2/apiserver/internal/core/events.go
```diff
@@ -549,6 +563,27 @@ func (e *eventsService) CancelMany(
 	// efficient way to do this?
 	go func() {
 		for _, event := range events.Items {
+			// Only iterate through a worker's jobs if the worker status is not
+			// Pending; otherwise no jobs would have been created.
+			if event.Worker.Status.Phase != WorkerPhasePending {
+				for job := range event.Worker.Jobs {
+					if err = e.jobsStore.Cancel(
+						context.Background(),
```
krancour, line 571: Can you just add a comment here that it's deliberate that we're not using the `ctx` that was passed in?

v2/apiserver/internal/core/events.go
```diff
@@ -549,6 +563,27 @@ func (e *eventsService) CancelMany(
 	// efficient way to do this?
 	go func() {
 		for _, event := range events.Items {
+			// Only iterate through a worker's jobs if the worker status is not
+			// Pending; otherwise no jobs would have been created.
+			if event.Worker.Status.Phase != WorkerPhasePending {
```
krancour, line 568: I'd need to double-check the store code-- you may have already. Are we sure the events we're iterating over here have the _pre_-cancellation phases?

vdice: Oh great point... they may indeed be post-cancellation -- well, most assuredly since immediately prior to this go func, we issue the event cancellation 😂 

Let me look at what we can do.  May also try your suggestion of building complex db queries in the event cancellation functions to include job cancellations...

vdice: Ok, as it stands, we don't have a way to discern the phases of the original events (designated by the provided selector).  We could do so if we added a GetMany() (or similar) method to the events store to first get them and inspect, so we have this info by the time we run the loop.  Not sure if these added calls would be worth the exchange for not running the job cancellation loop if an event phase was pending (pre-cancellation)...

krancour:
Ok... before we proceed any further on this, I think we might need to put our heads together and come up with an improved approach because even the _existing_ cancel many (and delete many) logic has some problems (for which I am to blame).

The big problem is that if a delete or cancellation impacts a large number of events (_millions_ are a realistic possibility), then the `EventList` returned is enormous and could even blow out the heap.

Everywhere else in the entire API, we either deal with streams or we paginate results. Here we were not doing that. It's "cancel and give me back a list of everything affected."

Besides the potential for OOM, the "give me everything affected" approach is clearly not the correct approach, as we have uncovered. Events whose workers were running (and therefore may have created jobs) require more cleanup than events whose workers were pending. We need to be able to distinguish between those cases somehow.

Off the top of my head, I think we should consider something like this:

1. Event service calls cancel many on event store. Event store launches a goroutine and returns a _channel(s?)_ of events that require (different levels of?) further cleanup.
2. All events with pending workers can be canceled in one fell swoop. _Then_ we can get a scrollable cursor of those, iterate over them and stream them back to the service (over the channel). The service can invoke the substrate to perform cleanup. (The extent of cleanup required for events with pending workers would be deletion of some k8s secrets.)
3. Similar to no. 2, all events with starting or running workers can be canceled in one fell swoop. Again, we can get a scrollable cursor of those, iterate over them and stream them back to the service (over the channel). The service can invoke the substrate to perform cleanup. (The extent of cleanup is more extensive this time. It will be killing off initializing or running pods.) _Also_ at this juncture, we would need to iterate over all jobs belonging to such events and cancel them. The key thing here is that this WON'T be a giant loop. Unlike Brigade 1.x, Brigade 2's scheduler component limits the amount of concurrent workers (it's configurable). With this being the case, the max number of starting/running event workers should never be large enough to be of any real concern.

This is all off the top of my head. Take it all with a grain of salt. We're going to have to iterate on this a bit.

vdice:
With https://github.com/brigadecore/brigade/pull/1262#issuecomment-773420895 in mind, I'll close this PR and start on the work of streaming events/results on a cancelMany instead...

PR #1263, opened 2021-02-04
fix(integration tests): ignore error if of ErrNotFound variety on initial project delete

vdice:

- Only check for project delete error if not of the ErrNotFound variety.

**If applicable**:

### DIFF
--- a/v2/tests/git_initializer_test.go
+++ b/v2/tests/git_initializer_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/brigadecore/brigade/sdk/v2/core"
 	"github.com/brigadecore/brigade/sdk/v2/meta"
 	"github.com/brigadecore/brigade/sdk/v2/restmachinery"
+	"github.com/pkg/errors"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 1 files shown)


v2/tests/git_initializer_test.go
```diff
@@ -271,7 +272,9 @@ func TestMain(t *testing.T) {
 
 			// Delete the test project (we're sharing the name between tests)
 			err = client.Core().Projects().Delete(ctx, tc.project.ID)
-			require.NoError(t, err, "error deleting project")
+			if err, ok := errors.Cause(err).(*meta.ErrNotFound); !ok {
+				require.NoError(t, err, "error deleting project")
+			}
```
krancour, line 277: Non-blocking, but should we consider generating new, unique project names each time through the loop so that we don't have to deep six the project from the previous test case in order to create the project for the current test case?

As I said... non-blocking.

krancour (approved):
LGTM

v2/tests/git_initializer_test.go
vdice: Sure, I'd be in support of that change.  We could then reverse the order: create new test proj, run tests, delete test proj (perhaps optionalized)

krancour: Oh ya... I like that. I was thinking of skipping the project delete entirely, but you're right... we can have the best of both worlds-- clean up after ourselves _and_ eliminate the condition that cleanup of the previous test case must succeed for the current test case to proceed.

PR #1267, opened 2021-02-05
Add command aliases for CLI

### LINKED ISSUE #1216: CLI command naming/aliasing - singular/plural
A follow-up from https://github.com/brigadecore/brigade/pull/1214, as per @krancour : 

>I may actually be in favor of adding aliases for all resource-oriented commands so that the CLI is not at all picky about singular/plural. I'd want us to be consistent if we do that-- official commands singular across the board and aliases plural across the board. We can tackle that in a follow up. This still LGTM for now.

radu-matei:
closes #1216 

This commit adds plural aliases for the `brig` event, project, service accounts, and user commands.

### DIFF
--- a/v2/cli/event_commands.go
+++ b/v2/cli/event_commands.go
@@ -20,8 +20,9 @@ import (
 )
 
 var eventCommand = &cli.Command{
-	Name:  "event",
-	Usage: "Manage events",
+	Name:    "event",
+	Aliases: []string{"events"},
+	Usage:   "Manage events",
 	Subcommands: []*cli.Command{
 		{
 			Name:  "cancel",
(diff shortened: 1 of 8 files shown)


krancour:
This PR adds plural aliases for some singular nouns. There are two other nouns (roles and secrets) where the corresponding commands are plural with singular aliases.

I don't have any preference for whether we do singular with plural aliases or plural with singular aliases, but I do think we should be consistent across the board.

Any chance of amending this PR to make that consistent?

radu-matei:
I added aliases for the remaining commands and updated the source files to reflect the default command name.

krancour (approved):
LGTM

PR #1268, opened 2021-02-10
fix(git_initializer_test.go): don't overwrite err when checking cause

vdice:

😅 This is a fix for when we check the cause of the error that results from deleting a project.  We don't want to overwrite `err` here.

**If applicable**:


krancour (approved):
LGTM

PR #1269, opened 2021-02-10
rework, improve ci; add cd

krancour:
All the build/test and release automation we could ever want.


vdice (approved):
👏 Just in time for the first alpha release!  All looks excellent to me.

krancour:
@vdice we will have some additional secrets to add to the project now to enable:

1. Publishing the Helm chart to ghcr
1. Publishing Brigadier to npmjs

Upcoming release is v2.0.0-alpha.1... just realized I probably need to verify whether npmjs is tolerant of the alpha.1 after the dash... they may not be.

Will follow open a follow-up PR if necessary.

PR #1272, opened 2021-02-11
feat(events): stream canceled/deleted (many) events

vdice:

* Updates the Cancel/DeleteMany logic for events to stream instead of accrue in memory a potentially large collection of canceled/deleted events

See https://github.com/brigadecore/brigade/pull/1262#issuecomment-773420895

**If applicable**:

### DIFF
--- a/v2/apiserver/internal/core/events.go
+++ b/v2/apiserver/internal/core/events.go
@@ -551,22 +551,20 @@ func (e *eventsService) CancelMany(
 	for i := 0; i < concurrency; i++ {
 		go func() {
 			for {
-				select {
-				case event, ok := <-eventCh:
-					if !ok { // event channel closed; we're done
-						return
-					}
-					if err := e.substrate.DeleteWorkerAndJobs(
-						context.Background(), // deliberately not using ctx
-						project,
-						event,
-					); err != nil {
-						log.Println(errors.Wrapf(
-							err,
-							"error deleting event %q worker and jobs from the substrate",
-							event.ID,
-						))
-					}
+				event, ok := <-eventCh
+				if !ok { // event channel closed; we're done
+					return
+				}
+				if err := e.substrate.DeleteWorkerAndJobs(
+					context.Background(), // deliberately not using ctx
+					project,
+					event,
+				); err != nil {
+					log.Println(errors.Wrapf(
+						err,
+						"error deleting event %q worker and jobs from the substrate",
+						event.ID,
+					))
 				}
 			}
 		}()
@@ -658,23 +656,20 @@ func (e *eventsService) DeleteMany(
 	for i := 0; i < concurrency; i++ {
 		go func() {
 			for {
-				select {
-				case event, ok := <-eventCh:
-					if ok {
-						if err := e.substrate.DeleteWorkerAndJobs(
-							context.Background(), // deliberately not using ctx
-							project,
-							event,
-						); err != nil {
-							log.Println(errors.Wrapf(
-								err,
-								"error deleting event %q worker and jobs from the substrate",
-								event.ID,
-							))
-						}
-					} else { // event channel closed; we're done
-						return
-					}
+				event, ok := <-eventCh
+				if !ok { // event channel closed; we're done
+					return
+				}
+				if err := e.substrate.DeleteWorkerAndJobs(
+					context.Background(), // deliberately not using ctx
+					project,
+					event,
+				); err != nil {
+					log.Println(errors.Wrapf(
+						err,
+						"error deleting event %q worker and jobs from the substrate",
+						event.ID,
+					))
 				}
 			}
 		}()
--- a/v2/apiserver/internal/core/events.go
+++ b/v2/apiserver/internal/core/events.go
@@ -537,41 +537,43 @@ func (e *eventsService) CancelMany(
 		)
 	}
 
-	eventCh, err := e.eventsStore.CancelMany(ctx, selector)
+	eventCh, count, err := e.eventsStore.CancelMany(ctx, selector)
 	if err != nil {
 		return result, errors.Wrap(err, "error canceling events in store")
 	}
+	result.Count = count
 
-	// When an event comes in via the corresponding channel,
-	// kick off a go routine to execute cleanup.
-	for {
-		select {
-		case event, ok := <-eventCh:
-			if ok {
-				result.Count++
-				// TODO: This could take a while, so we don't do it synchronously. But
-				// what if the process dies while this is in-progress? Can we find a
-				// quicker, more efficient way to do this?
-				go func() {
-					if err := e.substrate.DeleteWorkerAndJobs(
-						context.Background(), // deliberately not using ctx
-						project,
-						event,
-					); err != nil {
-						log.Println(errors.Wrapf(
-							err,
-							"error deleting event %q worker and jobs from the substrate",
-							event.ID,
-						))
-					}
-				}()
-			} else { // event channel closed; we're done
-				return result, nil
-			}
-		case <-ctx.Done():
-			return result, nil
-		}
+	// Fan out to a finite number of goroutines to handle cleanup duties
+	concurrency := 10
+	if result.Count < int64(concurrency) {
+		concurrency = int(result.Count)
 	}
+	for i := 0; i < concurrency; i++ {
+		go func() {
+			for {
+				select {
+				case event, ok := <-eventCh:
+					if ok {
+						if err := e.substrate.DeleteWorkerAndJobs(
+							context.Background(), // deliberately not using ctx
+							project,
+							event,
+						); err != nil {
+							log.Println(errors.Wrapf(
+								err,
+								"error deleting event %q worker and jobs from the substrate",
+								event.ID,
+							))
+						}
+					} else { // event channel closed; we're done
+						return
+					}
+				}
+			}
+		}()
+	}
+
+	return result, nil
 }
 
 func (e *eventsService) Delete(ctx context.Context, id string) error {
@@ -643,41 +645,43 @@ func (e *eventsService) DeleteMany(
 		)
 	}
 
-	eventCh, err := e.eventsStore.DeleteMany(ctx, selector)
+	eventCh, count, err := e.eventsStore.DeleteMany(ctx, selector)
 	if err != nil {
 		return result, errors.Wrap(err, "error deleting events from store")
 	}
+	result.Count = count
 
-	// When an event comes in via the corresponding channel,
-	// kick off a go routine to execute cleanup.
-	for {
-		select {
-		case event, ok := <-eventCh:
-			if ok {
-				result.Count++
-				// TODO: This could take a while, so we don't do it synchronously. But
-				// what if the process dies while this is in-progress? Can we find a
-				// quicker, more efficient way to do this?
-				go func() {
-					if err := e.substrate.DeleteWorkerAndJobs(
-						context.Background(), // deliberately not using ctx
-						project,
-						event,
-					); err != nil {
-						log.Println(errors.Wrapf(
-							err,
-							"error deleting event %q worker and jobs from the substrate",
-							event.ID,
-						))
-					}
-				}()
-			} else { // event channel closed; we're done
-				return result, nil
-			}
-		case <-ctx.Done():
-			return result, nil
-		}
+	// Fan out to a finite number of goroutines to handle cleanup duties
+	concurrency := 10
+	if result.Count < int64(concurrency) {
+		concurrency = int(result.Count)
 	}
+	for i := 0; i < concurrency; i++ {
+		go func() {
+			for {
+				select {
+				case event, ok := <-eventCh:
+					if ok {
+						if err := e.substrate.DeleteWorkerAndJobs(
+							context.Background(), // deliberately not using ctx
+							project,
+							event,
+						); err != nil {
+							log.Println(errors.Wrapf(
+								err,
+								"error deleting event %q worker and jobs from the substrate",
+								event.ID,
+							))
+						}
+					} else { // event channel closed; we're done
+						return
+					}
+				}
+			}
+		}()
+	}
+
+	return result, nil
 }
 
 // EventsStore is an interface for components that implement Event persistence
@@ -714,12 +718,12 @@ type EventsStore interface {
 	// parameter in the underlying data store to reflect that they have been
 	// canceled. Implementations MUST only cancel events whose Workers have not
 	// already reached a terminal state.
-	CancelMany(context.Context, EventsSelector) (<-chan Event, error)
+	CancelMany(context.Context, EventsSelector) (<-chan Event, int64, error)
 	// Delete unconditionally deletes the specified Event from the underlying data
 	// store. If the specified Event does not exist, implementations MUST
 	// return a *meta.ErrNotFound error.
 	Delete(context.Context, string) error
 	// DeleteMany unconditionally deletes multiple Events specified by the
 	// EventsSelector parameter from the underlying data store.
-	DeleteMany(context.Context, EventsSelector) (<-chan Event, error)
+	DeleteMany(context.Context, EventsSelector) (<-chan Event, int64, error)
 }
--- a/v2/apiserver/internal/core/events.go
+++ b/v2/apiserver/internal/core/events.go
@@ -550,11 +550,7 @@ func (e *eventsService) CancelMany(
 	}
 	for i := 0; i < concurrency; i++ {
 		go func() {
-			for {
-				event, ok := <-eventCh
-				if !ok { // event channel closed; we're done
-					return
-				}
+			for event := range eventCh {
 				if err := e.substrate.DeleteWorkerAndJobs(
 					context.Background(), // deliberately not using ctx
 					project,
@@ -655,11 +651,7 @@ func (e *eventsService) DeleteMany(
 	}
 	for i := 0; i < concurrency; i++ {
 		go func() {
-			for {
-				event, ok := <-eventCh
-				if !ok { // event channel closed; we're done
-					return
-				}
+			for event := range eventCh {
 				if err := e.substrate.DeleteWorkerAndJobs(
 					context.Background(), // deliberately not using ctx
 					project,
--- a/v2/apiserver/internal/core/events.go
+++ b/v2/apiserver/internal/core/events.go
@@ -553,20 +553,19 @@ func (e *eventsService) CancelMany(
 			for {
 				select {
 				case event, ok := <-eventCh:
-					if ok {
-						if err := e.substrate.DeleteWorkerAndJobs(
-							context.Background(), // deliberately not using ctx
-							project,
-							event,
-						); err != nil {
-							log.Println(errors.Wrapf(
-								err,
-								"error deleting event %q worker and jobs from the substrate",
-								event.ID,
-							))
-						}
-					} else { // event channel closed; we're done
+					if !ok { // event channel closed; we're done
 						return
+					}
+					if err := e.substrate.DeleteWorkerAndJobs(
+						context.Background(), // deliberately not using ctx
+						project,
+						event,
+					); err != nil {
+						log.Println(errors.Wrapf(
+							err,
+							"error deleting event %q worker and jobs from the substrate",
+							event.ID,
+						))
 					}
 				}
 			}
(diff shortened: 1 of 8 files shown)


v2/apiserver/internal/core/mongodb/events_store.go
```diff
@@ -260,7 +261,7 @@ func (e *eventsStore) CancelMany(
 
 	// Bail if we're not canceling pending, starting, or running events
 	if !cancelPending && !cancelStarting && !cancelRunning {
-		return events, nil
+		return eventCh, errCh, nil
```
krancour, line 264: Nothing will ever close these channels, so whoever's reading from these channels will get stuck forever.

You probably want to return nils here.

v2/apiserver/internal/core/mongodb/events_store.go
```diff
@@ -285,7 +286,7 @@ func (e *eventsStore) CancelMany(
 				},
 			},
 		); err != nil {
-			return events, errors.Wrap(err, "error updating events")
+			return eventCh, errCh, errors.Wrap(err, "error updating events")
```
krancour, line 289: It would probably be more clean to return nils here instead of open channels. Same in a few other error handling conditionals below.

v2/apiserver/internal/core/events.go
```diff
@@ -537,35 +537,59 @@ func (e *eventsService) CancelMany(
...
+	for {
+		select {
+		case event, ok := <-eventCh:
+			if ok {
+				result.Count++
+				// TODO: This could take a while, so we don't do it synchronously. But
+				// what if the process dies while this is in-progress? Can we find a
+				// quicker, more efficient way to do this?
+				go func() {
+					if err := e.substrate.DeleteWorkerAndJobs(
```
krancour, line 558: I was about to propose doing this conditionally-- only if the event's worker's current phase is `ABORTED`. If the current phase were `CANCELED`, that worker never made it past `PENDING` and cannot possibly have any pods associated with it...

Then I remembered that there's still a secret associated with every pending event that still needs to be removed upon cancellation.

But what we could do to speed up the whole process is do that phase in the substrate's `DeleteWorkerAndJobs()` function. If canceled, only delete secrets associated with the event. If aborted, delete associated secrets _and_ pods.

v2/apiserver/internal/core/events.go
```diff
@@ -537,35 +537,59 @@ func (e *eventsService) CancelMany(
...
+				}()
+			} else {
+				// eventCh was closed, but want to keep looping through this select
+				// in case there are pending errors on the errCh still. nil channels are
+				// never readable, so we'll just nil out eventCh and move on.
+				eventCh = nil
+			}
+		case err, ok := <-errCh:
+			if ok {
+				return result, err
```
krancour, line 578: I'm not sure we should be returning here. Do we really want to stop all further attempts at cleaning up the substrate because there was one error?

Might be better to just log it.

If we're only going to log it, let's log the error in the store and get rid of the `errCh` entirely.

If we get rid of the `errCh`, this `select` becomes a lot simpler because you don't have to go to lengths to exhaust both channels before exiting the loop.

krancour:
I think `DeleteMany()` requires the same optimizations that have been applied to `CancelMany()`.

v2/apiserver/internal/core/events.go
krancour: I'm fine with that being a in a follow-up, but it could also make sense to tack it onto this PR...

Dealer's choice.

vdice: Good call; it appeared simple enough to include.  Does https://github.com/brigadecore/brigade/pull/1272/commits/2daed15b0d954d210e74b090fcfd08c05ae2d0d6 look good?

v2/apiserver/internal/core/events.go
vdice: Thanks, updated.

vdice:
> I think DeleteMany() requires the same optimizations that have been applied to CancelMany().

Ah yes, thank you.  Updated in https://github.com/brigadecore/brigade/pull/1272/commits/a2de8faf48628948bdcab33573f587769e344e7b

v2/apiserver/internal/core/events.go
```diff
@@ -537,35 +537,41 @@ func (e *eventsService) CancelMany(
...
-				log.Println(
-					errors.Wrapf(
-						err,
-						"error deleting event %q worker and jobs from the substrate",
-						event.ID,
-					),
-				)
+	// When an event comes in via the corresponding channel,
+	// kick off a go routine to execute cleanup.
+	for {
```
krancour, line 547: I think this whole loop should be in its own goroutine, otherwise, we could be here a while for a large number of cancelations. We may be launching a goroutine for _each_ cancelation (more on that below), but for say a million cancelations, we'd still be waiting around while we drain 1m events from the channel and launch those goroutines. That can be time consuming in and of itself.

v2/apiserver/internal/core/events.go
```diff
@@ -537,35 +537,41 @@ func (e *eventsService) CancelMany(
...
+		select {
+		case event, ok := <-eventCh:
+			if ok {
+				result.Count++
+				// TODO: This could take a while, so we don't do it synchronously. But
+				// what if the process dies while this is in-progress? Can we find a
+				// quicker, more efficient way to do this?
+				go func() {
```
krancour, line 555: It looks like we have the potential to launch an unbounded number of goroutines. It's possible and practical to launch very large numbers of goroutines (I think the Go docs say tens of thousands is practical), but there's the potential here to be dealing with much larger numbers. And each of these goroutines closes over an event that was read off the channel. So if we fire up 1m goroutines, we have 1m events-- we're back to a condition that could blow out the heap.

What if we did something like this instead:

```
eventCh, err := e.eventsStore.CancelMany(ctx, selector)
if err != nil {
  return result, errors.Wrap(err, "error canceling events in store")
}
// Fan out to a finite number of goroutines
for (i := 0; i < 10; i++) {
  go func() {
    for {
      select {
        // ... cleanup ...
      }
    }
  }()
}
```

v2/apiserver/internal/core/events.go
```diff
@@ -537,35 +537,41 @@ func (e *eventsService) CancelMany(
...
+							"error deleting event %q worker and jobs from the substrate",
+							event.ID,
+						))
+					}
+				}()
+			} else { // event channel closed; we're done
+				return result, nil
 			}
+		case <-ctx.Done():
+			return result, nil
```
krancour, line 572: This context is ultimately tied to the inbound http request. If the client hangs up, cleanup of the substrate will abort... probably not what we want.

v2/apiserver/internal/core/events.go
```diff
@@ -637,35 +643,41 @@ func (e *eventsService) DeleteMany(
...
-				event,
-			); err != nil {
-				log.Println(
-					errors.Wrapf(
-						err,
-						"error deleting event %q worker and jobs from the substrate",
-						event.ID,
-					),
-				)
+	// When an event comes in via the corresponding channel,
```
krancour, line 651: Comments on cancelation logic equally applicable to delete logic.

v2/apiserver/internal/core/kubernetes/substrate.go
```diff
@@ -643,40 +643,45 @@ func (s *substrate) DeleteWorkerAndJobs(
...
-		},
-	); err != nil {
-		return errors.Wrapf(
-			err,
-			"error deleting event %q pods in namespace %q",
-			event.ID,
+	// If a worker's phase is CANCELED, it could have only reached this terminal
+	// phase if it was previously in PENDING and hence, no pods would have been
+	// created. Therefore, we just skip to cleaning up the event secret(s) below.
+	if event.Worker.Status.Phase != core.WorkerPhaseCanceled {
```
krancour, line 649: 👍

v2/apiserver/internal/core/mongodb/events_store.go
```diff
@@ -370,13 +371,25 @@ func (e *eventsStore) CancelMany(
...
+	go func() {
+		defer close(eventCh)
+		for cur.Next(ctx) {
+			event := core.Event{}
+			if err := cur.Decode(&event); err != nil {
+				log.Println(errors.Wrap(err, "unable to decode event"))
+			}
+			select {
+			case eventCh <- event:
+			case <-ctx.Done():
```
krancour, line 387: Another case where this context is tied to the inbound http request. If the client hangs up, cleanup will abort.

v2/apiserver/internal/core/mongodb/events_store.go
```diff
@@ -446,22 +457,32 @@ func (e *eventsStore) DeleteMany(
...
+				return
+			}
+		}
+	}()
 
 	// Final deletion
 	if _, err := e.collection.DeleteMany(ctx, criteria); err != nil {
-		return events, errors.Wrap(err, "error deleting events")
+		return eventCh, errors.Wrap(err, "error deleting events")
 	}
```
krancour, line 485: This final deletion can happen while we're still iterating (goroutine above) over the things that were _slated_ for deletion. Those two concerns are going to race with each other and yield funky results.

v2/apiserver/internal/core/events.go
vdice: The goroutine batch approach seems fine... though, say we have just one canceled event to cleanup after... we'll still launch 10 routines (in this example), right?  Would the other 9 routines (quickly returning, admittedly) incur any concerning overhead?

vdice: Playing around with things now... realizing we also have a need to synchronously tally the canceled/deleted event count and hand this back to the user, which wouldn't work if we increment in a go routine.  Thinking how we might still 'batch' go routines with this in mind...

krancour: There's an atomic counter doohickey in the atd lib's sync package, I think.

vdice: That would segue nicely into adding a `-w/--wait` option for these commands, since we'd have to wait for all routines to finish to return the final tally.  Lemme just see if it's feasible to do so w/o waiting...

krancour: As for whether it's overkill to launch n goroutimes to clean up just one event... we have no way of knowing headed into the loop that we're only cleaning up after m < n events. Thankfully the amount of overhead should be pretty low. The excess goroutines should idle almost immediately and then exit very shortly after that.

Also, we _could_ modify the store code to return the count of affected events along with the channel. That means going into the loop we would know how many events we expect to clean up after and we could use that info to set a floor on the number of goroutines.... it would also stop us from having to mess with an atomic counter.

vdice: Ohhh great point re: getting the count from the store.  I like that.  That's probably the best way to go.

vdice: Ok, updated in https://github.com/brigadecore/brigade/pull/1272/commits/dd20cdc7ff22a4bbaa3522f71fc585774bf7062f (apologies, forgot to group commits by scope).  PTAL when convenient.

v2/apiserver/internal/core/events.go
```diff
@@ -537,33 +537,41 @@ func (e *eventsService) CancelMany(
 		)
 	}
 
-	events, err := e.eventsStore.CancelMany(ctx, selector)
+	eventCh, count, err := e.eventsStore.CancelMany(ctx, selector)
```
krancour, line 540: Non-blocking nit: Can we change `count` to `affectedCount` or something so that what it represents is more obvious?

v2/apiserver/internal/core/events.go
```diff
@@ -537,33 +537,41 @@ func (e *eventsService) CancelMany(
...
-						event.ID,
-					),
-				)
+	result.Count = count
+
+	// Fan out to a finite number of goroutines to handle cleanup duties
+	concurrency := 10
+	if result.Count < int64(concurrency) {
+		concurrency = int(result.Count)
+	}
```
krancour, line 550: 👍

v2/apiserver/internal/core/events.go
```diff
@@ -637,33 +645,41 @@ func (e *eventsService) DeleteMany(
 		)
 	}
 
-	events, err := e.eventsStore.DeleteMany(ctx, selector)
+	eventCh, count, err := e.eventsStore.DeleteMany(ctx, selector)
```
krancour, line 648: Same non-blocking comment as [here](https://github.com/brigadecore/brigade/pull/1272/files#r576925304)

v2/apiserver/internal/core/events.go
```diff
@@ -702,18 +718,12 @@ type EventsStore interface {
 	// parameter in the underlying data store to reflect that they have been
 	// canceled. Implementations MUST only cancel events whose Workers have not
 	// already reached a terminal state.
```
krancour, line 720: Can you amend this comment to reflect that implementations MUST return a count of affected Events?

v2/apiserver/internal/core/events.go
```diff
@@ -702,18 +718,12 @@ type EventsStore interface {
...
-		context.Context,
-		EventsSelector,
-	) (EventList, error)
+	CancelMany(context.Context, EventsSelector) (<-chan Event, int64, error)
 	// Delete unconditionally deletes the specified Event from the underlying data
 	// store. If the specified Event does not exist, implementations MUST
 	// return a *meta.ErrNotFound error.
 	Delete(context.Context, string) error
 	// DeleteMany unconditionally deletes multiple Events specified by the
 	// EventsSelector parameter from the underlying data store.
```
krancour, line 727: Same comment as [here](https://github.com/brigadecore/brigade/pull/1272/files#r576930336)

v2/apiserver/internal/core/kubernetes/substrate.go
krancour: Also true for an Event whose Worker is _still_ in a PENDING phase. That could happen if `DeleteWorkerAndJobs` was invoked in response to a `Delete` or `DeleteMany` on the `EventService`.

v2/apiserver/internal/core/mongodb/events_store.go
```diff
@@ -368,15 +369,30 @@ func (e *eventsStore) CancelMany(
 			{Key: "created", Value: -1},
 		},
 	)
-	cur, err := e.collection.Find(ctx, criteria, findOptions)
+	count, err := e.collection.CountDocuments(ctx, criteria)
```
krancour, line 372: We can get the count without making an extra database call.

In the one (or, conditionally, two) updates applied above, there's a result object we're currently ignoring that can give us a count of affected documents.

v2/apiserver/internal/core/mongodb/events_store.go
```diff
@@ -368,15 +369,30 @@ func (e *eventsStore) CancelMany(
...
+
+	eventCh := make(chan core.Event)
+	go func() {
+		defer close(eventCh)
+		for cur.Next(ctx) {
+			event := core.Event{}
+			if err := cur.Decode(&event); err != nil {
+				log.Println(errors.Wrap(err, "error decoding event"))
+			}
+			select {
```
krancour, line 390: No point using a select here.

v2/apiserver/internal/core/mongodb/events_store.go
```diff
@@ -444,24 +458,39 @@ func (e *eventsStore) DeleteMany(
 			{Key: "created", Value: -1},
 		},
 	)
+	count, err := e.collection.CountDocuments(ctx, criteria)
```
krancour, line 461: Same comment as [here](https://github.com/brigadecore/brigade/pull/1272/files#r576936375)

v2/apiserver/internal/core/mongodb/events_store.go
```diff
@@ -444,24 +458,39 @@ func (e *eventsStore) DeleteMany(
...
-	}
+	eventCh := make(chan core.Event)
+	go func() {
+		defer close(eventCh)
+		for cur.Next(ctx) {
+			event := core.Event{}
+			if err := cur.Decode(&event); err != nil {
+				log.Println(errors.Wrap(err, "error decoding event"))
+			}
+			select {
```
krancour, line 482: No point using a select here.

v2/apiserver/internal/core/mongodb/events_store_test.go
```diff
@@ -615,13 +614,45 @@ func TestEventsStoreCancelMany(t *testing.T) {
...
-			assertions: func(events core.EventList, err error) {
+			assertions: func(err error) {
 				require.Error(t, err)
 				require.Contains(t, err.Error(), "error updating events")
 				require.Contains(t, err.Error(), "something went wrong")
 			},
 		},
 
+		{
+			name: "error counting canceled events",
```
krancour, line 625: This whole test case can go away if we don't make the extra db call to count affected events.

v2/apiserver/internal/core/mongodb/events_store_test.go
```diff
@@ -799,15 +844,15 @@ func TestEventsStoreDeleteMany(t *testing.T) {
...
+			assertions: func(err error) {
 				require.Error(t, err)
 				require.Contains(t, err.Error(), "error logically deleting events")
 				require.Contains(t, err.Error(), "something went wrong")
 			},
 		},
 
 		{
-			name: "error finding deleted events",
+			name: "error counting deleted events",
```
krancour, line 855: This whole test case can go away if we don't make the extra db call to count affected events.

vdice:
Thank you @krancour !  I think I've now addressed all of your feedback.  PTAL when convenient.

v2/apiserver/internal/core/events.go
```diff
@@ -537,33 +537,41 @@ func (e *eventsService) CancelMany(
...
+	concurrency := 10
+	if result.Count < int64(concurrency) {
+		concurrency = int(result.Count)
+	}
+	for i := 0; i < concurrency; i++ {
+		go func() {
+			for {
+				select {
+				case event, ok := <-eventCh:
+					if ok {
```
krancour, line 556: Nit: If we lead with the following, everything that follows it is indented one level less, which is generally considered to reduce the complexity.

```
if !ok {
  return
}
```

v2/apiserver/internal/core/events.go
```diff
@@ -637,33 +644,41 @@ func (e *eventsService) DeleteMany(
...
```
krancour, line 663: This conditional can still be simplified with an early return.

v2/apiserver/internal/core/events.go
```diff
@@ -637,33 +644,41 @@ func (e *eventsService) DeleteMany(
...
+
+	// Fan out to a finite number of goroutines to handle cleanup duties
+	concurrency := 10
+	if result.Count < int64(concurrency) {
+		concurrency = int(result.Count)
+	}
+	for i := 0; i < concurrency; i++ {
+		go func() {
+			for {
+				select {
```
krancour, line 661: This select isn't doing anything.

v2/apiserver/internal/core/events.go
```diff
@@ -537,33 +537,40 @@ func (e *eventsService) CancelMany(
...
```
krancour, line 554: This select isn't doing anything.

v2/apiserver/internal/core/events.go
```diff
@@ -637,33 +642,38 @@ func (e *eventsService) DeleteMany(
...
+	if result.Count < int64(concurrency) {
+		concurrency = int(result.Count)
+	}
+	for i := 0; i < concurrency; i++ {
+		go func() {
+			for {
+				event, ok := <-eventCh
+				if !ok { // event channel closed; we're done
+					return
+				}
```
krancour, line 662: Nit: These five lines can be condensed to this. (I wasn't sure at first, but I tested it. You never get an event that is a zero value `Event{}`.)

```
for event := range  eventCh {
```

Could be applied in one other spot as well.

krancour: _But_ this is totally non-blocking. LGTM!

krancour:
I know a _lot_ of hard work went into this one. LGTM @vdice. Thank you.

krancour (approved):

v2/apiserver/internal/core/events.go
vdice: Oooh nice, thanks for testing and proposing.  Implemented.

krancour:
LGTM stands

vdice:
Thanks for all your help refining these changes into a much cleaner and more efficient approach @krancour !

Issue #1270, opened 2021-02-11
Don't use tailable cursor when streaming logs from cool storage


krancour:
Our cool log storage currently uses a tailable cursor. This was necessary at one time, because before we divided the logs into warm (k8s) and cool (mongo) storage, we could be streaming new log lines _out_ of mongo as they were written _to_ mongo by the log agent.

This can't happen anymore, because we only ever stream logs out of cool storage when warm storage doesn't exist anymore (i.e. the worker or job pod has been cleaned up and no longer exists). That means that for any given worker or job, every log line that gets stashed in cool storage is there _before_ we ever start streaming from there.

If we ditch the tailable cursor, we can also ditch the requirement that the underlying log collection be a capped collection. _This is the real improvement we want._ With that assumption undone, we clear one impediment (possibly of several) to utilizing cosmos as a drop-in replacement for mongo. Cosmos doesn't support capped collections.

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1274

(Spoiler alert around the note on the capped collection for logs -- we're intending to indeed move away from this type in the near future, most likely in a PR to close https://github.com/brigadecore/brigade/issues/1271.)

PR #1274, opened 2021-02-11
ref(logs_store.go): rm tailable cursor option

vdice:

* Removes the tailable cursor option when streaming logs from the cool store (mongodb), as the log service will only ever fetch from this store when all logs have been persisted from workers/jobs.
* Ensure the warm logs hang up (without error) after all logs have been streamed from the pod and the pod's phase is terminal.

Closes https://github.com/brigadecore/brigade/issues/1270

**If applicable**:

### DIFF
--- a/v2/apiserver/internal/core/kubernetes/logs_store.go
+++ b/v2/apiserver/internal/core/kubernetes/logs_store.go
@@ -106,20 +106,6 @@ func (l *logsStore) StreamLogs(
 			}
 		}
 		podLogs.Close()
-		if opts.Follow {
-			// If following, we let this goroutine hang until the context times out or
-			// is canceled. Why? When logs are followed from COLD storage (e.g.
-			// MongoDB) we never know whether the log aggregator has stored all the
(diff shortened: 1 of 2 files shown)


krancour:
This looks great.

I do want to request one change.

After this change, the cool logs do not respect the follow option anymore. I know, I know... there's nothing _to_ follow, but it means a command like `brig event logs -i <event id> -f` ends up behaving differently depending on whether the logs come from the warm logs or cool logs. The behavior of one vs the other should be indistinguishable to the user.

So there are two ways to bring the follow behavior of the warm and cool log stores back into sync...

1. "Honor" the follow option in the cool logs by simply not hanging up until the context is canceled.

2. Ensure the warm logs hang up (without error) after all logs have been streamed from the pod _and_ the pod's phase is terminal.

No. 2 seems preferable to me.

vdice:
@krancour Ok, updated to allow warm/k8s logs to exit when all entries are processed (even w/ follow true).

krancour (approved):
LGTM!

PR #1275, opened 2021-02-11
use array to store worker's jobs instead of map

### LINKED ISSUE #1191: pending and running "child" jobs should get appropriate status updates when a running worker is aborted

krancour:
Fixes #1191 

The main challenge to cascading cancelations was to do it efficiently. Transitioning from workers having a map of job names --> job to workers having an _array_ of jobs with intrinsic names (the were extrinsic before) allows us to cascade cancellations much more efficiently on the database end.

The side effect of this change is that jobs associated with a given worker are an ordered list now instead of an unordered map. I initially I had preferred the map because map semantics enforced unique job names and didn't directly imply execution order. The downside to the map, however, was that if retrieving and displaying an event multiple times, the order of the jobs appears to shift around. `watch brig event get -i <id>` was really triggering my OCD. After cutting over to an array, the order of jobs in the list is the order in which the jobs were created.... I think that actually makes a lot of sense.

### DIFF
--- a/v2/apiserver/internal/core/jobs.go
+++ b/v2/apiserver/internal/core/jobs.go
@@ -73,6 +73,8 @@ func (j JobPhase) IsTerminal() bool {
 // Job represents a component spawned by a Worker to complete a single task
 // in the course of handling an Event.
 type Job struct {
+	// Name is the Job's name. It should be unique among a given Worker's Jobs.
+	Name string `json:"name" bson:"name"`
 	// Spec is the technical blueprint for the Job.
 	Spec JobSpec `json:"spec" bson:"spec"`
 	// Status contains details of the Job's current state.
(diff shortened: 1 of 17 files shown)


v2/apiserver/internal/core/jobs.go
```diff
@@ -522,8 +515,9 @@ func (j *jobsService) WatchStatus(
 				log.Printf("error retrieving event %q from store: %s", eventID, err)
 				return
 			}
+			job, _ := event.Worker.Job(jobName)
```
krancour, line 518: We're not checking if the job exists. We know it does. We checked that well before we entered this loop.

vdice (approved):
Nice.  I was anticipating this transition would simplify things and looking at the code changes, I'm definitely happy w/ the decision.  Especially considering the relatively straightforward mongo criteria/opts on update/cancel/etc.  LGTM!

PR #1278, opened 2021-02-12
simplify brigadier logger; eliminate brigadier's third-party dependendecnies

krancour:
In preparation for the upcoming release, I did a rehearsal of publishing brigadier to npmjs.com today. It went smoothly, but when I went to _use_ it in a project, I noticed it dragged along quite a large number of third party dependencies with it. They were all transitive dependencies of winston-- the leveled logging library we're using.

This PR cuts us over to the same approach used for jobs, groups, etc. Brigadier provides a minimal (and more or less no-op) implementation with no third-party dependencies. At runtime, the _worker_ polyfills the logger with a more sophisticated winston logger.

This brings the third-party dependencies for brigadier down to zero.

### DIFF
--- a/v2/brigadier/package.json
+++ b/v2/brigadier/package.json
@@ -34,8 +34,5 @@
     "mocha": "^8.2.1",
     "ts-node": "^9.1.1",
     "typescript": "^4.1.2"
-  },
-  "dependencies": {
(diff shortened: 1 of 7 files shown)


vdice (approved):
Nice!

PR #1281, opened 2021-02-16
change api version to be v2-alpha.1

krancour:
I feel as if this CYA maneuver is in order before we cut an alpha release. This way we safeguard ourselves against anyone who may come along months from now trying to use an alpha.1 SDK or CLI with a newer API server that may have taken on breaking changes in the interim.

cc @radu-matei you'd want to account for this in your Rust SDK. I'll handle the JS/TS one.

### DIFF
--- a/examples/git/github-demo.yaml
+++ b/examples/git/github-demo.yaml
@@ -1,4 +1,4 @@
-apiVersion: brigade.sh/v2
+apiVersion: brigade.sh/v2-alpha.1
 kind: Project
 metadata:
   id: github-demo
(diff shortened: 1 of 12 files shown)


vdice (commented):
I understand the motivation for this change but wanted to clarify what our intentions are as we work towards the official v2 release.  Would we issue these same PRs for each alpha and beta release and then once we're ready for the final v2.0.0 release, change apiVersion back to `brigade.sh/v2`, with no intentions to update for the remainder of the lifespan of v2 (e.g. no updating for any/all minor and patch releases)?

krancour:
@vdice what you described is pretty much exactly what I'd been thinking. That would preserve our perogative to make breaking changes, sparingly and judiciously, until a fornal v2 release, at which point we will guarantee backwards compatibility for the life of 2.x.

vdice (approved):

PR #1282, opened 2021-02-17
Streamline brigadier publishing

krancour:
This is a faster, more efficient way to publish brigadier to npmjs.com without a need for any third-party tools.

This approach is _already_ employed by the 1.x brigadier.


technosophos (approved):
LGTM

PR #1290, opened 2021-02-19
[WIP] Replace groups with composite jobs

itowlson:

The current Brigadier library allows you to run a set of jobs in parallel or in sequence using the `Group` class.  However, this restricts composition to a single level.  For example, this does not support running a set of pipelines in parallel where each pipeline is sequential.  Of course users can code this up themselves using JavaScript (just as they could code up groups), but it would be nice to build this in.

This PR therefore proposes that jobs can be composed to create new jobs.  For example, the `Job.sequence(...)` combinator produces a Job which, when run, runs all its sub-jobs in sequence, e.g.:

```javascript
function renderFrame(source) {
  return Job.sequence([
    Job.container(`triangulate-${source}`, "blender/triangles:v0.9.0"),
    Job.container(`reticulate-${source}`, "itowlson/reticulate:v1.3.0"),
    Job.container(`render-${source}`, "blender/render:v0.8.0")
  ]);
}
```

Because the result of this is a Job, it can itself be passed to the `Job.parallel` combinator.  The result of this is a more composable API surface for job definition.

```
function renderAllFrames(sources) {
  const pipelines = sources.map((source) => renderFrames(source));
  return Job.parallel(pipelines);
}

on("render", (e) => {
  Job.sequence([
    renderAllFrames(e.sources),
    stitchFramesToMovie()
  ]).run();
});
```

As an exploration of how this approach might enable further idioms, the PR also includes a `Job.retryable` combinator which produces a job that attempts a sub-job but retries if it fails.

This is a proposal for feedback (and I can only apologise that it has come so late in the 2.x release cycle.)

1. At this stage the Group class is kept but marked as deprecated.  This maximises backward compatibility.

2. I tentatively used Job for the base/factory type and renamed the current Job class to ContainerJob.  This blows backward compatibility out of the water and I am concerned it could cause problems, especially for JavaScript users who will not get it typechecked.  A safer approach might be to declare a `Runnable` interface - the combinators would then compose Runnables into Runnables, and container jobs could go back to being called Job.  But an interface cannot contain static methods, so what would we call the class where those factory methods lived?  (Assuming we even like factory methods - I feel `Job.sequence(...)` is nicer than `new SequentialJob(...)` but others might disagree?)

3. We could also consider dot-notation syntax for some of this e.g. `job1.followedBy(job2)`, `job1.inParallelWith(job2)`, `job.withRetry(5)`.  I think the combinators work better on the whole, but very much open to discussion.

4. I updated the unit tests and added some JSDoc, but I haven't updated the samples or other conceptual documentation.  We can do this once we agree on item 2.

**If applicable**:

NOTE: contains some documentation but not enough to ship with

### DIFF
--- a/v2/brigadier/src/jobs.ts
+++ b/v2/brigadier/src/jobs.ts
@@ -2,7 +2,66 @@ import { Event } from "./events"
 
 const defaultTimeout: number = 1000 * 60 * 15
 
-export class Job {
+/**
+ * A Brigade job.
+ * 
+ * The factory methods on this type create jobs in various ways. Jobs do not
+ * run immediately when created. You must call the Job#run method to start
+ * the job running.
+ */
+export abstract class Job {
+  /**
+   * Runs the job.
+   */
+  public abstract run(): Promise<void>;
+
+  /**
+   * Specifies a job that runs in a container. By default, the container is simply
+   * executed (via its default entry point). Set the ContinerJob#tasks property to run
+   * specific commands in the container instead.
+   * @param name The name of the job
+   * @param image The container image to run in this job
+   * @param event The event that triggered the job
+   */
+  public static container(name: string, image: string,  event: Event): ContainerJob {
+    return new ContainerJob(name, image, event);
+  }
+
+  /**
+   * Specifies a job that consists of sub-jobs running in parallel.
+   * When run, all sub-jobs are started simultaneously (subject to scheduling
+   * constraints). The parallel job completes when all sub-jobs have
+   * completed.
+   * @param jobs The jobs to be run in parallel
+   */
+  public static parallel(jobs: Job[]): ParallelJob {
+    return new ParallelJob(jobs);
+  }
+
+  /**
+   * Specifies a job that consists of sub-jobs running one after another.
+   * A new sub-job is started only when the previous one completes.
+   * The sequence completes when the last job has completed (or when any
+   * job fails).
+   * @param jobs The jobs to be run in sequence
+   */
+  public static sequence(jobs: Job[]): SequentialJob {
+    return new SequentialJob(jobs);
+  }
+
+  /**
+   * Specifies a job that can be retried. The job will be run and re-run
+   * until either it succeeds or it has been tried `maxAttempts` times.
+   * (If maxAttempts is zero or less the job will not be tried at all.)
+   * @param job The job to be run and retried if necessary
+   * @param maxAttempts The maximum number of times to attempt the job
+   */
+  public static retryable(job: Job, maxAttempts: number): RetryableJob {
+    return new RetryableJob(job, maxAttempts);
+  }
+}
+
+export class ContainerJob extends Job {
   public name: string
   public primaryContainer: Container
   public sidecarContainers: { [key: string]: Container } = {}
@@ -15,6 +74,7 @@ export class Job {
     image: string,
     event: Event
   ) {
+    super();
     this.name = name
     this.primaryContainer = new Container(image)
     this.event = event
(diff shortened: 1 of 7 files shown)


krancour:
@itowlson thanks for the PR. I admittedly like the simplicity of `Group` as it currently exists, but this point, is spot-on and well-taken:

> However, this restricts composition to a single level.

So... I'm very open to a new approach. Now is certainly the right time.

Re: backwards compatibility, we have the flexibility to make breaking changes in v2.0.0, and our current state of release (alpha) doesn't prohibit us from further breaking changes. So again... timing is good.

Disclaimer here that I have _not_ looked at the code yet. I'm mainly reacting to your excellent write-up here.

> I tentatively used Job for the base/factory type and renamed the current Job class to ContainerJob. This blows backward compatibility out of the water and I am concerned it could cause problems, especially for JavaScript users who will not get it typechecked. A safer approach might be to declare a Runnable interface - the combinators would then compose Runnables into Runnables, and container jobs could go back to being called Job. But an interface cannot contain static methods, so what would we call the class where those factory methods lived? (Assuming we even like factory methods - I feel Job.sequence(...) is nicer than new SequentialJob(...) but others might disagree?)

Although breaking backwards compatibility is permitted right now, I feel like renaming `Job` to `ContainerJob` might go too far because it's _too_ different from 1.x, it will be certain to trip up a lot of people.

I _love_ the idea of the `Runnable` interface. Can we riff on that idea together a little further here?

How would you feel about keeping the existing `Job` class the way it is, introducing the `Runnable` interface as you have proposed, and then adding not one, but _two_ new classes-- something like `SerialGroup` and `ConcurrentGroup` (or a third, `RetryableGroup`). All of these _as well_ as `Job` would implement `Runnable`. i.e. Any one of these things could be executed with a `.run()` call and compositions could be arbitrarily complex-- which I think was your main goal.

Example pseudocode:

```
new ConcurrentGroup(
  new SerialGroup(job1, job2, job3),
  new SerialGroup(jobA, jobB, jobC, jobD),
).run();
```

Any thoughts?

v2/brigadier/src/jobs.ts
```diff
@@ -2,7 +2,66 @@ import { Event } from "./events"
...
 
-export class Job {
+/**
+ * A Brigade job.
+ * 
+ * The factory methods on this type create jobs in various ways. Jobs do not
+ * run immediately when created. You must call the Job#run method to start
+ * the job running.
+ */
+export abstract class Job {
```
technosophos, line 12: I know this would make things a little more Java-ish... but per @krancour's feedback, could we name this `AbstractJob` and rename `ContainerJob` to `Job`? That keeps naming nice and concise for regular usage, and also preserves backward compatibility. But it still would give us the lovely class hierarchy.

@adamreese would also point out that in ETL lingo, "Job" means "exactly one step in a pipeline". So in ETL, it would be a misnomer to say "A sequence job is a series of jobs..." Though `JobSequence` would be accurate.

So the terminology is "A task is made up of one or more jobs." We never really did define a concept called `Task` because we initially thought of the entire script as a task. So if we wanted to borrow the original terminiology, we _could_ make the abstract a `Task` and make `Job` one implementation, and `TaskSequence` or `SequenceTask` another.

Sorry... lots of background here.

I'll mark this comment as blocking, but I can definitely be convinced otherwise.

v2/brigadier/src/jobs.ts
```diff
@@ -50,3 +110,47 @@ export class JobHost {
...
+  }
+
+  public async run(): Promise<void> {
+    for (const job of this.jobs) {
+      await job.run();
+    }
+  }
+}
+
+export class RetryableJob extends Job {
```
technosophos, line 137: `RetryableJob` does indeed make sense to me as a name for this thing, given the above ETL-isms and stuff. So I very much like both this name and this addition to the API.

technosophos (changes requested):
Basically, I just have some questions about naming in order to keep things in line with the previous nomenclature. But I love the additions to the API, and this will vastly simplify some of the code I have seen in the wild.

v2/brigadier/src/jobs.ts
technosophos: Just to note the correction in chat... I totally got it backwards:

- Pipelines have jobs
- Jobs have tasks
- Task is the smallest unit of work (inside of a job)

Sorry...

adamreese: There isn't really a standard terminology it varies by tools and the layers of abstraction provided.  The abstractions should follow how each step is initialized and how data is communicated.

Another example could be:
  Task: smallest unit of work. Extract, load, or transform information. Kick off a Job
  Job: an abstract grouping of Tasks or Jobs

adamreese: There can also be an abstraction for the linkage between tasks.  Filtering, merging or any other routing of data.

itowlson:
Looking at the comments and #1292 confirms my reservations about renaming `Job`.  Between the `Runnable` interface and `AbstractJob` base class I would definitely go with `Runnable` - the only reason for a base class rather than an interface was somewhere to put the factories/combinators.

So my feeling is that #1292, as the 'less breaking' approach, is probably the way to go.  I would still yearn for factories/combinators as I find these more discoverable (type `Factory.` and let code completion tell you the name) but that is very much cosmetic.  As Kent says, the goal is simply to allow arbitrary composition, and #1292 achieves that 100%.

I am happy to close this and continue discussion on #1292 unless people have anything more to add here.  Thanks heaps for the feedback and for being willing to consider this so positively at such a late stage.

krancour:
#1292 went in instead.

PR #1292, opened 2021-02-19
better groups in brigadier / worker + docs

krancour:
This is a possible alternative to what is proposed by #1290.

I'm not seeking to get this merged at this time.

This is to fuel discussion, particularly with @itowlson.

### DIFF
--- a/v2/brigadier/src/events.ts
+++ b/v2/brigadier/src/events.ts
@@ -1,26 +1,58 @@
 import { Worker } from "./workers"
 import { Project } from "./projects"
 
+/**
+ * An event received by Brigade that resulted in triggering an event handler.
+ */
 export interface Event {
+  /** A unique identifier for the event. */
   id: string
+  /** The project that registered the handler being called for the event. */
   project: Project
+  /** The gateway which created the event. */
   source: string
+  /** The type of event. Values and meanings are source-specific. */
   type: string
+  /** A short title for the event, suitable for display in space-limited UI such as lists. */
   shortTitle?: string
+  /** A detailed title for the event. */
   longTitle?: string
+  /** The content of the event. This is source- and type-specific. */
   payload?: string
+  /** The Brigade worker assigned to handle the event. */
   worker: Worker
 }
 
+/** The type of the procedure to handle an Event. */
 export type EventHandler = (event: Event) => void
 
+/**
+ * Contains event handler registrations for a script.
+ * 
+ * Access the registry through the global `events` object.
+ */
 export class EventRegistry {
   protected handlers: { [key: string]: EventHandler } = {}
   
+  /**
+   * Registers a handler to be run when a particular kind of event occurs.
+   * When Brigade receives a matching event, it will run the specified
+   * handler.
+   * 
+   * Possible event sources depend on the gateways configured in your Brigade
+   * system, and possible event types depend on the gateway. See the Brigade
+   * and gateway documentation for details.
+   * 
+   * @param eventSource The event source (gateway) to register the handler for
+   * @param eventType The event type to register the handler for
+   * @param eventHandler The handler to run when an event with the given source and
+   * type occurs
+   */
   public on(eventSource: string, eventType: string, eventHandler: EventHandler): this {
     this.handlers[`${eventSource}:${eventType}`] = eventHandler
     return this
   }
 }
 
+/** Contains event handler registrations for a script. */
 export const events = new EventRegistry()
--- a/v2/brigadier/src/events.ts
+++ b/v2/brigadier/src/events.ts
@@ -9,7 +9,7 @@ export interface Event {
   id: string
   /** The project that registered the handler being called for the event. */
   project: Project
-  /** The gateway which created the event. */
+  /** The unique identifier of the gateway which created the event. */
   source: string
   /** The type of event. Values and meanings are source-specific. */
   type: string
--- a/v2/brigadier/src/groups.ts
+++ b/v2/brigadier/src/groups.ts
@@ -1,45 +1,46 @@
-import { Job } from "./jobs"
+import { Runnable } from "./runnables"
 
-export class Group {
+class Group {
+  protected runnables: Runnable[]
 
-  public static async runAll(jobs: Job[]): Promise<void[]> {
-    const group = new Group(jobs)
-    return group.runAll()
+  public constructor(...runnables: Runnable[]) {
+    this.runnables = runnables || []
   }
 
-  public static async runEach(jobs: Job[]): Promise<void> {
-    const group = new Group(jobs)
-    return group.runEach()
-  }
-
-  private jobs: Job[]
-
-  public constructor(jobs?: Job[]) {
-    this.jobs = jobs || []
-  }
-
-  public add(...jobs: Job[]): void {
-    for (const job of jobs) {
-      this.jobs.push(job)
+  public add(...runnables: Runnable[]): void {
+    for (const runnable of runnables) {
+      this.runnables.push(runnable)
     }
   }
 
   public length(): number {
-    return this.jobs.length
+    return this.runnables.length
   }
+}
 
-  public async runEach(): Promise<void> {
-    for (const job of this.jobs) {
-      await job.run()
+export class SerialGroup extends Group implements Runnable {
+
+  public async run(): Promise<void> {
+    for (const runnable of this.runnables) {
+      await runnable.run()
     }
   }
 
-  public async runAll(): Promise<void[]> {
+}
+
+export class ConcurrentGroup extends Group implements Runnable {
+
+  public async run(): Promise<void> {
     const promises: Promise<void>[] = []
-    for (const job of this.jobs) {
-      promises.push(job.run())
+    for (const runnable of this.runnables) {
+      promises.push(runnable.run())
     }
-    return Promise.all(promises)
+    try {
+      await Promise.all(promises)
+      return Promise.resolve()
+    } catch(e) {
+      return Promise.reject(e)
+    }
   }
 
 }
--- a/v2/brigadier/src/groups.ts
+++ b/v2/brigadier/src/groups.ts
@@ -1,7 +1,7 @@
 import { Runnable } from "./runnables"
 
 /**
- * The base type for Runnables that contain multiple other Runnables.
+ * The base type for Runnables composed of other Runnables.
  * Do not construct the base Group type; use Job.sequential (or SerialGroup)
  * or Job.concurrent (or ConcurrentGroup) instead.
  * 
@@ -26,7 +26,7 @@ class Group {
 }
 
 /**
- * A Runnable that consists of sub-Runnables (such as jobs
+ * A Runnable composed of sub-Runnables (such as jobs
  * or concurrent groups) running one after another.
  * A new Runnable is started only when the previous one completes.
  * The sequence completes when the last Runnable has completed (or when any
@@ -56,7 +56,7 @@ export class SerialGroup extends Group implements Runnable {
 }
 
 /**
- * A Runnable that consists of sub-Runnables (such as jobs
+ * A Runnable composed of sub-Runnables (such as jobs
  * or sequential groups) running concurrently.
  * When run, all Runnables are started simultaneously (subject to
  * scheduling constraints).
--- a/v2/brigadier/src/groups.ts
+++ b/v2/brigadier/src/groups.ts
@@ -1,9 +1,16 @@
 import { Runnable } from "./runnables"
 
+/**
+ * The base type for Runnables that contain multiple other Runnables.
+ * Do not construct the base Group type; use Job.sequential (or SerialGroup)
+ * or Job.concurrent (or ConcurrentGroup) instead.
+ * 
+ * @abstract
+ */
 class Group {
   protected runnables: Runnable[]
 
-  public constructor(...runnables: Runnable[]) {
+  protected constructor(...runnables: Runnable[]) {
     this.runnables = runnables || []
   }
 
@@ -18,7 +25,19 @@ class Group {
   }
 }
 
+/**
+ * A Runnable that consists of sub-Runnables (such as jobs
+ * or concurrent groups) running one after another.
+ * A new Runnable is started only when the previous one completes.
+ * The sequence completes when the last Runnable has completed (or when any
+ * Runnable fails).
+ * 
+ * @param runnables The work items to be run in sequence
+ */
 export class SerialGroup extends Group implements Runnable {
+  public constructor(...runnables: Runnable[]) {
+    super(...runnables)
+  }
 
   public async run(): Promise<void> {
     for (const runnable of this.runnables) {
@@ -28,7 +47,19 @@ export class SerialGroup extends Group implements Runnable {
 
 }
 
+/**
+ * A Runnable that consists of sub-Runnables (such as jobs
+ * or sequential groups) running concurrently.
+ * When run, all Runnables are started simultaneously (subject to
+ * scheduling constraints).
+ * The concurrent group completes when all Runnables have completed.
+ * 
+ * @param runnables The work items to be run in parallel
+ */
 export class ConcurrentGroup extends Group implements Runnable {
+  public constructor(...runnables: Runnable[]) {
+    super(...runnables)
+  }
 
   public async run(): Promise<void> {
     const promises: Promise<void>[] = []
--- a/v2/brigadier/src/groups.ts
+++ b/v2/brigadier/src/groups.ts
@@ -39,6 +39,14 @@ export class SerialGroup extends Group implements Runnable {
     super(...runnables)
   }
 
+  /**
+   * Runs the serial group.
+   * 
+   * @returns A Promise which completes when the last item in the group completes (or
+   * any item fails). If all items ran successfully, the Promise resolves; if any
+   * item failed (that is, its Runnable#run Promise rejected), the Promise
+   * rejects.
+   */
   public async run(): Promise<void> {
     for (const runnable of this.runnables) {
       await runnable.run()
@@ -61,6 +69,14 @@ export class ConcurrentGroup extends Group implements Runnable {
     super(...runnables)
   }
 
+  /**
+   * Runs the concurrent group.
+   * 
+   * @returns A Promise which completes when all items in the group complete.
+   * If all items ran successfully, the Promise resolves; if any
+   * item failed (that is, its Runnable#run Promise rejected), the Promise
+   * rejects.
+   */
   public async run(): Promise<void> {
     const promises: Promise<void>[] = []
     for (const runnable of this.runnables) {
--- a/v2/brigadier/src/jobs.ts
+++ b/v2/brigadier/src/jobs.ts
@@ -1,4 +1,5 @@
 import { Event } from "./events"
+import { ConcurrentGroup, SerialGroup } from "./groups"
 import { Runnable } from "./runnables"
 
 const defaultTimeout: number = 1000 * 60 * 15
@@ -27,6 +28,22 @@ export class Job implements Runnable {
 
   public logs(): Promise<string> {
     return Promise.resolve("skipped logs")
+  }
+
+  public static pod(
+    name: string,
+    image: string,
+    event: Event
+  ): Job {
+    return new Job(name, image, event)
+  }
+
+  public static sequence(...jobs: Job[]): SerialGroup {
+    return new SerialGroup(...jobs)
+  }
+
+  public static concurrent(...jobs: Job[]): ConcurrentGroup {
+    return new ConcurrentGroup(...jobs)
   }
 }
 
--- a/v2/brigadier/src/jobs.ts
+++ b/v2/brigadier/src/jobs.ts
@@ -1,8 +1,9 @@
 import { Event } from "./events"
+import { Runnable } from "./runnables"
 
 const defaultTimeout: number = 1000 * 60 * 15
 
-export class Job {
+export class Job implements Runnable {
   public name: string
   public primaryContainer: Container
   public sidecarContainers: { [key: string]: Container } = {}
--- a/v2/brigadier/src/jobs.ts
+++ b/v2/brigadier/src/jobs.ts
@@ -10,12 +10,12 @@ const defaultTimeout: number = 1000 * 60 * 15
  * Instances of Job represent containers that your Brigade script can
  * run using the Job#run method.
  * 
- * The Job#primaryContainer, initialised from the constructor image argument, determines
- * how long the job runs and whether it is considered successful.By default, the container is simply
+ * The Job#primaryContainer, initialized from the constructor image argument, determines
+ * how long the job runs and whether it is considered successful. By default, the container is simply
  * executed (via its default entry point). Set the Job#primaryContainer#command
  * property to run a specific command in the container instead.
  * Other containers, specified via Job#sidecarContainers, are automatically
- * terminated after the primary container completes.
+ * terminated a short time after the primary container completes.
  * 
  * Job also provides static methods for building up runnable
  * elements out of more basic ones. For example, to compose
@@ -23,7 +23,7 @@ const defaultTimeout: number = 1000 * 60 * 15
  * Job.sequence method.
  */
 export class Job implements Runnable {
-  /** The name of the jon. */
+  /** The name of the job. */
   public name: string
 
   /** Provides configuration options for the job's primary container. */
@@ -39,7 +39,7 @@ export class Job implements Runnable {
    */
   public timeout: number = defaultTimeout
 
-  /** Provides configuration options for the job execution environment. */
+  /** Specifies requirements for the job execution environment. */
   public host: JobHost = new JobHost()
 
   /** The event that triggered the job. */
@@ -68,6 +68,10 @@ export class Job implements Runnable {
    * When the primary container exits, the job ends. If sidecars are still running
    * at this point, Brigade terminates them after a short delay.
    * 
+   * NOTE: In a local test environment, this function does not run the containers,
+   * but instead automatically succeeds. In the real Brigade runtime environment,
+   * the containers run as described.
+   * 
    * @returns A Promise which completes when the primary container completes. If the
    * primary container succeeded (exited with code 0), the Promise resolves; if the
    * primary container failed (exited with nonzero code) or timed out, the Promise
@@ -81,13 +85,16 @@ export class Job implements Runnable {
    * Gets the job logs.
    * 
    * If the job has multiple containers, this aggregates the logs from them all.
+   * 
+   * NOTE: In a local test environment, this function returns a dummy log. In the
+   * real Brigade runtime environment, it returns the actual container logs.
    */
   public logs(): Promise<string> {
     return Promise.resolve("skipped logs")
   }
 
   /**
-   * Specifies a Runnable that consists of one or more containers. The
+   * Specifies a Runnable that executes one or more containers. The
    * image argument specifies the primary container, which determines
    * how long the job runs and whether it is considered successful.
    * By default, the primary container is simply
@@ -113,7 +120,7 @@ export class Job implements Runnable {
   }
 
   /**
-   * Specifies a Runnable that consists of sub-Runnables (such as jobs
+   * Specifies a Runnable composed of sub-Runnables (such as jobs
    * or concurrent groups) running one after another.
    * A new Runnable is started only when the previous one completes.
    * The sequence completes when the last Runnable has completed (or when any
@@ -126,7 +133,7 @@ export class Job implements Runnable {
   }
 
   /**
-   * Specifies a Runnable that consists of sub-Runnables (such as jobs
+   * Specifies a Runnable composed of sub-Runnables (such as jobs
    * or sequential groups) running concurrently.
    * When run, all Runnables are started simultaneously (subject to
    * scheduling constraints).
@@ -139,6 +146,9 @@ export class Job implements Runnable {
   }
 }
 
+/**
+ * A single OCI container in a Job.
+ */
 export class Container {
   /** The OCI image that the container should run. */
   public image: string
@@ -160,8 +170,8 @@ export class Container {
    * array for subcommands and the arguments array for argument values.
    * 
    * @example
-   * job.command = ["helm", "install"]
-   * job.arguments = ["stable/nginx", "-g"]
+   * job.primaryContainer.command = ["helm", "install"]
+   * job.primaryContainer.arguments = ["stable/nginx", "-g"]
    */
   public command: string[] = []
   /**
@@ -170,24 +180,41 @@ export class Container {
    * array for subcommands and the arguments array for argument values.
    * 
    * @example
-   * job.command = ["helm", "install"]
-   * job.arguments = ["stable/nginx", "-g"]
+   * job.primaryContainer.command = ["helm", "install"]
+   * job.primaryContainer.arguments = ["stable/nginx", "-g"]
    * 
    */
   public arguments: string[] = []
-  /** Environment variables to set in the container. */
+  /**
+   * Environment variables to set in the container. These are often derived from
+   * project settings such as secrets.
+   * 
+   * You can safely pass secrets via environment variables, because Brigade treats
+   * all enviroment variables as secrets.
+   * 
+   * @example
+   * job.primaryContainer.env.AUTH_TOKEN = e.project.secrets.authToken  // e is event that triggered the handler
+   */
   public environment: { [key: string]: string } = {}
   /**
    * The path in the container's file system where, if applicable, the
    * Brigade worker's shared workspace should be mounted. If empty (the default),
    * the Job does not have access to the shared workspace.
+   * 
+   * The shared workspace must be enabled at the project configuration level
+   * for containers to access it. If it is not enabled, you should leave this
+   * property empty.
    */
   public workspaceMountPath = ""
   /**
    * The path in the container's file system where, if applicable,
    * source code retrieved from a version control system repository should be
-	 * mounted. If empty (the default), Brigade will not mount any source
+   * mounted. If empty (the default), Brigade will not mount any source
    * code automatically.
+   * 
+   * Source code mounting must be enabled at the project configuration level
+   * for containers to access it. If it is not enabled, you should leave this
+   * property empty.
    */
   public sourceMountPath = ""
   /**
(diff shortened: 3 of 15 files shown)


itowlson:
This looks good, and makes for fewer and less dramatic breaking changes than my effort.  We could even keep the static `runAll` and `runEach` methods on Group though possibly mark them as deprecated.

I would still like to see factory methods, and aesthetically I would like to see them on the Runnable type, but that name is so technical that it's not appealing, even if TypeScript allowed it!  But if we could find a better name we could make it a class instead of an interface (and have Job and Group inherit from it), and put the combinators there.  But maybe that's not worth the effort when the `new`-based approach is entirely functional.

krancour:
@itowlson you have a lot of great ideas about how to improve this-- and far superior TS chops to my own. What are the chances you want to take this PR over and incorporate some of those improvements?

krancour:
@itowlson in #1290, you wrote:

> I would still yearn for factories/combinators as I find these more discoverable (type Factory. and let code completion tell you the name) but that is very much cosmetic. 

I've been thinking about the issue of "discoverability" a lot since that comment.

Is the assumption that something like the following is harder because it involves an _additional_ class of which the code's author need some foreknowledge?

```typsecript
import { Job, SerialGroup } from "@brigadecore/brigadier"

// ...

new SerialGroup(job1, job2, ..., jobN).run()
```

vs the following, which requires no foreknowledge of `SerialGroup` being something you might want to import and use?

```typescript
import { Job } from "@brigadecore/brigadier"

// ...

Job.sequence(job1, job2, ..., jobN).run()
```

Because I can totally see that's indisputably true.

Would this PR be improved by adding static functions to the the `Job` class that look something like this?

```TypeScript
  public static sequence(...runnables: Runnable[]): SerialGroup {
    return new SerialGroup(...runnables)
  }

  public static concurrent(...runnables: Runnable[]): ConcurrentGroup {
    return new ConcurrentGroup(...runnables)
  }
```

krancour:
Clarification to the title of this PR may be in order just in case is prompted any misunderstanding. This was proposed as "better" than how groups work _today_. Although this PR proposes a _alternative_ implementation of what's in #1290, the word "better" in the title wasn't being applied in relation to that effort. There is no assertion here that this is superior to #1290.

I believe there are valuable things in #1290 than can be ported over to here-- or vice versa.

itowlson:
@krancour Yes, the popup/context help from typing `Job.` was what I was thinking of.  It gives you a one-stop shop for "I want to make a Runnable, either from nothing or out of other Runnables."

It's perhaps a little bit odd that a function on Job produces Groups rather than Jobs (which was the motivation for rearranging everything under the Job type in #1290).  But that's a small price to pay for the improved compatibility.

I'm very happy to pick this up if you would like - let me know.  I'd probably need help taking it through integration testing etc. though.

krancour:
> It's perhaps a little bit odd that a function on Job produces Groups rather than Jobs

I agree. So if we went in this direction, we gain discoverability and context help, but we introduce this small peculiarity.

Do you think it improves the situation (tho perhaps only moderately) if `Group` (and it's `*-Group`) derivatives were renamed as `JobGroup`?

krancour:
> I'd probably need help taking it through integration testing etc.

Reach out to @vdice on that. He's been building out the integration tests and knows that area of the codebase better than myself.

vdice:
>Reach out to @vdice on that. He's been building out the integration tests and knows that area of the codebase better than myself.

We should have the integration tests running in CI shortly (https://github.com/brigadecore/brigade/pull/1297).  This spins up a kind cluster, builds Brigade images, loads them into the cluster, deploys Brigade and then runs the integration tests.

While the [tests](https://github.com/brigadecore/brigade/blob/v2/v2/tests/git_initializer_test.go) are still very much git-centric, we certainly hope to expand on them by adding non-git tests and, more pertinently to this effort, tests focusing on different brigade.js/ts scripts.

Definitely open to suggestions/collab on how to best support this expansion.  One next step might be to create a test data dir with project defs and/or brigade.js/ts scripts to read in and then run the standard "create project, create event, assert success (or not), assert logs, etc." for each.

v2/brigadier/src/jobs.ts
```diff
@@ -28,6 +29,22 @@ export class Job implements Runnable {
   public logs(): Promise<string> {
     return Promise.resolve("skipped logs")
   }
+
+  public static pod(
```
krancour, line 33: Tricky thing on the name here...

It's been a major goal of Brigade 2 to abstract everything away from Kubernetes so that k8s becomes merely an implementation detail. If it were ever to fall out of fashion, and some new orchestrator came en vogue, or the likes of Mesos experienced a renaissance, Brigade could live on with relatively few changes on the back end.

In light of this, we may want to reconsider `pod()` as a function name since it possibly tips its hand too much in terms of the underlying orchestrator.

Naming is hard.

itowlson: Hah, I was already planning to switch this helper to `container` - would that be okay?

krancour: Maybe it depends on what "container" means here.

If it means "a container-_based_ job, that seems accurate.

But it could also mislead people into believing a job (singular) is implemented as container (singular), when in reality a job can be composed of multiple containers.

Add in my general reservations about revealing too much about the underlying implementation........ and then I have to back-pedal and acknowledge that the word container is already used in a bunch of other places. ::sigh::

Then another consideration would be that (if I understand correctly), Krustlet is enabling things that _aren't_ containers at all (namely web assembly modules) to run in k8s. If that works as advertised, I believe a job can conceivably be composed of those instead, without a single change to Brigade.

Smoke is coming out of my head as I grapple with whether "container" can simultaneously be accurate-enough _and_ not overly-specific about the underlying implementation.

But as I grapple with it, I'm coming to wonder if we need it...

What would `Job.container(...)` buy us that we don't already get from `new Job(...)`? In terms of the other two new static functions, I see that they address concerns over discoverability / context help, and make `Job` a one-stop shop, but I'm less clear on the benefits of `Job.container(...)`....... but I'm also not as well versed in common TS idioms for this sort of thing. 🤷‍♂️

itowlson: I was hesitant about proposing it and could be talked out if it.  The reasons I proposed it were essentially it's for consistency and explicitness.  Consistency in that it makes `Job.` a one-stop shop for _all_ the ways you can construct things to run; and explicitness in saying "although Job can produce other kinds of Runnable, this gives you the _particular kind_ which is executed by running a container."

And the trouble is... if "a Job runs a container" isn't true, then _how do we want to document Job?_  What do we tell people a Job is?  Is it "it runs an OCI image"?  But then you can say "but what if one day we do WASM stored in Bindle."  Is it "an external program, packaged in whatever way"?  But then what if we decide that Bindled WASM jobs need to be represented differently, and we need to contract the meaning of Job to be specifically OCI?

So... I'm wary of second guessing.  Today, a Job is basically a container (Krustlet notwithstanding).  I'm happy documenting it as such, and therefore I'm happy using the term in a factory method.  Tomorrow, maybe we'll need to reconsider,

But possibly that's a third benefit of the factory method.  Suppose we want to support Bindled WASM, and we find that requires different info from a container.  (Say, it needs a selector, and it can't have a `command`.)  Well, `Job.container` and `Job.bindle` make that difference clear and can have different signatures.  Whereas `new Job` will become ambiguous and might need optional parameters and runtime configuration errors.

Of course, the converse is that if we are happy to say "a Job today _is_ a container", then perhaps we should be happy to defer the explicitness argument, and stick to `new Job` until Job _does_ become ambiguous.  I dunno.  I'm thinking out loud here, trying to figure out how we document Job (and whether I can then turn that into a terse method name).

For myself, I like the explicitness and I like the "I can type `Job.` and find out what my options are" effect.  But I totally agree that it is close to redundant, and may be more specific than we want to be, and if you do object then I won't shed many tears about taking it out again.

Hope this all makes sense; happy to discuss on a call if face-to-face would be easier!

krancour: Well now both our heads are smoking grappling with naming.

I really like your point supposing we did need to someday define different sorts of jobs differently-- like a WASM based job hypothetically requiring additional fields.

Based on that argument, consider me sold. Your approach leaves room for expansion without ambiguity.

Any thoughts on changing `container()` to `containerBased()`? Could that convey the manufacture of a container-_based_ job without opening the door to misinterpretation that a job is a (singular) container?

krancour:
@itowlson I am super appreciative of the JSdocs you've started adding in this PR. It's a glaring and embarrassing oversight that we didn't already have any. Our Go code is thoroughly documented and the brigade-sdk-for-js (separate repo) is also.... I have no idea how we overlooked docs here, but I sure appreciate you starting to set that right.

v2/brigadier/src/jobs.ts
itowlson: For my money `containerBased` kind of muddies the water (though I see where you are coming from).  My interpretation of a job was that it ran a container (or commands in that container) _but_ you could stick sidecars on in case you wanted to proxy network calls or something - but the sidecars were ancillary not job-defining.  But if the secondary containers are expected to perform actual job-related work, then maybe we need to open the naming to that.

krancour: > But if the secondary containers are expected to perform actual job-related work, then maybe we need to open the naming to that.

We're not opinionated about what they do, but the "rules" are the exit code of the primary container is the determinant of job success or failure. The exit codes of the others aren't ever checked. When the primary container exits, the whole job is considered done. One minute later (configurable), we delete the whole pod, because k8s is really unintelligent about this. k8s makes not actual distinction between your "primary" container and a sidecar, so when the primary exits, the sidecars stupidly could remain running indefinitely. There are several extensive threads about this defficiency in k8s. We circumvent all of that by torching the whole pod-- something we have the luxury of doing in Brigade 2 that we couldn't have done in 1.x-- because unlike 1.x, 2.0 aggregates logs and stores them somewhere. 1.x had to hang on to your pods for as long as you wanted to keep the logs (and even then you could still lose them).

I'm getting off track. Long story short is that the view of _a_ job as _a_ container is probably not quite right. It is, without a doubt, the common case, but jobs can be an assembly of multiple containers.

v2/brigadier/src/projects.ts
```diff
@@ -1,4 +1,9 @@
+/**
+ * A Brigade project.
+ */
```
krancour, line 3: Nit: I don't know if it's misleading at all to just call this "a Brigade project." The reality is that Brigadier / the worker see only a relevant subset of all event and project details. i.e. It's a Brigadier / worker-specific _representation_ of the project or event.

A more accurate description for this might be something along the lines of "encapsulates project-related details," since that hints that a "project" is a broader thing that exists elsewhere and this is a narrower view of it-- which is true.

Again, though. This is a nit. Dealer's choice on whether this is worth rewording or not.

v2/brigadier/src/events.ts
```diff
@@ -1,26 +1,58 @@
 import { Worker } from "./workers"
 import { Project } from "./projects"
 
+/**
+ * An event received by Brigade that resulted in triggering an event handler.
+ */
```
krancour, line 6: My initial reaction to the term "handler" was that it was inconsistent with how we talk about Brigade. We might use the term "handler" when defining what a "worker" is (a containerized event handler), but otherwise we haven't used that term much.

Then I realized that "handler" here is more in reference to the _Brigadier_ programming model. Since we do things like `events.on("someEvent", someFunction)`............. yup....... that's a handler alright.

So contextually, this makes perfect sense. 👍

v2/brigadier/src/events.ts
```diff
@@ -1,26 +1,58 @@
...
 export interface Event {
+  /** A unique identifier for the event. */
   id: string
+  /** The project that registered the handler being called for the event. */
```
krancour, line 10: Related to [this comment](https://github.com/brigadecore/brigade/pull/1292/files?file-filters%5B%5D=.ts#r591609582)

It might be better to say something along the lines of "relevant details of the project to which this event belongs."

But this is a non-blocking nit.

v2/brigadier/src/events.ts
```diff
@@ -1,26 +1,58 @@
...
   project: Project
+  /** The gateway which created the event. */
```
krancour, line 12: It's worth it to mention here that it's a unique identifier.

Docs on creating or configuring gateways will emphasize that the value of the `source` field on any event, in order to avoid collisions, should be a URL that is controlled exclusively by the author(s) of the gateway. So, for instance, the GitHub gateway I am working on right now uses a `source` value `github.com/brigadecore/brigade-github-gateway` (although I'm realizing now that `brigade.sh/github-gateway` would be more succinct and fulfill the same purpose). This is pretty much the same as Java's package naming strategy of using your own domain name in reverse as a package name prefix to avoid package name collisions (e.g. `com.example`).

Not all of this info needs to be captured here, but it's worth adding the word "unique" here.

v2/brigadier/src/groups.ts
```diff
@@ -1,45 +1,93 @@
-import { Job } from "./jobs"
+import { Runnable } from "./runnables"
 
-export class Group {
+/**
+ * The base type for Runnables that contain multiple other Runnables.
```
krancour, line 4: Non-blocking nit: "composed of other Runnables" might read better.

v2/brigadier/src/groups.ts
```diff
@@ -1,45 +1,93 @@
...
-  private jobs: Job[]
+  public length(): number {
+    return this.runnables.length
+  }
+}
 
-  public constructor(jobs?: Job[]) {
-    this.jobs = jobs || []
+/**
+ * A Runnable that consists of sub-Runnables (such as jobs
```
krancour, line 29: Non-blocking nit: "composed of" language might convey what this does more clearly.

v2/brigadier/src/groups.ts
```diff
@@ -1,45 +1,93 @@
...
-    return this.jobs.length
-  }
+}
 
-  public async runEach(): Promise<void> {
-    for (const job of this.jobs) {
-      await job.run()
-    }
+/**
+ * A Runnable that consists of sub-Runnables (such as jobs
```
krancour, line 59: Same non-blocking comment about "composed of" language.

v2/brigadier/src/groups.ts
```diff
@@ -1,45 +1,93 @@
-import { Job } from "./jobs"
+import { Runnable } from "./runnables"
 
-export class Group {
+/**
+ * The base type for Runnables that contain multiple other Runnables.
+ * Do not construct the base Group type; use Job.sequential (or SerialGroup)
+ * or Job.concurrent (or ConcurrentGroup) instead.
```
krancour, line 6: I think this is another case where you're going to know better than me what the best JS/TS approach is here, so consider this a question more than a comment...

`Group` doesn't get re-exported from `index.ts`. Unless I am mistaken (and I well could be), this makes direct instantiation of this `Group` base class by a user of the Brigadier package an impossibility. Or doesn't it?

If I'm right, we could lose the admonishment not to instantiate this directly. If I'm wrong, then forget I said anything.

And this is non-blocking anyway.

v2/brigadier/src/jobs.ts
```diff
@@ -1,15 +1,56 @@
...
 const defaultTimeout: number = 1000 * 60 * 15
 
-export class Job {
+/**
+ * A Brigade job.
+ * 
+ * Instances of Job represent containers that your Brigade script can
+ * run using the Job#run method.
+ * 
+ * The Job#primaryContainer, initialised from the constructor image argument, determines
```
krancour, line 13: Nit: We've used American English spellings throughout the project, so "initialized" would be more consistent.

Non-blocking.

v2/brigadier/src/jobs.ts
```diff
@@ -1,15 +1,56 @@
...
+ * how long the job runs and whether it is considered successful.By default, the container is simply
```
krancour, line 14: Missing space after period.

v2/brigadier/src/jobs.ts
```diff
@@ -1,15 +1,56 @@
...
+ * executed (via its default entry point). Set the Job#primaryContainer#command
+ * property to run a specific command in the container instead.
+ * Other containers, specified via Job#sidecarContainers, are automatically
+ * terminated after the primary container completes.
```
krancour, line 18: There's a lag on this which is configurable by the operator who installs Brigade. It's not worth getting into the technical details of why, but it might be worth mentioning the delay in order to calibrate user expectations vis-a-vis "automatically terminated."

v2/brigadier/src/jobs.ts
```diff
@@ -1,15 +1,56 @@
...
+ * 
+ * Job also provides static methods for building up runnable
+ * elements out of more basic ones. For example, to compose
+ * a set of workloads into a sequence, you can use the
+ * Job.sequence method.
+ */
+export class Job implements Runnable {
+  /** The name of the jon. */
```
krancour, line 26: s/jon/job

v2/brigadier/src/jobs.ts
```diff
@@ -1,15 +1,56 @@
...
   public primaryContainer: Container
+
+  /** Specifies sidecar containers to run alongside the primary container. */
   public sidecarContainers: { [key: string]: Container } = {}
+
+  /**
+   * The duration, in milliseconds, after which Brigade should automatically
+   * terminate and fail the job if it has not completed. The default is 15
+   * minutes.
+   */
```
krancour, line 39: Note to self-- the back-end components aren't enforcing this yet.

v2/brigadier/src/jobs.ts
```diff
@@ -1,15 +1,56 @@
...
   public timeout: number = defaultTimeout
+
+  /** Provides configuration options for the job execution environment. */
```
krancour, line 42: It might be more appropriate to call these "host selection options."

In a perfect world, no one would ever use these-- you wouldn't care what host your job runs on. In the real world, you're once in a while going to have to specify something like "must run on a Windows node."

v2/brigadier/src/jobs.ts
```diff
@@ -20,33 +61,177 @@ export class Job {
     this.event = event
   }
 
+  /**
+   * Runs the job.
+   * 
+   * When you run the job, Brigade runs all the containers, primary and sidecar.
+   * When the primary container exits, the job ends. If sidecars are still running
+   * at this point, Brigade terminates them after a short delay.
```
krancour, line 69: This "short delay" language is perfect... might be the answer to resolving this [other comment](https://github.com/brigadecore/brigade/pull/1292/files?file-filters%5B%5D=.ts#r591631849)

v2/brigadier/src/jobs.ts
```diff
@@ -20,33 +61,177 @@ export class Job {
...
+   */
   public run(): Promise<void> {
     return Promise.resolve()
   }
 
+  /**
+   * Gets the job logs.
+   * 
+   * If the job has multiple containers, this aggregates the logs from them all.
+   */
```
krancour, line 84: This comment might apply in another spot or two that I failed to identify.

The description here is inconsistent with that the logic below does. Is it worth explaining in the comment that the logic here always resolves with the string "skipped logs," but that the Brigade worker "poly-fills" this at runtime with logic that communicates with the Brigade API?

v2/brigadier/src/jobs.ts
```diff
@@ -20,33 +61,177 @@ export class Job {
...
+  public static container(
+    name: string,
+    image: string,
+    event: Event
+  ): Job {
+    return new Job(name, image, event)
+  }
+
+  /**
+   * Specifies a Runnable that consists of sub-Runnables (such as jobs
```
krancour, line 116: Another non-blocking nit about "composed of language" possibly reading better. Another dealer's choice.

v2/brigadier/src/jobs.ts
```diff
@@ -20,33 +61,177 @@ export class Job {
...
+   * Runnable fails).
+   * 
+   * @param runnables The work items to be run in sequence
+   */
+  public static sequence(...runnables: Runnable[]): SerialGroup {
+    return new SerialGroup(...runnables)
+  }
+
+  /**
+   * Specifies a Runnable that consists of sub-Runnables (such as jobs
```
krancour, line 129: Another non-blocking nit about "composed of language" possibly reading better. Another dealer's choice.

v2/brigadier/src/jobs.ts
```diff
@@ -20,33 +61,177 @@ export class Job {
...
+   * The concurrent group completes when all Runnables have completed.
+   * 
+   * @param runnables The work items to be run in parallel
+   */
+  public static concurrent(...runnables: Runnable[]): ConcurrentGroup {
+    return new ConcurrentGroup(...runnables)
+  }
 }
 
 export class Container {
```
krancour, line 142: Are we missing doc on the  `Container` type?

v2/brigadier/src/jobs.ts
```diff
@@ -20,33 +61,177 @@ export class Job {
...
+  /** The OCI image that the container should run. */
   public image: string
+  /**
+   * Specifies under what conditions the host node should re-fetch an image
+   * that is already in its local cache. The permitted values are as IfNotPresent
+   * (the default) and Always.
+   */
   public imagePullPolicy = "IfNotPresent"
```
krancour, line 150: Non-blocking for this PR-- certainly not an issue introduced by this PR-- should we try to use an enum or constant here?

v2/brigadier/src/jobs.ts
```diff
@@ -20,33 +61,177 @@ export class Job {
...
+   * 
+   * Only the first element of the array is the actual command. Subsequent elements
+   * are treated as arguments. For example, a command of ["echo", "hello"] is
+   * equivalent to running 'echo hello'. A common convention is to use the command
+   * array for subcommands and the arguments array for argument values.
+   * 
+   * @example
+   * job.command = ["helm", "install"]
+   * job.arguments = ["stable/nginx", "-g"]
+   */
```
krancour, line 165: Spot on. 👍

krancour: Whoops... usage needs a minor tweak...

```
job.primaryContainer.command = ["helm", "install"]
job.primaryContainer.arguments = ["stable/nginx", "-g"]
```

v2/brigadier/src/jobs.ts
```diff
@@ -20,33 +61,177 @@ export class Job {
...
   public command: string[] = []
+  /**
+   * The arguments to pass to Container#command.  If the command includes arguments
+   * already, the arguments property are appended. A common convention is to use the command
+   * array for subcommands and the arguments array for argument values.
+   * 
+   * @example
+   * job.command = ["helm", "install"]
+   * job.arguments = ["stable/nginx", "-g"]
```
krancour, line 174: Same comment as [here](https://github.com/brigadecore/brigade/pull/1292/files?file-filters%5B%5D=.ts#r591650274)

v2/brigadier/src/jobs.ts
```diff
@@ -20,33 +61,177 @@ export class Job {
...
+   * 
+   */
   public arguments: string[] = []
+  /** Environment variables to set in the container. */
```
krancour, line 178: A common thing that will occur is that users will inject project secrets into the environment variables here-- which is perfectly safe because on the back end we treat _all_ of these environment variables as if they might have been secrets.

Is it worth adding example usage? It looks like this:

```
job.primaryContainer.env.FOO = e.project.secrets.foo
```

v2/brigadier/src/jobs.ts
```diff
@@ -20,33 +61,177 @@ export class Job {
...
   public environment: { [key: string]: string } = {}
+  /**
+   * The path in the container's file system where, if applicable, the
+   * Brigade worker's shared workspace should be mounted. If empty (the default),
+   * the Job does not have access to the shared workspace.
+   */
```
krancour, line 184: It's worth mentioning that project configuration would also need to have the shared workspace enabled for this to work.

fwiw, the reason this is enabled or disabled at the project level is because, eventually, we will introduce some optimizations that can be realized only by "stateless" workers-- i.e. ones that aren't passing state from job to job in the form of a shared workspace. For instance, we think we'll eventually have a "retry" that can skip jobs that already succeeded. That's enormously complicated for workflows that pass state around in a shared workspace, but easy to do otherwise.

v2/brigadier/src/jobs.ts
```diff
@@ -20,33 +61,177 @@ export class Job {
...
   public workspaceMountPath = ""
+  /**
+   * The path in the container's file system where, if applicable,
+   * source code retrieved from a version control system repository should be
+	 * mounted. If empty (the default), Brigade will not mount any source
```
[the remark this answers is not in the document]
krancour: This also would need to have been enabled in the project configuration.... can't mount source into the job's container(s) if we don't know where to get source from.

v2/brigadier/src/jobs.ts
```diff
@@ -20,33 +61,177 @@ export class Job {
...
+   */
   public sourceMountPath = ""
+  /**
+   * Whether the container should run with privileged permissions. This is
+   * typically required only for "Docker in Docker" scenarios where the
+   * container must run its own Docker daemon.
+   * 
+   * Privileged execution may be disallowed by Brigade project configuration.
+   * If so, the container will run unprivileged.
+   */
```
krancour, line 200: Spot on. Good context too with the "Docker in Docker" explanation.

v2/brigadier/src/jobs.ts
```diff
@@ -20,33 +61,177 @@ export class Job {
...
   public privileged = false
+  /**
+   * Whether the container should mount the host's Docker socket into its own
+   * file system. This is typically required only for "Docker-out-of-Docker" ("DooD")
+   * scenarios where the container needs to use the host's Docker daemon.
+   * This is strongly discouraged for almost all use cases.
+   * 
+   * Host Docker socket access may be disallowed by Brigade project configuration.
+   * If so, the container will run without such access.
+   */
```
krancour, line 210: Again-- spot on and great context for why you might use this.

v2/brigadier/src/jobs.ts
```diff
@@ -20,33 +61,177 @@ export class Job {
...
+   * Constructs a new Container.
+   * @param image The OCI reference to the container image
+   */
   constructor(image: string) {
     this.image = image
   }
 }
 
+/**
+ * The execution environment required by a Job.
```
krancour, line 223: See related [comment](https://github.com/brigadecore/brigade/pull/1292/files?file-filters%5B%5D=.ts#r591634530)

v2/brigadier/src/jobs.ts
```diff
@@ -20,33 +61,177 @@ export class Job {
...
 export class JobHost {
+  /**
+   * The OS required by the Job. Valid values are "linux" and "windows". When empty,
+	 * Brigade assumes "linux".
+   */
   public os?: string
+  /**
+   * Specifies labels that must be present on the substrate node to
+	 * host a Job. This provides an opaque mechanism for communicating Job needs
+	 * such as specific hardware like an SSD or GPU.
```
krancour, line 234: Indent is off.

v2/brigadier/src/logger.ts
```diff
@@ -1,25 +1,53 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+/**
+ * Provides logging services for a script.
+ * 
+ * Access the logger through the global `logger` object.
+ */
```
krancour, line 7: Another case where idk if it's worth mentioning that we polyfill this at runtime.

This implementation does `console.debug(...)` and `console.error(...)`, etc... and iirc, there's actually no difference between what each of those does, right? They all right to stdout/stderr and don't have any real concept of log severity and whether the message in question should be written out at all or not?

The polyfill that the worker uses relies upon a leveled logging library (winston, iirc) to get this right. We _don't rely on it here in Brigadier because that library drags too many transitive dependencies along with it.

v2/brigadier/src/workers.ts
```diff
@@ -1,7 +1,20 @@
+/**
+ * Represents the Brigade worker within which an event handler runs.
+ */
 export interface Worker {
+  /** The address of the Brigade API server. */
   apiAddress: string
+  /** A token which can be used to authenticate to the API server. */
```
krancour, line 7: Non-blocking-- it's up to you whether you think it's worth mentioning that this token _only_ allows this worker to create jobs for the current event _and nothing else._

v2/brigadier/src/workers.ts
```diff
@@ -1,7 +1,20 @@
...
   apiToken: string
+  /**
+   * The directory where the worker stores configuration files,
+   * including event handler code files such as brigade.js and brigade.json.
+   */
```
krancour, line 12: This defaults to `./.brigade`

krancour: But maybe not worth mentioning. `Worker` isn't something the user instantiates and configures. It's something that's handed _to_ them as part of the event they receive.

Forget I said anything.

v2/brigadier/src/workers.ts
```diff
@@ -1,7 +1,20 @@
...
   configFilesDirectory: string
+  /**
+   * The default values to use for any configuration files that are not present.
+   */
   defaultConfigFiles: { [key: string]: string }
+  /** The desired granularity of worker logs. */
```
krancour, line 18: Is it worth mentioning that this has no bearing on a job's logs?

v2/brigadier/src/events.ts
itowlson: Updated the wording on the Project interface so they'll see that if they drill in; I feel "Information about..." or "Details of..." is probably clutter here (though I do see where you are coming from).

v2/brigadier/src/groups.ts
itowlson: I think you are right. So we could indeed lose the admonition. Though in turn this people probably won't see these docs, so it does no harm! So I'm currently thinking - leave the admonition in, not so much for end users but more for people who are _reading the source code_ and wondering about the role of the Group class. Is that potentially useful, or just absurdly defensive?

v2/brigadier/src/jobs.ts
itowlson: Oops! Thanks for catching this!

v2/brigadier/src/jobs.ts
itowlson: Tweaked to "automatically terminated a short time after..." - does that meet your needs?  I don't want to get too much into the details in a JSDoc lest it turn into an essay...

v2/brigadier/src/jobs.ts
itowlson: Duh. I mean obviously the jon's name would be, er, Jon. Thanks.

v2/brigadier/src/jobs.ts
itowlson: Proposed "Specifies requirements for the job execution environment" but happy to iterate

v2/brigadier/src/jobs.ts
itowlson: Yes, good point. I've proposed some wording for both this and the `run` function (which I think are the two affected). Would appreciate your feedback once it's pushed.

v2/brigadier/src/jobs.ts
```diff
@@ -20,33 +61,177 @@ export class Job {
...
+   * Gets the job logs.
+   * 
+   * If the job has multiple containers, this aggregates the logs from them all.
+   */
   public logs(): Promise<string> {
     return Promise.resolve("skipped logs")
   }
+
+  /**
+   * Specifies a Runnable that consists of one or more containers. The
```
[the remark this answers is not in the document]
itowlson: I want to avoid 'composed' here because it sounds too much like what we're doing with groups (and possibly other composites).  Proposing "a Runnable that executes one or more containers".

v2/brigadier/src/jobs.ts
itowlson: We could use a TypeScript union type of strings, which would give the JS folks a familiar stringy interface (and compatibility with v1?) but provide intellisense and compilet-time validation to the TS folks:

```
type ImagePullPolicy = "Always" | "IfNotPresent";
public imagePullPolicy: ImagePullPolicy = "IfNotPresent";
```

Agree out of scope for this PR though.

v2/brigadier/src/jobs.ts
itowlson: Gah, sorry.  Thanks for spotting.

v2/brigadier/src/jobs.ts
itowlson: Added some wording and example, please check when pushed

v2/brigadier/src/jobs.ts
itowlson: I copied it from your docs for the Go SDK!

v2/brigadier/src/jobs.ts
itowlson: I feel comfortable with this level of abstraction, that's it's about the environments under which the job can run, but I can add some wording that makes it more specific to host selection if you want.

v2/brigadier/src/jobs.ts
itowlson: Think I had some rogue tabs from somewhere (maybe I copy-pasted from the Go docs!).  Hopefully all gone now.

v2/brigadier/src/workers.ts
itowlson: Yeah we should really make this interface `readonly`, not that it would stop JavaScript consumers.

itowlson:
@krancour Thanks for the eagle-eyed feedback. This is now ready for re-review.

v2/brigadier/src/events.ts
krancour: That sounds reasonable.

v2/brigadier/src/groups.ts
krancour: It sounds fine to me.

v2/brigadier/src/jobs.ts
krancour: btw... purely about consistency and not anything else.

v2/brigadier/src/jobs.ts
krancour: Yep... that's fine as long as we're not accidentally implying it's immediate.

v2/brigadier/src/jobs.ts
krancour: That sgtm

v2/brigadier/src/jobs.ts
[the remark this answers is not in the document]
krancour: That sgtm

v2/brigadier/src/jobs.ts
krancour: That looks good... I'm learning a lot of TS/JS from your suggestions.

v2/brigadier/src/jobs.ts
```diff
@@ -68,6 +68,10 @@ export class Job implements Runnable {
    * When the primary container exits, the job ends. If sidecars are still running
    * at this point, Brigade terminates them after a short delay.
    * 
+   * NOTE: In a local test environment, this function does not run the containers,
+   * but instead automatically succeeds. In the real Brigade runtime environment,
+   * the containers run as described.
+   *
```
krancour, line 74: 👍

krancour:
@itowlson thank you for the tremendous effort that went into this. _This looks fantastic._

vdice (approved):
Looks fantastic to me as well!  Thanks @itowlson and @krancour !

itowlson (approved):
Kent can't approve because he created the PR. So I'm approving it on his behalf!

Issue #1271, opened 2021-02-11
Delete corresponding logs when a project or event is deleted


krancour:
Assuming that one benefit of deleting old events is to reclaim storage, we can reclaim more (potentially a lot more) than we do now by deleting logs associated with a deleted project or event.

vdice:
Closed by #1294

PR #1294, opened 2021-02-19
feat(*): delete logs on event and project deletion

vdice:

- Adds logic to delete logs from cool storage (mongo) on event and project deletion
- In doing so, converts the logs collection from capped to regular (no manual deletes/removals possible with a capped collection)

Closes https://github.com/brigadecore/brigade/issues/1271

**If applicable**:

### DIFF
--- a/v2/apiserver/internal/core/events.go
+++ b/v2/apiserver/internal/core/events.go
@@ -237,6 +237,7 @@ type eventsService struct {
 	authorize           libAuthz.AuthorizeFn
 	projectsStore       ProjectsStore
 	eventsStore         EventsStore
+	logsStore           LogsStore
 	substrate           Substrate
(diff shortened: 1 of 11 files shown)


v2/apiserver/internal/core/events.go
```diff
@@ -663,6 +666,17 @@ func (e *eventsService) DeleteMany(
...
+				if err := e.logsStore.DeleteEventLogs(
+					context.Background(), // deliberately not using ctx
+					event.ID,
+				); err != nil {
+					log.Println(errors.Wrapf(
+						err,
+						"error deleting logs for event %q",
+						event.ID,
+					))
+				}
```
krancour, line 679: I love that this is being done in the same fanned-out goroutines we're using for cleaning up the substrate.

krancour (approved):
🔥 🔥 🔥

Issue #1194, opened 2020-12-11
Add healthchecks to 2.0


krancour:
The API server, scheduler, and observer all need health check capabilities built in.

The health checks should also account for the health of connections to backing services.

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1304

PR #1297, opened 2021-02-23
feat(ci): add test-integration job

vdice:

* Adds the test-integration job to CI
* Shuffles around some make targets to facilitate

Ref associated PR bumping kind image/KindJob in https://github.com/brigadecore/brigade-utils/pull/51

**If applicable**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -331,6 +331,7 @@ hack-deploy:
 	helm upgrade brigade charts/brigade \
 		--install \
 		--namespace brigade \
+		--wait \
 		--set apiserver.image.repository=$(DOCKER_IMAGE_PREFIX)apiserver \
 		--set apiserver.image.tag=$(IMMUTABLE_DOCKER_TAG) \
 		--set apiserver.image.pullPolicy=Always \
--- a/Makefile
+++ b/Makefile
@@ -324,6 +324,8 @@ hack-push-images: hack-push-apiserver hack-push-scheduler hack-push-observer hac
 hack-push-%: hack-build-%
 	docker push $(DOCKER_IMAGE_PREFIX)$*:$(IMMUTABLE_DOCKER_TAG)
 
+IMAGE_PULL_POLICY ?= Always
+
 .PHONY: hack-deploy
 hack-deploy:
 	kubectl get namespace brigade || kubectl create namespace brigade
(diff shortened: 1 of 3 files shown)


vdice:
Currently troubleshooting the integration test failure after the successful deploy; tests ran fine on the AKS cluster I used while working on this branch.

Makefile
```diff
@@ -357,3 +366,12 @@ hack-expose-apiserver:
...
 	@kill -TERM $$(cat /tmp/brigade-apiserver.PID)
+
+# Convenience targets for loading images into a KinD cluster
+.PHONY: hack-load-images
+hack-load-images: load-apiserver load-scheduler load-observer load-logger-linux load-git-initializer load-worker
+
+load-%:
+	@echo "Loading $(DOCKER_IMAGE_PREFIX)$*:$(IMMUTABLE_DOCKER_TAG)"
+	@kind load docker-image $(DOCKER_IMAGE_PREFIX)$*:$(IMMUTABLE_DOCKER_TAG) \
+			|| echo >&2 "kind not installed or error loading image: $(DOCKER_IMAGE_PREFIX)$*:$(IMMUTABLE_DOCKER_TAG)"
```
krancour, line 377: Whoa! I didn't know about `kind load`. That is really neat and useful and now I am contemplating ways I can use this to streamline my own local workflow!

brigade.js
```diff
@@ -72,6 +72,32 @@ const lintJSJob = (e, p) => {
...
+const testIntegrationJob = (e, p) => {
+  let kind = new KindJob(testIntegrationJobName);
+  kind.mountPath = localPath;
+  kind.tasks.push(
+    // Install git and golang deps
+    "apk add --update --no-cache git gcc musl-dev",
+    // Install helm
+    `curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 \
+      && chmod 700 get_helm.sh \
+      && ./get_helm.sh`,
```
krancour, line 85: Non-blocking, but do we want to consider bundling all relevant tools into an image?

brigade.js
```diff
@@ -72,6 +72,32 @@ const lintJSJob = (e, p) => {
...
+    `cd ${localPath}`,
+    // It would be more efficient to mount built images from associated
+    // build jobs; however, they currently use kaniko, which doesn't preserve
+    // images in the local Docker cache.
+    "make hack-build-images hack-load-images",
```
krancour, line 90: This entire comment is non-blocking...

A `KindJob` must be using DooD to do this?

DooD weirds me out because it clutters the image cache in the k8s node we're running on.

I wonder if we could use a kind node that has a registry built in, and then with `DOCKER_REGISTRY` and `IMAGE_PULL_POLICY ` set correctly, `make push-images hack-deploy test-integration` would do the trick?

The benefit would be that when the kind node node goes away, so to do all the images that we don't care about keeping.

Again... non-blocking.

krancour (approved):
All my feedback is non-blocking. LGTM

radu-matei (approved):

brigade.js
vdice: Good concern.  The KindJob actually sets up DinD: https://github.com/brigadecore/brigade-utils/blob/master/src/kind.ts#L73  So, at least for these Brigade 2 images, no image cache-crowding on the k8s node will occur.  (For running kind itself in the KindJob, there are some host mounts necessary).

It could still be interesting to see if launching an internal registry in Kind helps efficiency, etc.

brigade.js
vdice: Yea, what do we think?  PR to bake them into the image that brigade-utils creates for the default KindJob image? Or create a new one in brigadecore that builds on that default image?  I think I might prefer the former, since this job (and others in brigade-utils) are really pretty opinionated for our uses anyhow....

krancour: No strong opinion on which way to go.

I do have an opinion about us having opinions tho...

I think that with both 1.x and now 2.x, we're uncovering a lot of useful patterns that others may seek to emulate, and may be worth documenting (or documenting further) and if we were to curate some useful images that facilitate those patterns, people would probably use them.

brigade.js
krancour: Sorry... I had a mental lapse. For some reason I couldn't see how this could be working with DinD, but I see it now.

DinD doesn't both me as much.

Forget I asked about this.

vdice:
This is looking satisfyingly consistent with infra cluster upgrades and a small pause added after deploy right before running the tests.  Note that the latter is scheduled for removal once we have more consistent deploy readiness via https://github.com/brigadecore/brigade/issues/1194.  Merging.

Issue #1301, opened 2021-02-24
Regex for validating event creator role's scope needs to be relaxed


krancour:
Currently it does not permit a source that looks like a URL (for instance `github.com/brigadecore/brigade-github-gateway/v2`), but that should be allowed, as the most realistic value that gateways can use to identify themselves as an event source _without_ colliding with other gateways is to use a domain or URL controlled by the project that produced the gateway in question.

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1302

PR #1302, opened 2021-02-24
fix(role-assignment.json): allow for url/domain-like scopes

vdice:

* Updates the event scope to allow for domain/url-like sources.

I'm assuming adding `.` and `/` chars are sufficient for our needs and we didn't need to allow for `:` or other chars for full url support.

Closes https://github.com/brigadecore/brigade/issues/1301

**If applicable**:

### DIFF
--- a/v2/apiserver/schemas/role-assignment.json
+++ b/v2/apiserver/schemas/role-assignment.json
@@ -50,7 +50,7 @@
 				"scope": {
 					"type": "string",
 					"description": "The event source this role should be scoped to",
-					"pattern": "^[a-z][a-z\\d-]*[a-z\\d]$",
+					"pattern": "^[a-z][a-z\\d./-]*[a-z\\d]$",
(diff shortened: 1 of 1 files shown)


v2/apiserver/schemas/role-assignment.json
```diff
@@ -50,7 +50,7 @@
 				"scope": {
 					"type": "string",
 					"description": "The event source this role should be scoped to",
-					"pattern": "^[a-z][a-z\\d-]*[a-z\\d]$",
+					"pattern": "^[a-z][a-z\\d./-]*[a-z\\d]$",
```
krancour, line 53: This still won't allow a lot of characters that are valid in URLs like _ and ?, but I think that's ok. This is a pretty reasonable regex for the purpose.

I do think we should allow uppercase alphas though.

krancour (approved):
LGTM. Thanks!

PR #1304, opened 2021-03-01
feat(*): add health checks

vdice:

* Adds health checks to the major components of Brigade
    - apiserver healthz endpoint to check db and queue comms
    - apiserver ping endpoint for quick comms test (used here by the observer)
    - scheduler reads/acks msgs off queue sent by apiserver (as spurred by healthz probe)

Closes https://github.com/brigadecore/brigade/issues/1194

**If applicable**:

### DIFF
--- a/brigade.js
+++ b/brigade.js
@@ -92,10 +92,6 @@ const testIntegrationJob = (e, p) => {
     // 'Always', the images loaded manually into kind will be ignored and the
     // pods will attempt to pull from remote registries.
     "IMAGE_PULL_POLICY=IfNotPresent make hack-deploy",
-    // TODO: Remove sleep when deploys produce dependably ready Brigade 2
-    // instances. Currently, there seems to be some delays for the messaging
-    // bus to be fully ready.
-    "sleep 60",
     "make test-integration",
   );
   return kind;
--- a/charts/brigade/templates/apiserver/deployment.yaml
+++ b/charts/brigade/templates/apiserver/deployment.yaml
@@ -142,6 +142,23 @@ spec:
           value: {{ .Values.worker.image.pullPolicy }}
         - name: WORKSPACE_STORAGE_CLASS
           value: {{ .Values.worker.workspaceStorageClass }}
+        ports:
+        - name: healthz
+          containerPort: 8080
+        livenessProbe:
+          httpGet:
+            port: healthz
+            path: /healthz
+            scheme: HTTPS
+          initialDelaySeconds: 10
+          periodSeconds: 10
+        startupProbe:
+          httpGet:
+            port: healthz
+            path: /healthz
+            scheme: HTTPS
+          failureThreshold: 10
+          periodSeconds: 10
         {{- if .Values.apiserver.tls.enabled }}
         volumeMounts:
         - name: cert
--- a/charts/brigade/templates/apiserver/deployment.yaml
+++ b/charts/brigade/templates/apiserver/deployment.yaml
@@ -152,12 +152,19 @@ spec:
             scheme: HTTPS
           initialDelaySeconds: 10
           periodSeconds: 10
+        readinessProbe:
+          httpGet:
+            port: healthz
+            path: /healthz
+            scheme: HTTPS
+          initialDelaySeconds: 10
+          periodSeconds: 10
         startupProbe:
           httpGet:
             port: healthz
             path: /healthz
             scheme: HTTPS
-          failureThreshold: 10
+          failureThreshold: 30
           periodSeconds: 10
         {{- if .Values.apiserver.tls.enabled }}
         volumeMounts:
--- a/charts/brigade/templates/apiserver/deployment.yaml
+++ b/charts/brigade/templates/apiserver/deployment.yaml
@@ -149,21 +149,27 @@ spec:
           httpGet:
             port: healthz
             path: /healthz
+            {{- if .Values.apiserver.tls.enabled }}
             scheme: HTTPS
+            {{- end }}
           initialDelaySeconds: 10
           periodSeconds: 10
         readinessProbe:
           httpGet:
             port: healthz
             path: /healthz
+            {{- if .Values.apiserver.tls.enabled }}
             scheme: HTTPS
+            {{- end }}
           initialDelaySeconds: 10
           periodSeconds: 10
         startupProbe:
           httpGet:
             port: healthz
             path: /healthz
+            {{- if .Values.apiserver.tls.enabled }}
             scheme: HTTPS
+            {{- end }}
           failureThreshold: 30
           periodSeconds: 10
         {{- if .Values.apiserver.tls.enabled }}
--- a/charts/brigade/templates/artemis/primary-stateful-set.yaml
+++ b/charts/brigade/templates/artemis/primary-stateful-set.yaml
@@ -88,6 +88,11 @@ spec:
             port: core
           initialDelaySeconds: 10
           periodSeconds: 10
+        startupProbe:
+          tcpSocket:
+            port: core
+          failureThreshold: 10
+          periodSeconds: 10
         volumeMounts:
         - name: config-overrides
           mountPath: /var/lib/artemis/etc-override
--- a/charts/brigade/templates/artemis/primary-stateful-set.yaml
+++ b/charts/brigade/templates/artemis/primary-stateful-set.yaml
@@ -91,7 +91,7 @@ spec:
         startupProbe:
           tcpSocket:
             port: core
-          failureThreshold: 10
+          failureThreshold: 30
           periodSeconds: 10
         volumeMounts:
         - name: config-overrides
--- a/charts/brigade/templates/artemis/secondary-stateful-set.yaml
+++ b/charts/brigade/templates/artemis/secondary-stateful-set.yaml
@@ -88,6 +88,11 @@ spec:
             port: core
           initialDelaySeconds: 10
           periodSeconds: 10
+        startupProbe:
+          tcpSocket:
+            port: core
+          failureThreshold: 10
+          periodSeconds: 10
         volumeMounts:
         - name: config-overrides
           mountPath: /var/lib/artemis/etc-override
--- a/charts/brigade/templates/artemis/secondary-stateful-set.yaml
+++ b/charts/brigade/templates/artemis/secondary-stateful-set.yaml
@@ -91,7 +91,7 @@ spec:
         startupProbe:
           tcpSocket:
             port: core
-          failureThreshold: 10
+          failureThreshold: 30
           periodSeconds: 10
         volumeMounts:
         - name: config-overrides
--- a/charts/brigade/templates/artemis/stateful-set.yaml
+++ b/charts/brigade/templates/artemis/stateful-set.yaml
@@ -72,7 +72,7 @@ spec:
         startupProbe:
           tcpSocket:
             port: core
-          failureThreshold: 10
+          failureThreshold: 30
           periodSeconds: 10
         volumeMounts:
         - name: config-overrides
--- a/charts/brigade/templates/artemis/stateful-set.yaml
+++ b/charts/brigade/templates/artemis/stateful-set.yaml
@@ -69,6 +69,11 @@ spec:
             port: core
           initialDelaySeconds: 10
           periodSeconds: 10
+        startupProbe:
+          tcpSocket:
+            port: core
+          failureThreshold: 10
+          periodSeconds: 10
         volumeMounts:
         - name: config-overrides
           mountPath: /var/lib/artemis/etc-override
--- a/sdk/v2/core/healthcheck.go
+++ b/sdk/v2/core/healthcheck.go
@@ -1,0 +1,43 @@
+package core
+
+import (
+	"context"
+	"net/http"
+
+	rm "github.com/brigadecore/brigade/sdk/v2/internal/restmachinery"
+	"github.com/brigadecore/brigade/sdk/v2/restmachinery"
+)
+
+// HealthcheckClient is the specialized client for accessing healthchecks on
+// the Brigade API.
+type HealthcheckClient interface {
+	// Ping pings the API Server
+	Ping(ctx context.Context) error
+}
+
+type healthcheckClient struct {
+	*rm.BaseClient
+}
+
+// NewHealthcheckClient returns a specialized client for accessing the
+// healthcheck API.
+func NewHealthcheckClient(
+	apiAddress string,
+	apiToken string,
+	opts *restmachinery.APIClientOptions,
+) HealthcheckClient {
+	return &healthcheckClient{
+		BaseClient: rm.NewBaseClient(apiAddress, apiToken, opts),
+	}
+}
+
+func (h *healthcheckClient) Ping(ctx context.Context) error {
+	return h.ExecuteRequest(
+		ctx,
+		rm.OutboundRequest{
+			Method:      http.MethodGet,
+			Path:        "ping",
+			SuccessCode: http.StatusOK,
+		},
+	)
+}
--- a/sdk/v2/system/api_client.go
+++ b/sdk/v2/system/api_client.go
@@ -1,4 +1,4 @@
-package core
+package system
 
 import (
 	"context"
@@ -8,31 +8,30 @@ import (
 	"github.com/brigadecore/brigade/sdk/v2/restmachinery"
 )
 
-// HealthcheckClient is the specialized client for accessing healthchecks on
-// the Brigade API.
-type HealthcheckClient interface {
+// APIClient is the client for system checks involving the Brigade API.
+type APIClient interface {
 	// Ping pings the API Server
 	Ping(ctx context.Context) error
 }
 
-type healthcheckClient struct {
+type apiClient struct {
 	*rm.BaseClient
 }
 
-// NewHealthcheckClient returns a specialized client for accessing the
-// healthcheck API.
-func NewHealthcheckClient(
+// NewAPIClient returns a client to access system-related Brigade API
+// endpoints.
+func NewAPIClient(
 	apiAddress string,
 	apiToken string,
 	opts *restmachinery.APIClientOptions,
-) HealthcheckClient {
-	return &healthcheckClient{
+) APIClient {
+	return &apiClient{
 		BaseClient: rm.NewBaseClient(apiAddress, apiToken, opts),
 	}
 }
 
-func (h *healthcheckClient) Ping(ctx context.Context) error {
-	return h.ExecuteRequest(
+func (a *apiClient) Ping(ctx context.Context) error {
+	return a.ExecuteRequest(
 		ctx,
 		rm.OutboundRequest{
 			Method:      http.MethodGet,
--- a/sdk/v2/system/api_client.go
+++ b/sdk/v2/system/api_client.go
@@ -8,10 +8,16 @@ import (
 	"github.com/brigadecore/brigade/sdk/v2/restmachinery"
 )
 
+// PingResponse represents the expected response object returned by the
+// API Server's v2/ping endpoint
+type PingResponse struct {
+	Version string
+}
+
 // APIClient is the client for system checks involving the Brigade API.
 type APIClient interface {
 	// Ping pings the API Server
-	Ping(ctx context.Context) error
+	Ping(ctx context.Context) (PingResponse, error)
 }
 
 type apiClient struct {
@@ -30,13 +36,15 @@ func NewAPIClient(
 	}
 }
 
-func (a *apiClient) Ping(ctx context.Context) error {
-	return a.ExecuteRequest(
+func (a *apiClient) Ping(ctx context.Context) (PingResponse, error) {
+	pingResponse := PingResponse{}
+	return pingResponse, a.ExecuteRequest(
 		ctx,
 		rm.OutboundRequest{
 			Method:      http.MethodGet,
-			Path:        "ping",
+			Path:        "v2/ping",
 			SuccessCode: http.StatusOK,
+			RespObj:     &pingResponse,
 		},
 	)
 }
--- a/v2/apiserver/config.go
+++ b/v2/apiserver/config.go
@@ -22,9 +22,9 @@ import (
 	"golang.org/x/oauth2"
 )
 
-// databaseConfig returns a *mongo.Database connection based on configuration
-// obtained from environment variables.
-func databaseConfig(ctx context.Context) (*mongo.Database, error) {
+// databaseConnection returns a *mongo.Database connection based on
+// configuration obtained from environment variables.
+func databaseConnection(ctx context.Context) (*mongo.Database, error) {
 	hosts, err := os.GetRequiredEnvVar("DATABASE_HOSTS")
 	if err != nil {
 		return nil, err
--- a/v2/apiserver/config.go
+++ b/v2/apiserver/config.go
@@ -22,9 +22,9 @@ import (
 	"golang.org/x/oauth2"
 )
 
-// database returns a *mongo.Database connection based on configuration obtained
-// from environment variables.
-func database(ctx context.Context) (*mongo.Database, error) {
+// databaseConfig returns a *mongo.Database connection based on configuration
+// obtained from environment variables.
+func databaseConfig(ctx context.Context) (*mongo.Database, error) {
 	hosts, err := os.GetRequiredEnvVar("DATABASE_HOSTS")
 	if err != nil {
 		return nil, err
--- a/v2/apiserver/internal/core/rest/healthcheck_endpoints.go
+++ b/v2/apiserver/internal/core/rest/healthcheck_endpoints.go
@@ -1,0 +1,94 @@
+package rest
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"time"
+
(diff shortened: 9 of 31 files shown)


brigade.js
```diff
@@ -92,10 +92,6 @@ const testIntegrationJob = (e, p) => {
     // 'Always', the images loaded manually into kind will be ignored and the
     // pods will attempt to pull from remote registries.
     "IMAGE_PULL_POLICY=IfNotPresent make hack-deploy",
-    // TODO: Remove sleep when deploys produce dependably ready Brigade 2
-    // instances. Currently, there seems to be some delays for the messaging
-    // bus to be fully ready.
-    "sleep 60",
```
krancour, line 98: Oh... I love any PR that lets us get rid of something like this!

charts/brigade/templates/apiserver/deployment.yaml
```diff
@@ -138,6 +138,23 @@ spec:
...
+            scheme: HTTPS
+          initialDelaySeconds: 10
+          periodSeconds: 10
+        startupProbe:
+          httpGet:
+            port: healthz
+            path: /healthz
+            scheme: HTTPS
+          failureThreshold: 10
+          periodSeconds: 10
```
krancour, line 157: Should we have a readiness probe as well?

krancour: I wasn't previously familiar with startup probes. #TIL

Glancing at [the docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes) for my own edification, I noticed this:

> The trick is to set up a startup probe with the same command, HTTP or TCP check, with a failureThreshold * periodSeconds long enough to cover the worse case startup time.

So, I'm wondering if the failure threshold here should be higher than that of the liveness probe-- perhaps even substantially higher.

charts/brigade/templates/artemis/primary-stateful-set.yaml
```diff
@@ -88,6 +88,11 @@ spec:
             port: core
           initialDelaySeconds: 10
           periodSeconds: 10
+        startupProbe:
+          tcpSocket:
+            port: core
+          failureThreshold: 10
+          periodSeconds: 10
```
krancour, line 95: Same comment as [here](https://github.com/brigadecore/brigade/pull/1304/files#r585681169) though perhaps even more relevant in relation to Artemis?

charts/brigade/templates/artemis/secondary-stateful-set.yaml
```diff
@@ -88,6 +88,11 @@ spec:
             port: core
           initialDelaySeconds: 10
           periodSeconds: 10
+        startupProbe:
+          tcpSocket:
+            port: core
+          failureThreshold: 10
+          periodSeconds: 10
```
krancour, line 95: Same comment as [here](https://github.com/brigadecore/brigade/pull/1304/files#r585682226)

charts/brigade/templates/artemis/stateful-set.yaml
```diff
@@ -69,6 +69,11 @@ spec:
             port: core
           initialDelaySeconds: 10
           periodSeconds: 10
+        startupProbe:
+          tcpSocket:
+            port: core
+          failureThreshold: 10
+          periodSeconds: 10
```
krancour, line 76: Same comment as [here](https://github.com/brigadecore/brigade/pull/1304/files#r585682226)

sdk/v2/core/healthcheck.go
```diff
@@ -0,0 +1,43 @@
...
+	rm "github.com/brigadecore/brigade/sdk/v2/internal/restmachinery"
+	"github.com/brigadecore/brigade/sdk/v2/restmachinery"
+)
+
+// HealthcheckClient is the specialized client for accessing healthchecks on
+// the Brigade API.
+type HealthcheckClient interface {
+	// Ping pings the API Server
+	Ping(ctx context.Context) error
+}
```
krancour, line 16: A thought...

We should probably at some point also expose an endpoint that returns the server's version information.

I'm wondering if it might make sense to lump `Ping()` and `Version()` together in one client-- maybe one named "system" API client or some such?

To clarify-- not asking for `Version()` to be added to this PR. Just trying to plan ahead is all.

v2/apiserver/config.go
```diff
@@ -22,9 +22,9 @@ import (
 	"golang.org/x/oauth2"
 )
 
-// database returns a *mongo.Database connection based on configuration obtained
-// from environment variables.
-func database(ctx context.Context) (*mongo.Database, error) {
+// databaseConfig returns a *mongo.Database connection based on configuration
+// obtained from environment variables.
+func databaseConfig(ctx context.Context) (*mongo.Database, error) {
```
krancour, line 27: I see this change was made for consistency with the other function names probably, but it might be misleading as it doesn't return configuration, rather it returns a database connection.

v2/apiserver/internal/lib/queue/writer.go
```diff
@@ -23,8 +23,16 @@ type WriterFactory interface {
...
 	// function when they are done with the Writer.
 	Close(context.Context) error
 }
+
+// MessageOptions represent options for the message being written via
+// the Writer.
+type MessageOptions struct {
+	// Ephemeral indicates whether or not the message should be durable/persisted
+	// (false) or not durable/persisted (true)
+	Ephemeral bool
```
krancour, line 37: How would you feel about changing this to `Durable` so that we don't have to translate !Ephemeral == Durable in our heads?

fwiw, I think Durable should default to false (like Ephemeral does now) i.e. This would be the exact opposite behavior of what we have now. You'd need to _ask_ for a message to be durable. That tracks well with the fact that durability isn't a default on Artemis' end either. (And in general, durability isn't a default for message-oriented middleware.)

v2/observer/healthcheck.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/brigadecore/brigade/v2/internal/retries"
+)
+
+// runHealthcheckLoop checks Observer -> API Server comms
+func (o *observer) runHealthcheckLoop(ctx context.Context) {
+	ticker := time.NewTicker(o.config.healthcheckInterval)
+	defer ticker.Stop()
+
+	for {
+		if err := retries.ManageRetries(
```
krancour, line 16: Retries on GETs are already built into the SDK... so you can lose the retry loop here.

charts/brigade/templates/apiserver/deployment.yaml
vdice: Yes, will add a readinessProbe as well.

Re: StartupProbe; Sure...`30` sound good?

charts/brigade/templates/artemis/primary-stateful-set.yaml
vdice: The Artemis stateful-set yamls indeed already came with liveness and readiness probes; just added startupProbes to each in this PR.

sdk/v2/core/healthcheck.go
vdice: Yea, I like this.  `SystemClient`?

v2/apiserver/config.go
vdice: That's reasonable.  How about `databaseConnection`?

v2/apiserver/internal/lib/queue/writer.go
vdice: Ah, can do.  Went this approach to maintain defaults with what they were, but if we are good with flipping them, agreed, sticking to Durable would be best.  I'll do this.

v2/observer/healthcheck.go
vdice: Oh yea, you'd mentioned this and I promptly forgot!  Will remove.

charts/brigade/templates/apiserver/deployment.yaml
krancour: At 10 second intervals-- that's a five minute grace period to come up. That sounds plenty forgiving to me.

👍

charts/brigade/templates/artemis/primary-stateful-set.yaml
krancour: Sorry... the comment here was really about needing a higher failure threshold.

charts/brigade/templates/artemis/secondary-stateful-set.yaml
krancour: Sorry... the comment here was really about needing a higher failure threshold.

charts/brigade/templates/artemis/stateful-set.yaml
krancour: Sorry... the comment here was really about needing a higher failure threshold.

sdk/v2/core/healthcheck.go
krancour: Yes... or even call it `APIClient` and stick it in the `system` package.

krancour: `core` probably isn't the right place for it. `core` is just for projects, events, etc.

v2/apiserver/config.go
krancour: sgtm

v2/apiserver/internal/lib/queue/writer.go
krancour: > Went this approach to maintain defaults with what they were

I figured as much. I don't feel married to the the way things were.

v2/observer/healthcheck.go
krancour: No worries. Figure it's always refreshing when we get to delete code.

v2/apiserver/internal/core/rest/healthcheck_endpoints.go
```diff
@@ -0,0 +1,92 @@
...
+
+// ping returns the api server version and http.StatusOK.  This is handy for
+// auxiliary components to verify their connectivity.
+func (h *HealthcheckEndpoints) ping(w http.ResponseWriter, r *http.Request) {
+	restmachinery.ServeRequest(
+		restmachinery.InboundRequest{
+			W: w,
+			R: r,
+			EndpointLogic: func() (interface{}, error) {
+				return fmt.Sprintf(`{"version": "%s"}`, h.APIServerVersion), nil
```
krancour, line 87: Does our server side rest machinery handle it correctly when we return a string instead of a struct?

v2/apiserver/internal/core/rest/healthcheck_endpoints.go
```diff
@@ -0,0 +1,92 @@
+package rest
```
krancour, line 1: Should we move this to `system/rest`?

krancour:
Looking good @vdice.

I do think we still have an infinite retry in this function:

```
func (w *writerFactory) NewWriter(queueName string) (queue.Writer, error)
```

That's could end up making health check failures spin indefinitely instead of returning.

v2/apiserver/internal/core/rest/healthcheck_endpoints.go
vdice: Good call; the output was kinda weird w/ escaped quotes... updated to return an actual struct, which looks much nicer when returned.

vdice:
> I do think we still have an infinite retry in this function:
func (w *writerFactory) NewWriter(queueName string) (queue.Writer, error)
That's could end up making health check failures spin indefinitely instead of returning.

What are your thinking as an approach for avoiding the infinite retry in that function?

krancour:
@vdice 

> What are you thinking as an approach for avoiding the infinite retry in that function?

Can we make out usual retry strategy work? I can't really remember why I didn't use it here in the first case... it might have been laziness or "it's just a prototype" thinking. Of course, we're past that now.

krancour:
At a glance, the latest changes look great... I'm stopping short of LGTM only because I'm fried and this deserves fresh eyes tomorrow morning.

v2/observer/observer.go
```diff
@@ -110,6 +125,17 @@ func (o *observer) run(ctx context.Context) error {
...
 	var err error
 	select {
+	// In essence, this comprises the Observer's "healthcheck" logic.
+	// Whenever we receive an error on this channel, we cancel the context and
+	// shut down.  E.g., if one loop fails, everything fails.
+	// This includes:
+	//   1. an error pinging the API server endpoint
+	//      (Observer <-> API comms)
+	//   2. TODO
+	//      (Observer <-> K8s comms)
```
krancour, line 135: Out of scope for this PR or no?

v2/observer/observer.go
```diff
@@ -110,6 +125,17 @@ func (o *observer) run(ctx context.Context) error {
...
+	//
+	// Note: Currently, errors updating or cleaning up worker or job statuses
+	//       are handled by o.errFn, which currently simply logs the error
```
krancour, line 138: fwiw, I think that's fine. We already have a period ping of the API server, so any failure here is less likely to be "zomg the API server is down" and more likely to be a specific issue with the event at hand.

v2/scheduler/scheduler.go
```diff
@@ -140,6 +151,16 @@ func (s *scheduler) run(ctx context.Context) error {
...
+	// In essence, this comprises the Scheduler's "healthcheck" logic.
+	// Whenever we receive an error on this channel, we cancel the context and
+	// shut down.  E.g., if one loop fails, everything fails.
+	// This includes:
+	//   1. an error listing projects at the start of the project loop
+	//      (Scheduler <-> API comms)
+	//   2. an error instantiating a reader in any of the worker/job loops
+	//   3. an error instantiating a reader, reading a message or acking a
+	//      message in the healhcheck loop
+	//      (Scheduler <-> Messaging queue comms)
```
krancour, line 163: Maybe worth explaining why we need no. 3, when no. 2 might seem to have the bases covered?

My sense is that no. 3 detects issues when the rest of the system is quiet and might not "naturally" encounter a problem.

v2/apiserver/internal/system/rest/healthcheck_endpoints.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"github.com/brigadecore/brigade/v2/apiserver/internal/lib/queue"
+	"github.com/brigadecore/brigade/v2/apiserver/internal/lib/restmachinery"
+	"github.com/gorilla/mux"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+// HealthcheckEndpoints implements restmachinery.Endpoints to provide
+// Healthcheck-related URL --> action mappings to a restmachinery.Server.
+type HealthcheckEndpoints struct {
```
krancour, line 17: Naming nit. I might prefer if this were `SystemEndpoints` rather than `HealthcheckEndpoints` since it handles pings as well and those aren't really used exclusively for health-related purposes. On the flip side, it feels fine to have health check requests responded to by "system endpoints."

Again... a nit. Non-blocking. Dealer's choice on whether to act.

v2/apiserver/internal/system/rest/healthcheck_endpoints.go
```diff
@@ -0,0 +1,95 @@
...
+				defer writer.Close(ctx)
+				if err := writer.Write(
+					ctx,
+					"ping",
+					&queue.MessageOptions{},
+				); err != nil {
+					return false, errors.New("error writing to messaging queue")
+				}
+
+				return []byte("ok"), nil
```
krancour, line 71: This may require some more research. I have a vague recollection that k8s prefers heath checks to return an empty JSON object. A quick search isn't uncovering anything that backs that up, but I swear I've read that _somewhere_. I'll keep looking.

v2/apiserver/internal/system/rest/healthcheck_endpoints.go
```diff
@@ -0,0 +1,95 @@
...
+func (h *HealthcheckEndpoints) Register(router *mux.Router) {
+	// Get /healthz
+	router.HandleFunc(
+		"/healthz",
+		h.healthz,
+	).Methods(http.MethodGet)
+
+	// Get /ping
+	router.HandleFunc(
+		"/ping",
```
krancour, line 34: Since the ping returns version information, do you think we should version this path? i.e. Use `/v2/ping`. This could provide two benefits:

1. Client can easily assert that the API server it's talking to is a v2 API server.
2. We get cover to change the shape of the ping response in v3 and beyond.

sdk/v2/system/api_client.go
```diff
@@ -0,0 +1,42 @@
...
+	apiAddress string,
+	apiToken string,
+	opts *restmachinery.APIClientOptions,
+) APIClient {
+	return &apiClient{
+		BaseClient: rm.NewBaseClient(apiAddress, apiToken, opts),
+	}
+}
+
+func (a *apiClient) Ping(ctx context.Context) error {
```
krancour, line 33: On the server side, the ping returns version information (which I think was a great idea, btw), but we're not doing anything with it here. Perhaps we should return a `PingResponse` object that has a `Version` field?

v2/apiserver/internal/system/rest/healthcheck_endpoints.go
```diff
@@ -0,0 +1,95 @@
...
+
+// ping returns the api server version and http.StatusOK.  This is handy for
+// auxiliary components to verify their connectivity.
+func (h *HealthcheckEndpoints) ping(w http.ResponseWriter, r *http.Request) {
+	restmachinery.ServeRequest(
+		restmachinery.InboundRequest{
+			W: w,
+			R: r,
+			EndpointLogic: func() (interface{}, error) {
+				return struct {
```
krancour, line 86: We should probably make this is a versioned response... i.e. add the a `MarshalJSON(...)` function to weave in type metadata when the response is marshaled.

v2/scheduler/healthcheck.go
```diff
@@ -0,0 +1,73 @@
...
+			// TODO: do we want to allow for a certain threshold of retries here
+			// before sending an error (and hence shutting down the Scheduler)?
+
+			// Get the next message on the queue
+			msg, err := healthcheckReader.Read(ctx)
+			if err != nil {
+				// If we can't read from the queue, we consider it fatal
+				// (Scheduler cancels the context when an error is received)
+				s.errCh <- err
+				continue outerLoop // Try again with a new reader
```
krancour, line 60: What's the point of trying again with a new reader if the error we've experienced and placed on the error channel is counted as fatal? Can we simplify by losing the outer loop?

v2/apiserver/internal/lib/queue/amqp/writer.go
```diff
@@ -93,34 +93,49 @@ func (w *writerFactory) NewWriter(queueName string) (queue.Writer, error) {
...
+			if amqpSession, err = w.amqpClient.NewSession(); err != nil {
+				// Assume this happened because the existing connection is no good. Try
+				// to reconnect.
+				if err = w.connectFn(); err != nil {
+					// The connection function handles its own retries. If we got an error
+					// here, it's pretty serious. Bail.
+					return true, err
+				}
+				// We're reconnected now, so retry getting a session again.
+				return true, err
```
krancour, line 123: If we successfully reconnected, `err` was effectively handled. Return a `nil` here instead.

v2/apiserver/internal/lib/queue/amqp/writer.go
```diff
@@ -93,34 +93,49 @@ func (w *writerFactory) NewWriter(queueName string) (queue.Writer, error) {
...
+				amqpSession.Close(context.TODO()) // nolint: errcheck
+			}
+
+			if amqpSession, err = w.amqpClient.NewSession(); err != nil {
+				// Assume this happened because the existing connection is no good. Try
+				// to reconnect.
+				if err = w.connectFn(); err != nil {
+					// The connection function handles its own retries. If we got an error
+					// here, it's pretty serious. Bail.
+					return true, err
```
krancour, line 120: Probably return `false` here to prevent more retries. We know we can't get a connection right now, so more attempts to get a session will fail.

v2/apiserver/internal/lib/queue/amqp/writer.go
```diff
@@ -93,34 +93,49 @@ func (w *writerFactory) NewWriter(queueName string) (queue.Writer, error) {
...
+				return true, err
+			}
+			return false, nil
+		},
+	); err != nil {
+		return nil, errors.Wrapf(
+			err,
+			"error attempting to create writer after %d retries",
+			maxRetryCount,
+		)
```
krancour, line 138: iirc correctly (you may want to verify), the retry manager already returns an error that looks like this, indicating how many retries were performed, so you can probably just return the error without doing anything to it.

v2/apiserver/internal/system/rest/healthcheck_endpoints.go
vdice: Good call; renamed 👍

v2/scheduler/scheduler.go
vdice: Added further context; indeed, your sense above is exactly it.

v2/observer/observer.go
vdice: Thanks for the reminder; healthcheck added.

v2/apiserver/internal/system/rest/healthcheck_endpoints.go
vdice: Interestingly, performing a GET on the `/healthz` endpoint of the K8s api server's REST client  - at least via `DoRaw()` - returns `[]byte("ok")` if all is well (Discovered this via the recent addition of adding such a check to the observer)

v2/apiserver/internal/system/rest/healthcheck_endpoints.go
vdice: Excellent foresight.  Updated.

v2/scheduler/healthcheck.go
vdice: Good point; simplified.  PTAL when convenient.

charts/brigade/templates/apiserver/deployment.yaml
```diff
@@ -142,6 +142,30 @@ spec:
...
         - name: WORKSPACE_STORAGE_CLASS
           value: {{ .Values.worker.workspaceStorageClass }}
+        ports:
+        - name: healthz
+          containerPort: 8080
+        livenessProbe:
+          httpGet:
+            port: healthz
+            path: /healthz
+            scheme: HTTPS
```
krancour, line 152: I think this would be dependent on `.Values.apiserver.tls.enabled`

This applies in some other spots also.

v2/apiserver/internal/system/rest/system_endpoints.go
```diff
@@ -0,0 +1,116 @@
...
+			EndpointLogic: func() (interface{}, error) {
+				return PingResponse{Version: h.APIServerVersion}, nil
+			},
+
+			SuccessCode: http.StatusOK,
+		},
+	)
+}
+
+// PingResponse represents the esponse object returned by the ping endpoint
```
krancour, line 96: s/esponse/response/

v2/apiserver/main.go
```diff
@@ -243,6 +245,11 @@ func main() {
 					),
 					Service: eventsService,
 				},
+				&systemREST.SystemEndpoints{
+					APIServerVersion: version.Version(),
+					DatabaseClient:   database.Client(),
+					WriterFactory:    queueWriterFactory,
+				},
```
krancour, line 252: Total nit: The list of endpoints passed to the server was alphabetical prior to this. Might want to move this down, but it's not the end of the world if you don't.

v2/observer/healthcheck.go
```diff
@@ -0,0 +1,43 @@
...
+			// Check Observer -> K8s
+			// Not actually capturing response; just want to verify our API call
+			// is successful
+			if _, err := o.checkK8sAPIServer(ctx); err != nil {
+				o.errCh <- errors.Wrap(
+					err,
+					"error checking K8s API server connectivity",
+				)
+			}
+		default:
```
krancour, line 40: This default case should go away. If it's here, you're going to continuously cycle through the loop, eating CPU. With out it, this goroutine will idle while waiting to receive from one of the other channels.

v2/scheduler/healthcheck.go
```diff
@@ -0,0 +1,46 @@
...
+		}
+
+		if msg != nil {
+			if err = msg.Ack(ctx); err != nil {
+				// If we can't ack a message, we consider it fatal
+				// (Scheduler cancels the context when an error is received)
+				s.errCh <- err
+			}
+		}
+		continue // Next message
```
krancour, line 43: Don't need `continue` here.

v2/scheduler/scheduler.go
krancour: 👍

krancour:
@vdice looking good. I added a few more comments. Couple nits and typos-- only a couple comments that are really consequential.

I have a question about the ping endpoint. It's currently versioned under `/v2` and returns API server version info...

As a client, I'd need to know I'm talking to v2 in order to ping. But I also don't know for sure that I'm talking to v2 until I've gotten version info _from_ a ping-- sort of a chicken and egg problem. (At least in theory. In practice, clients are probably more likely to assume they're talking to v2 and call it a day.)

Would you have any thoughts about providing an _unversioned_ endpoint for retrieving API server version _separate_ from the ping? _This could be in a follow-up and a ping response can still contain version info like it does currently._

krancour (approved):
LGTM!

Issue #1146, opened 2020-10-23
set a time to live on user (and root) sessions


krancour:
There's already an expiration field that is set and honored by the auth filter, _but_ MongoDB supports TTL on documents, which means we can get free cleanup of expired sessions. We should take advantage of that.

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1306

PR #1306, opened 2021-03-01
feat(sessions_store.go): add index on 'expires' field for ttl cleanup

vdice:

* Adds an expiry index for each document in the sessions collection, to auto-cleanup per the 'expires' field

Closes https://github.com/brigadecore/brigade/issues/1146

**If applicable**:

### DIFF
--- a/v2/apiserver/internal/authn/mongodb/sessions_store.go
+++ b/v2/apiserver/internal/authn/mongodb/sessions_store.go
@@ -26,6 +26,9 @@ func NewSessionsStore(database *mongo.Database) (authn.SessionsStore, error) {
 		context.WithTimeout(context.Background(), createIndexTimeout)
 	defer cancel()
 	unique := true
+	// Set the default expiration value to 0 so that each record's 'expires'
+	// field is used to determine actual expiration
+	defaultExpiry := int32(0)
 	collection := database.Collection("sessions")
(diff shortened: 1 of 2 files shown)


krancour:
This looks good, but I do have a question.

I think, at present, if someone tries to use an expired token, the 403 they get back comes with a value in the reason field that explains their token is expired and they should log in again.

After this change, it's as if an expired token never existed in the first place-- which I think is fine.

My question is whether we should change the wording of the error message for the case where a token _isn't found_ to reflect that the token might simply have expired.

fwiw, we should still keep the logic in the filter that checks whether a token has expired or not. This way if something goes wrong on the mongo end (like the background thread that cleans up after TTLs have expired is disabled) we don't inadvertently end up with tokens that live indefinitely.

vdice:
> My question is whether we should change the wording of the error message for the case where a token isn't found to reflect that the token might simply have expired.

I'll take a look and see about updating this.

> An aside-- if any auth filter modifications are made, we should still keep the logic that checks whether a token has expired or not. This way if something goes wrong on the mongo end (like the background thread that cleans up after TTLs have elapsed is disabled) we don't inadvertently end up with tokens that live indefinitely.

Good point; I'll see about adding code comments around this scenario.

krancour (approved):
LGTM

PR #1307, opened 2021-03-01
Brigadier ESLint mandates Unix line endings

itowlson:
**What this PR does / why we need it**: When v2 is checked out on Windows, every line in Brigadier reports an error because the `.eslintrc` specifies Unix line endings, and the default for Git is to check out on Windows as CRLF and commit as LF. This PR ~adds a `.gitattributes` to Brigadier to tell Git to check out these files specially as LF~ removes the line break specifier.

**Special notes for your reviewer**: ~An alternative approach would be to remove the `linebreak-style` from `.eslintrc`. This is more convenient for developers, because even with the LF directives Windows tools will default to creating files as CRLF. But it could result in a misconfigured Windows install checking in CRLF files unless we add a separate CI lint that does check for this.~ After discussion with @krancour, we adopted this approach instead.

**NOTE:** This recreates #1305 which got closed through bizarre gitulatory shenanigans.

**If applicable**:


krancour (approved):
LGMT

PR #1309, opened 2021-03-03
don't allow client disconnects to interrupt writes

krancour:
#1304 got me thinking. There was some discussion related to that PR having to do with an infinite retry we currently have when writing new events to a queue and how that should probably just change to one of our usual exponential backoffs.

Well... that got me thinking more about how a request to do something like create an event, for instance, could run long (and still ultimately succeed) if it got into a backoff situation. And what happens if an impatient client hangs up in the meantime?

Just about everything we do throughout the whole system is currently interruptible by a canceled context... and request-bound contexts are canceled the moment the client disconnects.

Do we actually _want_ that behavior in certain cases? Suppose when creating a new event, the event is successfully written to the database in a pending status, but then the whole process is interrupted before the event's ID is written to the scheduler's queue-- this would leave us in an invalid state, with an event whose worker is in a pending phase with no path for transitioning out of that phase (short of an end-user  manually canceling or deleting the event).

It seems we'd be in better shape if requests utilizing HTTP methods that commonly mutate the underlying system (e.g. DELETE, PATCH, POST, and PUT) were considered _uninterruptible_. That can be achieved by substituting the background context for the default request-bound context.

This PR implements a new filter that does just that, and adds that filter as the first in our filter chain.

### DIFF
--- a/v2/apiserver/internal/lib/restmachinery/request_context_filter.go
+++ b/v2/apiserver/internal/lib/restmachinery/request_context_filter.go
@@ -6,11 +6,11 @@ import (
 )
 
 // requestContextFilter is a filter that can decorate an HTTP handler function
-// such that requests are examined and IFF the request utilizes an HTTP method
-// that commonly mutates the system (for instance, POST commonly is used for
-// creates and PUT and PATCH are commonly used for updates), then the request's
-// context is REPLACED with the background context. The usefulness of this lies
-// in creating a guarantee that operations that mutate the system are
+// such that requests are examined and if and only if the request utilizes an
+// HTTP method that commonly mutates the system (for instance, POST commonly is
+// used for creates and PUT and PATCH are commonly used for updates), then the
+// request's context is REPLACED with the background context. The usefulness of
+// this lies in creating a guarantee that operations that mutate the system are
 // uninterruptible by impatient clients that might hang up in the middle of a
 // request that is running long, perhaps due to retry logic with exponential
 // backoff. The client can hang up if they wish, but operations that mutate the
--- a/v2/apiserver/internal/lib/restmachinery/request_context_filter.go
+++ b/v2/apiserver/internal/lib/restmachinery/request_context_filter.go
@@ -1,0 +1,43 @@
+package restmachinery
+
+import (
+	"context"
+	"net/http"
+)
+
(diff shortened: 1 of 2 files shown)


v2/apiserver/internal/lib/restmachinery/request_context_filter.go
```diff
@@ -0,0 +1,43 @@
+package restmachinery
+
+import (
+	"context"
+	"net/http"
+)
+
+// requestContextFilter is a filter that can decorate an HTTP handler function
+// such that requests are examined and IFF the request utilizes an HTTP method
```
vdice, line 9:
```suggestion
// such that requests are examined and IF the request utilizes an HTTP method
```

vdice (commented):
This makes sense to me; I think of async or otherwise long-running REST api endpoints that may immediately return with 202 Accepted and then proceed with the task at hand.  Do we want to entertain doing the same here?  Or reserve that approach for truly async/long-running requests (not just _possibly_ long-running)?

v2/apiserver/internal/lib/restmachinery/request_context_filter.go
krancour: That was deliberate... IFF is a common shorthand for "if and only if," but perhaps it is not as well-known as I thought... in which case, I should probably change it.

krancour:
> I think of async or otherwise long-running REST api endpoints that may immediately return with 202 Accepted and then proceed with the task at hand. Do we want to entertain doing the same here? Or reserve that approach for truly async/long-running requests (not just possibly long-running)?

I thought about this, and I think I landed in the same place you did. 202 accepted seems appropriate when accepting the request and then running in the background is the _intended_ behavior, but this PR is mean to account for clients hanging up on interactions that were intended to be synchronous.

v2/apiserver/internal/lib/restmachinery/request_context_filter.go
krancour: Wording is revised in f6fdbf303312266f646f712203d1ef2de335f8ec.

vdice: Oh!  TIL.  hehe at least now I'm savvy :)

vdice (approved):

Issue #1255, opened 2021-01-29
SDK should not retry 500 errors


krancour:
Our SDK for go does automatic retries for certain types of error. This seems to _include_ 500 (internal server error).

I'd prefer that 500s didn't automatically get retried.

Retrying something that is quite probably a transient network fault (like a 502 or 503) makes all the sense in the world to me.

If the server is self-reporting that something went wrong on its end, I don't believe that should be retried automatically as the probability of continued failure is high and could even compound the issue.

__EDIT:__ Example of an issue being compounded:

Due to a coding error, I have observed a 500 that was thrown toward the very tail end of a project create operation. When this is automatically retried, the project already exists and a 409 (conflict) is returned stating that the project already exists. This sweeps the original issue under the rug.

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1314

PR #1314, opened 2021-03-11
feat(restmachinery/client.go): use custom retry policy to exclude 500 from retries

vdice:

* Defines our own retry policy for our base restmachinery client so that we have finer-grained control.  In this case, https://github.com/brigadecore/brigade/issues/1255 requested to not retry on 500s, so this is what differs from the default defined by the go-retryablehttp library.

Closes https://github.com/brigadecore/brigade/issues/1255

The output as tested via the cli looks like the following when a 500 is returned:

```
 $ b2 event create -p test-project


An internal server error occurred.
```

Is this acceptable?  Or do we want a follow up to try to add additional context?

**If applicable**:

### DIFF
--- a/sdk/v2/internal/restmachinery/client.go
+++ b/sdk/v2/internal/restmachinery/client.go
@@ -4,12 +4,15 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


krancour:
@vdice LGTM, but you got nailed by the linter.

krancour (approved):
LGTM

Issue #1174, opened 2020-11-24
improvement: create constants to represent resource type names


krancour:
There are numerous places in v2.0 code where the name of some resource type (e.g. Worker, Job, etc.) is referenced using a string literal. To promote consistency and avoid fallout from possible misspellings, we should create constants to represent these.

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1315

PR #1315, opened 2021-03-12
ref(*): add/use resource consts

vdice:

* Adds consts for commonly used resource names and updates applicable logic to use them

Closes https://github.com/brigadecore/brigade/issues/1174

There are still some resources not const-ified (search the repo for all instances of `Kind:` in creating meta.TypeMeta structs)... do we want to update these as well, even though there currently might only be one use of such strings?

**If applicable**:

### DIFF
--- a/sdk/v2/authn/users.go
+++ b/sdk/v2/authn/users.go
@@ -12,8 +12,8 @@ import (
 	"github.com/brigadecore/brigade/sdk/v2/restmachinery"
 )
 
-// UserLabel represents the canonical User label string
-const UserLabel = "User"
+// UserKind represents the canonical User kind string
+const UserKind = "User"
 
 // User represents a (human) Brigade user.
 type User struct {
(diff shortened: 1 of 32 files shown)


radu-matei (approved):

sdk/v2/authn/users.go
```diff
@@ -12,6 +12,9 @@ import (
 	"github.com/brigadecore/brigade/sdk/v2/restmachinery"
 )
 
+// UserLabel represents the canonical User label string
+const UserLabel = "User"
```
krancour, line 16: Based on how this is being used, would `UserKind` be more appropriate?

krancour: Same comment applies in several other places.

vdice: Sure, shall I update all of these to `*Kind`?

krancour: Yes... unless you disagree that my suggestion adds clarity to how those constants are used.

krancour (approved):

krancour:
LGTM. Integration tests failed... re-running.

PR #1318, opened 2021-03-18
a more elegant way of managing CODEOWNERS

krancour:
If we agree this is an improvement, I have PRs ready to to rip for most other repos in this org.


vdice (approved):
Nice!

PR #1319, opened 2021-03-18
a more elegant way of managing CODEOWNERS

krancour:
Like #1318 but for the master branch... necessary since it's still going to be some time before `v2` is merged into `master`.


krancour:
Converting to draft. Not sure yet if #1318 is working out how we'd have hoped it would.

See [this comment](https://github.com/brigadecore/brigade/pull/1321#issuecomment-802199518) for further context.

PR #1321, opened 2021-03-18
export Event from brigadier and worker's brigadier polyfill

### LINKED ISSUE #1317: Brigadier mistakenly does not export the Event type
This is making it a touch challenging to write `brigade.ts` scripts since `any` type has to be used anywhere the `Event` type should have been used.

krancour:
Fixes #1317

While touching these files anyway, I also organized exports alphabetically, which seems cleaner and more consistent.

Note that there are two new TODOs here that I do not intend to resolve at this time. It seems that approach we used for validating something is exported as expected, which works beautifully for constants, variables, and classes, simply does NOT work for interfaces. I'm not clear on why.

I have, however, manually verified that these changes address #1317.

### DIFF
--- a/v2/brigadier/src/index.ts
+++ b/v2/brigadier/src/index.ts
@@ -1,4 +1,4 @@
-export { events } from "./events"
-export { SerialGroup, ConcurrentGroup } from "./groups"
-export { Job, Container, JobHost } from "./jobs"
+export { Event, events } from "./events"
+export { ConcurrentGroup, SerialGroup } from "./groups"
+export { Container, Job, JobHost } from "./jobs"
 export { logger } from "./logger"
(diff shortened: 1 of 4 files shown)


krancour:
@brigadecore/maintainers peculiar... after #1318, this new PR didn't automatically request reviews from all the maintainers.

So either there's invalid syntax in `CODEOWNERS` now (according to docs tho, it looks good) _or_ GH simply doesn't request reviews from all individual @brigadecore/maintainers members. If the latter is the case, it's fine by me as long as the branch protection rule that requires approval from a code owner still honors the rights of everyone on @brigadecore/maintainers to approve.

So let's just use this PR as a test case for whether the changes in #1318 are acceptable.

vdice (approved):

PR #1316, opened 2021-03-16
v2 demo

krancour:
This PR is WIP and is intended to demonstrate Brigade 2 being used in conjunction with a prototype v2-compatible GitHub gateway to run Brigade 2 CI!

PR #1322, opened 2021-03-18
expose worker git details to brigadier-based scripts

krancour:
When working on #1316 I discovered that scripts had no way of getting to commit/ref info.

This PR fixes that.

The API server already includes this information when it creates the k8s secret for an Event. The only issue was that we weren't exposing it neatly via Brigadier.

### DIFF
--- a/v2/brigadier/src/workers.ts
+++ b/v2/brigadier/src/workers.ts
@@ -25,4 +25,33 @@ export interface Worker {
    * configuration.
    */
   logLevel?: string
+  /**
+   * If applicable, contains git-specific Worker details.
+   */
+  git?: GitConfig
+}
+
+/**
+ * Represents git-specific Worker details.
+ */
+export interface GitConfig {
+  /**
+   * Specifies the remote repository where, if applicable, the Worker will have
+   * obtained source code from.
+   */
+  cloneURL: string
+  /**
+   * Specifies the exact commit where, if applicable, the Worker will have
+   * obtained source code from.
+   */
+  commit?: string
+  /**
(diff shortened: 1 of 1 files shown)


v2/brigadier/src/workers.ts
```diff
@@ -25,4 +25,33 @@ export interface Worker {
...
+/**
+ * Represents git-specific Worker details.
+ */
+export interface GitConfig {
+  /**
+   * Specifies the remote repository where, if applicable, the Worker will have
+   * obtained source code from.
+   */
+  cloneURL: string
+	/**
```
vdice, line 43: nit: spacing inconsistency

v2/brigadier/src/workers.ts
```diff
@@ -25,4 +25,33 @@ export interface Worker {
...
+   * Specifies the exact commit where, if applicable, the Worker will have
+   * obtained source code from.
+   */
+	commit?: string
```
vdice, line 47: nit: spacing inconsistency

vdice (approved):

v2/brigadier/src/workers.ts
krancour: Gah. Thanks!!

v2/brigadier/src/workers.ts
krancour: Something must be wrong in my editor...

krancour: Tabs snuck in somehow even though .ts files are set to use space. 🤷 

Fixing it.

PR #1324, opened 2021-03-19
add "source state" to events

krancour:
Hacking away on a prototype v2-compatible GitHub gateway has exposed the fact that some gateways may need to preserve _state_.

Using the aforementioned gateway as an example, after adding events of certain types to Brigade's event bus, the gateway monitors the status of those events to report that status "upstream" to GitHub. If the process dies and resumes, it needs to somehow "remember" which events it had been monitoring. This sort of requirement is almost certainly not unique to this one gateway.

If we expect gateways to manage _simple_ state like this entirely by themselves, we significantly raise the bar to implementing and operating a gateway. Will we expect gateways to have their own databases? Will we expect operators to be ok with how sprawling their Brigade installation becomes when gateways come with these sort of dependencies?

A viable alternative is to provide:

1. A mechanism for gateways to store gateway-specific state _in_ events that they have created.
2. A mechanism for gateways to query for events in a particular gateway-specific state.

This PR implements the above by adding a `SourceState` field to the `Event` type and adding additional criteria to the `EventSelector` type.

Using the aforementioned GitHub gateway to illustrate usage:

1. Gateway creates Event with "watching = true" in the source state.
2. Once the Event enters a terminal state _and_ has had its final status reported upstream to GitHub, update the the source state to set "watching = false" or _remove_ "watching" altogether.
3. If the gateway component that watches Events dies and restarts, it queries the GitHub API upon startup to find all Events with "watching = true" in the source state. It picks up where it left off.


v2/apiserver/internal/core/rest/events_endpoints.go
```diff
@@ -213,3 +235,41 @@ func (e *EventsEndpoints) deleteMany(w http.ResponseWriter, r *http.Request) {
 		},
 	)
 }
+
+func eventsSelectorFromURLQuery(
```
vdice, line 239: Should we add unit test coverage for this helper?  (Or covered implicitly by others?)

v2/apiserver/internal/core/mongodb/events_store.go
```diff
@@ -71,16 +71,27 @@ func (e *eventsStore) List(
...
-			"$in": selector.WorkerPhases,
-		},
 		"deleted": bson.M{
 			"$exists": false, // Don't grab logically deleted events
 		},
 	}
 	if selector.ProjectID != "" {
 		criteria["projectID"] = selector.ProjectID
 	}
+	if selector.Source != "" {
```
vdice, line 81: Should we consider consolidating the shared bits of criteria-building based on the provided selector, in a similar vein to `eventsSelectorToQueryParams`, etc., added in this PR?

v2/apiserver/internal/core/rest/events_endpoints.go
krancour: Good question. Not covered implicitly by other tests because the endpoints are the current weak point in our coverage.

I will add a test for this!

v2/apiserver/internal/core/mongodb/events_store.go
krancour: We could, but I had opted against it. Reasons being:

1. For `CancelMany` and `DeleteMany`, `ProjectID` is required (and the service layer enforces that), but for `List`, it's optional, so it comes with an extra `if selector.ProjectID != ""...` guard that doesn't occur in the other two places.

2. For `CancelMany` the criteria based on worker phase are different. We build and execute two separate statements for things that need to go from phase `PENDING` to `CANCELED` and things that need to go from `STARTING` or `RUNNING` to `ABORTED`.

In light of these differences, we could use a common function to handle extracting _some_ query/statement criteria from the selector, but for other bits of criteria, we'd need varying, more specific logic. It felt to me like it would be more confusing to spread the criteria-building logic around than to just keep it how it is here, even if that means being a little bit duplicative.

All that said, I am, as ever, willing to entertain any counter-argument to that.

vdice: Appreciate your pointing out the finer-grained differences between these functions.  Let's leave as is for now!

v2/apiserver/internal/core/rest/events_endpoints.go
vdice: Tests look great; thank you!

vdice (approved):
LGTM

PR #1326, opened 2021-03-23
refactor brigadier, brigadier polyfill, and worker for max efficieny and flexibility

krancour:
@itowlson recently uncovered that support for third-party dependencies expressed in `package.json` aren't accessible at runtime.

Some history is in order.

In 1.x, we didn't respect `package.json`. Third-party dependencies could be specified using `brigade.json` instead. The worker's prestart process exec'ed out to `npm` to install each prescribed dependency individually _in the same directory as the worker's own script_. So, in 1.x, when the worker `require`d the user's `brigade.js` script, its third-party dependencies were accessible.

Because respecting the conventional approach to JS dependency management seems an obvious benefit to users, v2 has long-since abandoned support for `brigade.json` and adopted support for `package.json` instead. Since users can now provide a detailed and _familiar_ manifest of their script's dependencies, we wanted to tamper with those expectations as little as possible. Not only was the prestart process adapted to run `npm install` (or `yarn install`, depending), it consciously does so _in the directory containing the user's `brigade.js` script_ so that its dependencies are managed and resolved _separately_ from the worker's own.

The fallout from the above change is that when the worker `require`s the user's `brigade.js` script, the worker's own dependencies are available, but the `brigade.js` script's dependencies are not-- it is, after all _one_ OS process that is seeded only with the worker's dependencies. fwiw, I was fooled into believing things were working because I tested using a `brigade.js` with a third-party dependency that coincidentally happened to be among the worker's own.

Fixing this requires that the worker's own script(s) and the user's `brigade.js` script run _in separate OS processes_, each having been seeded with the correct dependencies.

The approach implemented by this PR is the following:

As before, the worker's `prestart.js` script is responsible for all that must happen _in advance_ of executing the user's `brigade.js` script, although the particulars have changed. It:

1. Executes dependency resolution, if applicable, using `npm install` or `yarn install`.

1. _Polyfills_ the `@brigadecore/brigadier` dependency with an alternative Brigadier implementation that isn't full of "no-ops" and is actually capable of communicating with the Brigade API server. Also, it does this using a symbolic link to pre-built polyfill and _does not_ interfere at all with the remainder of the script's dependencires. It also doesn't introduce any pulls from npmjs.com just to resolve dependencies of something we've actually pre-built.

1. `exec`'s to spawn _a new OS process_ to execute the user's `brigade.js` script with its _own_ dependencies (and the Brigadier polyfill).
    
The most significant change from the previous implementation is that the worker's _other_ process (`index.js`) has gone away in its entirety, with its narrow set of responsibilities (loading and dispatching the event) having been absorbed into the Brigadier polyfill.

_In essence, the worker now _only_ handles script-prestart and nothing else._

There have been a few structural changes to the project that accompany this. Mainly the Brigadier polyfill is its own separate module from the worker now because I was finding it much easier to rationalize the relationships between the worker, the real Brigadier, and the Brigadier polyfill when they were each their own module. Not too mention, it made it a bit easier to build them all as well.

__One thing that is critical to note is that after this change, `brigade.js` scripts _must_ include an `events.process()` line toward the end. This is the line that (via the polyfill) will load and dispatch the event and it's necessary because the `brigade.js` script is now, for the first time, the actual _entrypoint_ into executing user-defined behavior and it isn't `require`d by some other process. i.e. It's a reversal of the previous "don't call us, we'll call you," approach.__

While this is different from what we've been accustomed to with Brigade 1.x (and v2 up until now), imho, it both makes sense _and_ makes the entire interplay between worker and user-defined script much, much easier to rationalize. Overall, this is a _drastic_ reduction in worker complexity and I view that as a positive.

@vdice and @itowlson if either of you are amenable, I'd like to actually schedule some time to discuss this change as a group. 1. Just because it's a large-ish one and 2. because either of you might have insights into making this even smoother.


krancour:
fwiw, here is my (personal, for now) WIP suite of examples that are all compatible with these changes.

https://github.com/krancour/brigade2-examples

itowlson:
I feel like the requirement for `events.process()` is likely to be problematic.  This is the sort of thing that is super easy to overlook, and possibly not easy to diagnose.

If we can't remove this requirement, should we perhaps have a scaffolding command e.g. `brig script create` that creates a basic `.brigade/brigade.js` file (and `package.json`) with the necessary incantation?  That might at least reduce how many people forget it...

krancour:
> I feel like the requirement for events.process() is likely to be problematic. This is the sort of thing that is super easy to overlook, and possibly not easy to diagnose.

@itowlson, good thought. I just had a chat with @radu-matei and @vdice. Here's a quick summary, which also touches on that issue.

* Everyone was clear on why the support for `package.json` isn't working as intended and we all generally agreed that the solution was to use two, separate OS processes-- one to "prestart" (prepare) the environment, and a second to execute the user-defined script. Keeping these separate means they can each run with their own distinct set of module dependencies. To satisfy this, a `brigade.js` needs to be standalone / executable in a way it never has been before. i.e. It can't merely register event handlers, it has to register handler _and_ dispatch the event.

* We discussed whether the added requirement of calling `events.process()` in `brigade.js` would trip users up. The conclusion we came to was that it may indeed, but v2 Brigadier already includes a number of breaking changes that make a v1.x `brigade.js` incompatible with Brigade v2, and because most of those changes are less conspicuous, the requirement of including `events.process()` is, in some respects, beneficial as a canary in the coal mine. :wink:

* We agreed there was a large and tangible _benefit_ to the addition of the `events.process()`-- namely, for the first time ever you can now type `node brigade.js` and expect the script to actually _do something_. It won't do anything more than tell you what Brigade _would_ have executed and in what order, but it wouldn't even do that much in v1.x. We agreed this benefit outweighed any inconvenience. We also toyed with the idea that this could lend itself to other implementations of the Brigadier library int he future that do _more_ than this (whilst still stopping short of full-blown Brigade integration). For instance, imagine being able to type `node brigade.js` and have all the jobs defined therein execute as containers managed by the local `dockerd` or `containerd`. (But to be clear, this is still firmly in the realm of the hypothetical.)

* We discussed whether we could abandon all this "polyfill" nonsense and have a Brigadier library that, like today, is just interfaces and trivial or "stubbed-out" implementations, with various implementations of those interfaces provided by other libraries. We got stuck on how we'd go about selecting / loading those at runtime _without_ coming back to the polyfill-based approach, so in a way we go stuck in a cyclic loop of wanting to do things a little more simply, but also being dependent on the very technique we wished to get away from. We decided to put a pin in that and explore it _separately_ form this issue.

* We agreed this PR is a clean fix for the original issue you reported (`package.json` support not working as advertised) and we think we want to merge this and then _continue_ exploring better options.

Now to address this:

> should we perhaps have a scaffolding command e.g. brig script create that creates a basic .brigade/brigade.js file (and package.json) with the necessary incantation?

I'd already been thinking we should do something along those lines to generate _project definition YAML_ that you can feed to `brig project create -f <file>`. I think extending that concept to cover `brigade.js` + `package.json` generation as you have suggested would be even _more_ useful. So, I can now imagine something along the lines of `brig init` (with a few options, etc.) that would set _everything_ up for you-- lots and lots of possibilities here. 👍

krancour:
Also, @radu has suggested that while changes are being made to the prestart script, that npm should be the default package manager and yarn should be used only if we find a `yarn.lock`. This is the opposite of what it's currently doing. (It currently used yarn unless it finds `package-lock.json`.)

His rationale is that in the absence of a hint that the script's author was already using yarn, we should cater to the lowest common denominator. I'm cool with this and will amend the PR accordingly.

radu-matei:
I am firmly in favor of a `brig init` command that would create everything for the user.

krancour:
@radu-matei I'll write up an issue for that shortly unless someone beats me to it.

radu-matei (approved):

PR #1332, opened 2021-03-25
new and improved examples

krancour:
New examples.

These cease the annoying practice (my own fault) of having _two_ version of each exampe-- one in JS and one in TS. That doesn't actually help anyone.

These examples are numbered in a logical order. If you were demoing this, that logical order is order of increasing complexity.

Longer term, I'd like to add a `README.md` to every example as well.


vdice:
Darn, the integration tests actually succeeded but it looks like that happened shortly after the default kind job timeout was hit (`time limit (1800000 ms) exceeded` aka 30 mins).  If we start seeing this often, we can bump up the timeout.  I re-requested this check.

PR #1334, opened 2021-03-26
add --non-interactive and --continue flags to CLI

### LINKED ISSUE #1328: brig event list should provide limit and non-interactive options
Currently `brig event list` displays the first 10 (I think) events and then prompts with `<n> results remain. Fetch more?`  This causes a couple of problems:

1. If I only want to see the most recent event(s), I have to hit `No` every time I run the query
2. I can't easily write scripts to process events

(The interactive paging, by the way, is particularly odd with `-o json` where it prints a 10-item `EventList` object, then prompts and if you choose Y prints another `EventList`.  I assume `-o json` is really intended to work only with redirects.)

The interactive prompt doesn't display if the output is redirected, so you can work around the prompt by writing `brig event list | cat`.  Or you could write `brig event list | head -n 3` to print the first event (3 because 1 blank line and 1 header row).  But it would be good to have a built-in option that could limit the number of results and/or suppress the prompt.

### LINKED ISSUE #1329: Do we want a `--follow` option for `brig event create`
v2 has a much stronger emphasis than v1.x on events being handled _asynchronously_, subject to scheduling constraints.

Because of this `brig run` which sounds like an imperative to "do it now," was replaced with `brig event create`.

While I agree with that change, I regardless find myself _frequently_ doing this:

```console
$ brig event create -p pipeline-demo
$ brig event logs -f -i <event uuid returned by previous command>
```

I think we _should_ entertain the idea of supporting a `--follow` (alias `-f`) flag on the `brig event create` command. It can either stream event status from the API or query event status periodically (with a progressive backoff?) until the event has advanced past a pending state. At that point, it can begin streaming even logs from the API.

I think such a flag would be a nice convenience and, provided it is documented properly, won't imply to the user that events are handled synchronously.

Anyone have thoughts about this?

krancour:
Attn: @itowlson 

Fixes #1328

* Adds a (global) non-interactive flag that can be used to explicitly disable interactive prompts. The prompts used to be enabled whenever there was a tty and disabled otherwise, but as it turns out, there are cases where there is a tty, but the CLI is still being used as a programmatic interface.

* Improved consolidation of all code for interactive prompts, especially related to pagination.

* Re: pagination, if additional results exist for a list operation, these are the rules for how we deal with that:

    * If output format is JSON or YAML do nothing. The response already contains information about how many results remain and how to access the next page.

    * If the output format is a table:

        * If `--non-interactive` was specified or there is no tty, a message is displayed to show the user how to access the next page of results. This is not super user friendly, but realistically, human users are going to do this interactively anyway.

        * If `--non-interactive` was not specified and there is a tty, permit users to page interactively as before.


krancour:
> How is/should --continue <value> be used?

If you're running non-interactively, the CLI should you the continue value for accessing the next page of results.

> I don't see a --non-interactive flag on brig event list

It's a global flag (like `-k`). This seems to bring with it a few peculiarities. To see help for the flag, try `brig -h` with no sub-command specified. Also, the `--non-interactive` / `-n` flag (like `-k`) must _precede_ any sub-commands.

More research is in order to determine if there is some way around this. Last resorts would be switching CLI frameworks (although, in general, I happen to like this one) or _not_ defining this as a global flag and opting instead to add it individually to every command that potentially has an interactive aspect to it.

vdice:
> If you're running non-interactively, the CLI should show you the continue value for accessing the next page of results.

Oh yea, I see now (having used the global `--non-interactive` flag correctly).  Thanks.

> To see help for the flag, try brig -h with no sub-command specified. Also, the --non-interactive / -n flag (like -k) must precede any sub-commands.

Ah, right.  Indeed, it'd be handy to have the global flags displayed on any sub-command's `-h` menu... then I think we'd be set.  Maybe look into this as a follow-up?  (Impressed w/ the current cli library as well; I wouldn't consider this a dealbreaker)

Just checked Porter (uses cobra)... something like this would be great:
```
 $ porter build -h
Builds the bundle in the current directory by generating a Dockerfile and a CNAB bundle.json, and then building the invocation image.

Usage:
  porter build [flags]

Examples:
  porter build
  porter build --name newbuns
  porter build --version 0.1.0
  porter build --file path/to/porter.yaml
  porter build --dir path/to/build/context


Flags:
  -d, --dir string         Path to the build context directory where all bundle assets are located.
  -f, --file porter.yaml   Path to the Porter manifest. Defaults to porter.yaml in the current directory.
  -h, --help               help for build
      --name string        Override the bundle name
      --no-lint            Do not run the linter
  -v, --verbose            Enable verbose logging
      --version string     Override the bundle version

Global Flags:
      --debug           Enable debug logging
      --debug-plugins   Enable plugin debug logging
```

krancour:
In general, I've been finding this easier than cobra. Longer term, still quite willing to switch if it comes to that.

Here's a possible workaround for dealing with the positioning constraints for global flags.

https://github.com/urfave/cli/issues/427#issuecomment-712292636

I'd love to look at that in a follow-up.

vdice (approved):

vdice:
Personally, I'm less concerned with adding the ability to declare global flags at any point in the command string and more concerned with visibility of relevant global flags to the user.  In other words, if we find some quick way to config the cli to auto-display global flags on the help menus for any subcommand, I think that'd be a big win.

krancour:
@vdice Thanks for the review. I opened #1336 so we don't lose track of anything.

Issue #1323, opened 2021-03-19
labels in event subscriptions don't work intuitively


krancour:
Currently, for a project's event subscription to match a given event, the subscription must use _all of the exact labels_ that the event does. This was a deliberate choice to prevent _accidental over-subscription_ -- an example of which would be someone who subscribed to "push" events from "github" _without_ specifying the "repo = whatever" label, but didn't _really_ want a deluge of events from every repo. Subscriptions requiring _all of the exact labels_ of an event in order to match it neatly circumvents this.

But for better or worse, the term "labels" probably comes pre-loaded with behavioral expectations of a sort. In Kubernetes, for instance, if you ask for all pods labeled "foo = bar", you get _all_ the pods that have that label, even if those pods also have other labels _that you didn't specify_. The fact that our labels don't work _just like this_ is almost guaranteed to cause confusion.

To put any proposed solution to the test, @itowlson has furnished a hypothetical use case wherein two Brigade projects might be interested in the same type of events originating from the same GitHub repository, but one is interested in those events only if they originated from a specific branch, and the other is interested in all such events, regardless of branch.

Here is one proposed method of handling this:

1. The existing "labels" field on both the event and event subscription types can be renamed to "attributes" (or something else if better suggestions are forthcoming). Name aside, everything would continue to work as it currently does. For a project's event subscription to match a given event, the subscription must use _all of the exact attributes_ that the event does. This would continue to prevent accidental over-subscriptions, but we would also avoid misalignment with the reasonable behavioral expectations users might associate with "labels."

1. We can introduce a _new_ "labels" field on both the event type and on the event subscription type. This will behave differently from how "attributes" (nee "labels") work, but consistently with the reasonable behavioral expectations users might already associate with "labels." i.e. Specifying a label in an event subscription _narrows_ the pool of matches, but _not_ including a label cannot implicitly preclude a match.

A "push" event from the GitHub gateway might look like this:

```
source: brigade.io/github-gateway
type: push
attributes:
  repo: krancour/some-repo
labels:
  branch: master
```

Someone who doesn't read the docs and accidentally subscribes like this won't get a deluge of events from all repos (they'll get _no_ events):

```
source: brigade.io/github-gateway
type: push
```

A project with this subscriptions will get all push events from the specified repo, regardless of branch:

```
source: brigade.io/github-gateway
type: push
attributes:
  repo: krancour/some-repo
```

A project with this subscriptions will get all push events from the specified repo, but only for the specified branch:

```
[...]

PR #1338, opened 2021-03-31
change how labels work; introduce qualifiers

krancour:
Potential fix for #1323 

cc @itowlson

If possible, I want to get this into v2-alpha.2, even if it means delaying the release a few days.

Note: Name "qualifiers" subject to change if anyone has a better suggestion.


v2/apiserver/internal/core/mongodb/projects_store.go
```diff
@@ -120,15 +121,81 @@ func (p *projectsStore) ListSubscribers(
 	event core.Event,
 ) (core.ProjectList, error) {
 	projects := core.ProjectList{}
-	subscriptionMatchCriteria := bson.M{
+	// Finding all Projects that are subscribed to a given Event is rather tricky.
```
vdice, line 124: Wow, accruing advanced mongo knowledge.  Hadn't seen this method.

v2/apiserver/internal/core/mongodb/projects_store.go
```diff
@@ -120,15 +121,81 @@ func (p *projectsStore) ListSubscribers(
...
+function() {
+	const project = this;
+	if (!project.spec.eventSubscriptions) {
+		return false;
+	}
+	const event = %s;
+	event.qualifiers = event.qualifiers || {};
+	event.labels = event.labels || {};
+	loop:
+	for (const subscription of project.spec.eventSubscriptions) {
```
vdice, line 173: I wonder if there's a way to cover under test?  (Could be in a follow-up).  Maybe we try to test the JS directly (not via mongo or e2e) with pre-loaded data, to verify its logic is correct/intended?

v2/apiserver/internal/core/events.go
```diff
@@ -44,15 +45,37 @@ type Event struct {
...
+	// subscribing to ALL Events or ONLY events originating from a specific
+	// channel, then that gateway can LABEL Events it emits into Brigade's event
+	// bus with channel=<channel name>. Projects wishing to subscribe to ALL
+	// Events from the Slack gateway MAY omit that Label from their
+	// EventSubscription, while Projects wishing to subscribe to only Events
+	// originating from a specific channel MAY include that Label in their
+	// EventSubscription. Note that the Labels field's "MAY match" subscription
+	// semantics differ from the Qualifiers field's "MUST match" subscription
+	// semantics.
+	Labels map[string]string `json:"labels,omitempty" bson:"labels,omitempty"`
```
vdice, line 78: I notice how the `Qualifiers` field here has its type as `Qualifiers` (to utilize its `MarshalBSONValue` method, right?)  Curious as to why we don't have a need for the same for Labels -- it looks like it _used_ to have that same method, etc.?

v2/apiserver/internal/core/mongodb/projects_store.go
krancour: Good question. That question actually applies to all our queries (i.e. not just this one on account of the JS). Our unit tests for store code all simulate responses from the database, but they don't account at all for the possibility that a query is simply _wrong_.

What I think I'm banking on here is that once Brigade 2 reaches a point where we're using it to test itself, the new support for jobs with sidecars means that we can stand up a live MongoDB instance alongside our Go test container and actually start to put the store queries to the test.

v2/apiserver/internal/core/events.go
krancour: Good observation. It has to do with how each of the fields is queried.

Technically, the new `Qualifiers` field _is_ the old `Labels` field, simply renamed. The _new_ `Labels` field is the _real_ new thing.

When we match on qualifiers, we are looking for an _exact_ match. (This is what we _used_ to be doing with the labels, but @itowlson pointed out, quite correctly, that such behavior really didn't comport with reasonable assumptions people are likely to have about how "labels" work.) In MongoDB, an _exact_ match on a subdocument requires you to be using `bson.D` instead of `bson.M` and that's basically the reason for the `MarshalBSONValue` function on the `Qualifiers` (nee `Labels`) type. (The difference between `bson.D` and `bson.M` is that the order of the keys is canonicalized in a `bson.D`. `bosn.M` doesn't have that feature. Think of it like canonical JSON vs "regular" JSON.)

The match semantics on the _new_ `Labels` field don't involve making an exact match on a subdocment, so this concern of having to use `bson.D` doesn't present itself.

v2/apiserver/internal/core/mongodb/projects_store.go
krancour: I just learned this trick in the last few days. I'd rather not have done this, but it's the only way I could find to do the "reverse" label match.

v2/apiserver/internal/core/events.go
vdice: Ah, thank you.  That clarifies things!

vdice (approved):

vdice:
Just had a thought for a follow-up:  It may be useful to add an example showing a project with an eventSubscription config utilizing the new qualifiers and labels fields (and maybe others?)

krancour:
And fwiw, I still want @itowlson to weigh in on this before we merge.

itowlson (approved):
This is super useful - thank you.  I agree the naming might need to evolve but I don't have any ideas I'm afraid; not qualified to review the code without a longish ramp-up; so LGTM!

PR #1341, opened 2021-04-01
add script and make target for standing up a customized kind cluster

krancour:
As the number of contributors to v2 starts to grow, I figure it might be useful to start sharing some of the little hacks that I've kept to myself until now.

This PR adds a make target for standing up a new customized kind cluster that:

* Is integrated with a local Docker registry
* Forwards localhost:31600 to the API server
* Has NFS support pre-installed to support Brigade's "shared workspaces" feature

End-user documentation will obviously have to address what the prerequisites within a cluster are, in general, for running Brigade 2. This contribution is not aimed at end-users. It's aimed at contributors who want to rapidly set up a suitable, local environment to facilitate development.

A comprehensive guide to hacking on v2 will follow in another PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -301,6 +301,10 @@ publish-cli: build-cli
 # Targets to facilitate hacking on Brigade.                                    #
 ################################################################################
 
+.PHONY: hack-new-kind-cluster
+hack-new-kind-cluster:
+	hack/kind/new-cluster.sh
+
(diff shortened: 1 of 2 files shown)


vdice (approved):
Handy script!

krancour:
No lt sure why CI didn't run...

Issue #1247, opened 2021-01-25
JSON Schema Validation


vdice:
We should add validation for the Brigade JSON [schemas](https://github.com/brigadecore/brigade/tree/v2/v2/apiserver/schemas).  As a first go, just using something like the [ajv cli](https://github.com/ajv-validator/ajv-cli) to run validation against samples (e.g. the project samples in the `examples` dir) will be beneficial in vetting that the samples are constructed correctly and that the schemas themselves are valid.

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1343

PR #1343, opened 2021-04-01
feat(ci): add schema and example validation

vdice:

* Adds CI for validating schema compilation and examples against these schemas

I couldn't figure out how to keep the namespaced `$id`'s (e.g. `github.com/brigadecore/brigade/v2/job.schema.json`) for each schema and have that work with both ajv (it did not like the `file://` approach we had) and runtime validation by the api server.  So, I simplified the `$id` fields, which also simplifies the `$ref`'s for accessing common definitions.  Thoughts/concerns?

Closes https://github.com/brigadecore/brigade/issues/1247

* Not sure if this helps or hurts the VS Code syntax highlighting; I recall there may have been an issue in this realm cc @krancour 

**If applicable**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,6 +185,37 @@ test-integration: hack-expose-apiserver
 			./tests/... || (cd - && $(MAKE) hack-unexpose-apiserver && exit 1)
 	@$(MAKE) hack-unexpose-apiserver
 
+# Validates the schemas in the v2/apiserver/schemas dir
+#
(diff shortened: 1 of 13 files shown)


krancour (approved):
LGTM. I'm optimistic this _may_ even fix the text editor / IDE context help issue, but in the very least, I cannot see it making it worse.

If it doesn't fix it, we'll revisit that later.

But I _love_ what this PR accomplishes.

Issue #1276, opened 2021-02-12
disallow transition out of a terminal phase


krancour:
For both workers and jobs, it may sometimes be possible that post-cancellation of a running worker or job (abortion) the observer will see the corresponding pod come down with a non-zero exit code and attempt to transition the worker or job to a failed phase-- when in fact the pod came down only because it was deliberately terminated _by_ the cancel/abort process...

As a general rule, we should simply disallow any phase changes to a worker or job that has already entered a terminal phase.

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1344

PR #1344, opened 2021-04-02
fix(worker/job services): disallow transition of workers/jobs from terminal status

vdice:

* Disallows updating a worker's or job's status if their phase is already terminal

Closes https://github.com/brigadecore/brigade/issues/1276

**If applicable**:

### DIFF
--- a/v2/apiserver/internal/core/jobs.go
+++ b/v2/apiserver/internal/core/jobs.go
@@ -552,24 +552,30 @@ func (j *jobsService) UpdateStatus(
 		}
 	}
 
-	// Only update status if job's current phase is non-terminal
-	// TODO: do we want to return an error if the phase *is* terminal?
-	if !job.Status.Phase.IsTerminal() {
-		if err := j.jobsStore.UpdateStatus(
+	// We have a conflict if the job's phase is already terminal
+	if job.Status.Phase.IsTerminal() {
+		return &meta.ErrConflict{
+			Type: JobKind,
+			ID:   job.Name,
+			Reason: fmt.Sprintf(
+				"Event %q job %q has already reached a terminal phase.",
+				eventID,
+				job.Name,
+			),
+		}
+	}
+
+	return errors.Wrapf(
+		j.jobsStore.UpdateStatus(
 			ctx,
 			eventID,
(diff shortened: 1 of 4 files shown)


krancour:
@vdice I'm wondering if this change might better off in the update worker/job endpoints of the API.

If we hypothetically supported a different substrate someday, the observer component would be a different implementation. It would keep the bar for that kind of work lower if enforcement of these kind of rules is centralized in the API server.

vdice:
@krancour ah, good call.  Silly that I didn't think about that earlier (as it's a paradigm you've encouraged in the past, for good reasons).  Will do so!

vdice:
Ok @krancour PTAL when convenient.

v2/apiserver/internal/core/jobs.go
```diff
@@ -539,18 +539,35 @@ func (j *jobsService) UpdateStatus(
...
+	job, ok := event.Worker.Job(jobName)
+	if !ok {
+		return &meta.ErrNotFound{
+			Type: JobKind,
+			ID:   jobName,
+		}
+	}
+
+	// Only update status if job's current phase is non-terminal
+	// TODO: do we want to return an error if the phase *is* terminal?
```
krancour, line 556: Yes. It's a conflict.

v2/apiserver/internal/core/workers.go
```diff
@@ -384,16 +384,26 @@ func (w *workersService) UpdateStatus(
...
-			err,
-			"error updating status of event %q worker in store",
+	// Check current phase of event worker
+	event, err := w.eventsStore.Get(ctx, eventID)
+	if err != nil {
+		return errors.Wrapf(err, "error retrieving event %q from store", eventID)
+	}
+
+	// Only update status if worker's current phase is non-terminal
+	// TODO: do we want to return an error if the phase *is* terminal?
```
krancour, line 394: Yep. `meta.ErrConflict`

krancour:
@vdice LGTM after meta.ErrConflict is returned in the event of invalid state change attempts.

krancour (approved):
LGTM!

Issue #1257, opened 2021-01-29
Add capability to list/search principal <--> role assignments


krancour:

krancour:
Note that I have this in-progress already.

krancour:
I'm punting this to beta.1 because not having this completed isn't blocking anything else. I want to focus my own alpha.2 efforts on things that will unblock other streams-- like gateway development.

krancour:
This was closed by #1368

PR #1353, opened 2021-04-08
Refactor authz

krancour:
This is related to #1257, which I've been working on, on-and-off, for some time.

https://github.com/brigadecore/brigade/pull/1241 introduced endpoints, services, and stores having to do with role assignments (the association between a principal and a role).

In a (perhaps overzealous) effort to be DRY, a lot of the code that was committed was common to two different _types_ of roles-- these being system-level roles like `ADMIN` and project-level roles, like `PROJECT_ADMIN`, which are, necessarily, scoped to a given project.

As my work on #1257 continue, I'm finding it increasingly difficult to account for behavioral differences between these two types of roles while keeping things DRY. I've finally concluded that #1257 is exposing the fact that the differences between the two types of roles outweigh their rather superficial similarities by a wide margin.

I'm invoking "clear is better than clever" and unDRYing some things in this PR.

There are no functional changes in this PR, but it leaves me with two _distinct_ sets of components for dealing with the two types of roles so that I can mutate each of those independently from the other as work on #1257 continues.

I'm leaving this as a draft for right now because I want to start rebuilding my solution for #1257 on top of this to prove to myself that it really does improve the state of affairs.

### DIFF
--- a/sdk/v2/authn/principals.go
+++ b/sdk/v2/authn/principals.go
@@ -1,0 +1,26 @@
+package authn
+
+// PrincipalType is a type whose values can be used to disambiguate one type of
+// principal from another. For instance, when assigning a Role to a principal
+// via a RoleAssignment, a PrincipalType field is used to indicate whether the
+// value of the PrincipalID field reflects a User ID or a ServiceAccount ID.
+type PrincipalType string
+
+const (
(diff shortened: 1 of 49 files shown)


krancour:
I'm going to break this up into several smaller PRs.

Issue #1352, opened 2021-04-07
need an unversioned ping endpoint


krancour:
Our versioned "v2" pings return server version information. But you'd need to know you're talking to a v2 server in order to send a ping or get (more precise) server version information.

We should add an additional, unversioned ping endpoint where clients that are clueless about what API version they're talking to can find out.

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1354

PR #1354, opened 2021-04-09
feat(*): add unversioned ping endpoint

vdice:

* adds an unversioned ping endpoint to return a raw api server version string
* re-orgs the integration tests to set up the api clients first/once (and issue a ping check) before running all of the scenarios

Closes https://github.com/brigadecore/brigade/issues/1352

**If applicable**:

### DIFF
--- a/sdk/v2/system/api_client.go
+++ b/sdk/v2/system/api_client.go
@@ -64,13 +64,14 @@ func (a *apiClient) UnversionedPing(ctx context.Context) ([]byte, error) {
 			SuccessCode: http.StatusOK,
 		},
 	)
+	defer resp.Body.Close()
 	if err != nil {
-		return []byte{}, errors.Wrap(err, "error submitting request")
+		return nil, errors.Wrap(err, "error submitting request")
 	}
 
 	respBodyBytes, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return []byte{}, errors.Wrap(err, "error reading response body")
+		return nil, errors.Wrap(err, "error reading response body")
 	}
 	return respBodyBytes, nil
 }
--- a/sdk/v2/system/api_client.go
+++ b/sdk/v2/system/api_client.go
@@ -2,10 +2,12 @@ package system
 
 import (
 	"context"
+	"io/ioutil"
 	"net/http"
 
 	rm "github.com/brigadecore/brigade/sdk/v2/internal/restmachinery"
 	"github.com/brigadecore/brigade/sdk/v2/restmachinery"
+	"github.com/pkg/errors"
 )
 
 // PingResponse represents the expected response object returned by the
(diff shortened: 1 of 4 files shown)


sdk/v2/system/api_client.go
```diff
@@ -16,8 +16,11 @@ type PingResponse struct {
 
 // APIClient is the client for system checks involving the Brigade API.
 type APIClient interface {
-	// Ping pings the API Server
+	// Ping sends a GET request to the API Server's versioned ping endpoint
 	Ping(ctx context.Context) (PingResponse, error)
+
+	// PingRaw sends a GET request to the API Server's unversioned ping endpoint
+	PingRaw(ctx context.Context) (string, error)
```
krancour, line 23: Nit: I'm not sure how I feel about "PingRaw" because (unless one as has read the code), the name doesn't unambiguously denote its function. It returns a _version_ string, but nothing about the name hints that that is what you'll get when you invoke this.

imho, if we named this `UnversionedPing` its function would be more obvious.

What would you think about still returning JSON instead of just a string? That leaves the possibility of including additional information in the response someday if we deem it necessary. Already, I can imagine that there are _two_ useful bit of info to returns-- server version and server _API_ version.

Then there's the "raw" aspect of it. If we don't assume a particular API version, we don't know what sort of type to unmarshal the response into. Maybe we can just return a slice of bytes?

Something like this?

```
UnversionedPing(ctx context.Context) ([]byte, error)
```

vdice: > If we don't assume a particular API version, we don't know what sort of type to unmarshal the response into.

Yea, that's what I was thinking, hence returning a string (and not structured JSON).  But I'm down to switch to `[]byte` for sure.

`UnversionedPing` sgtm!

sdk/v2/system/api_client.go
```diff
@@ -48,3 +54,23 @@ func (a *apiClient) Ping(ctx context.Context) (PingResponse, error) {
...
+	resp, err := a.SubmitRequest(
+		ctx,
+		rm.OutboundRequest{
+			Method:      http.MethodGet,
+			Path:        "ping",
+			SuccessCode: http.StatusOK,
+		},
+	)
+	if err != nil {
+		return []byte{}, errors.Wrap(err, "error submitting request")
```
krancour, line 68: Non-blocking nit: you can return nil here instead of allocating a slice.

sdk/v2/system/api_client.go
```diff
@@ -48,3 +54,23 @@ func (a *apiClient) Ping(ctx context.Context) (PingResponse, error) {
...
+	}
+
+	respBodyBytes, err := ioutil.ReadAll(resp.Body)
```
krancour, line 71: iirc, we should `defer resp.body.Close()`

v2/apiserver/internal/system/rest/system_endpoints.go
```diff
@@ -77,6 +83,24 @@ func (h *SystemEndpoints) healthz(w http.ResponseWriter, r *http.Request) {
 	)
 }
 
+// unversionedPing returns the api server version and http.StatusOK.
+// This is handy for auxiliary components to verify their connectivity.
```
krancour, line 87: Should we mention that this exists mainly for the benefit of clients that may have no clue what version of the API they're talking to and might like to find out?

krancour:
All remaining comments are nits and non-blocking. This LGTM!

krancour (approved):

sdk/v2/system/api_client.go
vdice: Oops, indeed; will add.

Issue #1345, opened 2021-04-06
Fix Windows Logger Agent build


vdice:
The Windows logger agent build has been failing.  See for example: https://github.com/brigadecore/brigade/pull/1344/checks?check_run_id=2282294495

Relevant output:

```
Invoking msys2 shell command: -defterm -no-start -c "ps -ef | grep '[?]' | awk '{print $2}' | xargs -r kill"
mkdir: cannot change permissions of '/dev/shm': Permission denied
mkdir: cannot change permissions of '/dev/mqueue': Permission denied
'C:\Windows\system32\drivers\etc\hosts' -> '/etc/hosts'
'C:\Windows\system32\drivers\etc\protocol' -> '/etc/protocols'
'C:\Windows\system32\drivers\etc\services' -> '/etc/services'
'C:\Windows\system32\drivers\etc\networks' -> '/etc/networks'
gpg: /etc/pacman.d/gnupg/trustdb.gpg: trustdb created
gpg: no ultimately trusted keys found
gpg: starting migration from earlier GnuPG versions
gpg: porting secret keys from '/etc/pacman.d/gnupg/secring.gpg' to gpg-agent
gpg: migration succeeded
gpg: Generating pacman keyring master key...
gpg: key 650139CB43ABD93A marked as ultimately trusted
gpg: directory '/etc/pacman.d/gnupg/openpgp-revocs.d' created
gpg: revocation certificate stored as '/etc/pacman.d/gnupg/openpgp-revocs.d/888CD8F9848346FDE21E6470650139CB43ABD93A.rev'
gpg: Done
==> Updating trust database...
gpg: marginals needed: 3  completes needed: 1  trust model: pgp
gpg: depth: 0  valid:   1  signed:   0  trust: 0-, 0q, 0n, 0m, 0f, 1u
==> Appending keys from msys2.gpg...
==> Locally signing trusted keys in keyring...
  -> Locally signing key 123D4D51A1793859C2BE916BBBE514E53E0D0813...
  -> Locally signing key B91BCF3303284BF90CC043CA9F418C233E652008...
  -> Locally signing key 6E8FEAFF9644F54EED90EEA0790AE56A1D3CFDDC...
  -> Locally signing key 69985C5EB351011C78DF7F6D755B8182ACD22879...
  -> Locally signing key 9DD0D4217D75A33B896159E6DA7EF2ABAEEA755C...
[...]

PR #1360, opened 2021-04-14
fix(logger): fix Windows logger agent build

vdice:

- Removes option 2 (pacman update) from the `ridk install` command, which is currently failing (see https://github.com/brigadecore/brigade/issues/1345).  (The issue appears to be fixed upstream in ridk, but has yet to be released.)

Fixes https://github.com/brigadecore/brigade/issues/1345

**If applicable**:


vdice:
@krancour this fixes the build by forgoing the 'pacman update' optional step on `ridk install` (option 2).  I don't think we need this.  I do see that it appears to be [fixed in the installer](https://github.com/oneclick/rubyinstaller2/commit/9017818c301aa92041a647e16421c3a9d62fe1bc) but the installer itself hasn't seen a release since Dec '20.

Now we just run option 3; dev tools install.  See https://github.com/oneclick/rubyinstaller2/tree/master/lib/ruby_installer/build/components and https://stackoverflow.com/questions/44229081/what-do-these-rubyinstaller-2-4-components-do for context, etc.

Question is now: Do we have a way to vet this new image?  I don't have access to a Windows setup currently.

krancour (approved):
LGTM

Issue #1196, opened 2020-12-11
Add clone command


krancour:
1.x had `rerun`. I've always envisioned something similar in 2.0 like `brig event clone --id <event id>`.

vdice:
Here are the current options for v1's `brig rerun`:
```
  brig rerun BUILD_ID [flags]

Flags:
  -b, --background       Trigger the event and exit. Let the job rerun in the background.
  -c, --commit string    A VCS (git) commit
      --config string    The brigade.json config file
  -e, --event string     The name of the event to fire
  -f, --file string      The JavaScript file to execute
  -h, --help             help for rerun
  -l, --level string     Specified log level: log, info, warn, error (default "log")
      --no-color         Remove color codes from log output
      --no-progress      Disable progress meter
  -p, --payload string   The path to a payload file
  -r, --ref string       A VCS (git) version, tag, or branch
  ```
  
For this first implementation, which of these do we want to ensure we have?

I'd propose `-c/--commit`, `-r/--ref` and `-f/--file`.  (Where `-e/--event/-i/--id` is the given/necessary flag to designate the event id to clone)

krancour:
What's the utility in -c, -r, and -f? i.e. If someone is looking to _clone_ an event, why do they want to change details of the event? Is it that really a clone even if they are _changing_ details? Or is it simply a brand new event at that point?

To a certain extent, I have reservations about this functionality. I proposed it for some kind of "parity" with v1.x, but I guess I'm still generally skeptical of the notion of users creating events (other than cli:exec).

Painting a scenario for you:

Imagine that an event from system X comes into Brigade via an X-gateway. The X-gateway legitimately produces an event with certain values in the source and type fields. Should an end-user really be _allowed_ to copy the event and mutate details in the process, producing what is essentially an entirely new event that looks like it came from a particular gateway, but didn't?

I'm not arguing anything one way or the other here, but I do think we need to think through these things and have a clear position on them. If, for instance, we were to begin allowing what's described above, I think we need to add new fields to record the progeny of an Event. i.e. This event's source _says_ it's from X-gateway, but another field denotes that really vdice created it. Just thinking out loud.

vdice:
[...]

PR #1362, opened 2021-04-16
feat(event_commands.go): add clone command

vdice:

- Adds the `brig event clone` command per https://github.com/brigadecore/brigade/issues/1196

Closes https://github.com/brigadecore/brigade/issues/1196

**If applicable**:

### DIFF
--- a/sdk/v2/core/events.go
+++ b/sdk/v2/core/events.go
@@ -232,6 +232,13 @@ type EventsClient interface {
 	List(context.Context, *EventsSelector, *meta.ListOptions) (EventList, error)
 	// Get retrieves a single Event specified by its identifier.
 	Get(context.Context, string) (Event, error)
+	// Clones a pre-existing Event, removing the original's metadata and Worker
+	// config in the process. If the original Event references a Project by ID,
+	// one new Event will be created. Otherwise, the Event provided is treated
+	// as a template and zero or more discrete Events may be created-- one for
+	// each subscribed Project. An EventList is returned containing all newly
+	// created Events.
+	Clone(context.Context, string) (EventList, error)
 	// UpdateSourceState updates source-specific (e.g. gateway-specific) Event
 	// state.
 	UpdateSourceState(context.Context, string, SourceState) error
@@ -320,6 +327,22 @@ func (e *eventsClient) Get(
 			Path:        fmt.Sprintf("v2/events/%s", id),
 			SuccessCode: http.StatusOK,
 			RespObj:     &event,
+		},
+	)
+}
+
+func (e *eventsClient) Clone(
+	ctx context.Context,
+	id string,
+) (EventList, error) {
+	events := EventList{}
+	return events, e.ExecuteRequest(
+		ctx,
+		rm.OutboundRequest{
+			Method:      http.MethodPut,
+			Path:        fmt.Sprintf("v2/events/%s/clone", id),
+			SuccessCode: http.StatusCreated,
+			RespObj:     &events,
 		},
 	)
 }
--- a/sdk/v2/core/events.go
+++ b/sdk/v2/core/events.go
@@ -233,11 +233,8 @@ type EventsClient interface {
 	// Get retrieves a single Event specified by its identifier.
 	Get(context.Context, string) (Event, error)
 	// Clones a pre-existing Event, removing the original's metadata and Worker
-	// config in the process. If the original Event references a Project by ID,
-	// one new Event will be created. Otherwise, the Event provided is treated
-	// as a template and zero or more discrete Events may be created-- one for
-	// each subscribed Project. An EventList is returned containing all newly
-	// created Events.
+	// config in the process.  A new Event is created using the rest of the
+	// details preserved from the original.
 	Clone(context.Context, string) (EventList, error)
 	// UpdateSourceState updates source-specific (e.g. gateway-specific) Event
 	// state.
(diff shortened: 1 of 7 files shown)


krancour:
I'm wondering if this is another case wherein we should move the clone logic to the backend rather than rely on all potential clients to get it right.

vdice:
> I'm wondering if this is another case wherein we should move the clone logic to the backend rather than rely on all potential clients to get it right.

Sure.  We'd probably want to do the same for retry as well, huh?  I can add a note to that ticket as well.

krancour:
> Sure. We'd probably want to do the same for retry as well, huh? I can add a note to that ticket, if so.

I believe so.

vdice:
Ok, now updated with clone logic handled by the backend event service.  PTAL when convenient @krancour

sdk/v2/core/events.go
```diff
@@ -232,6 +232,13 @@ type EventsClient interface {
 	List(context.Context, *EventsSelector, *meta.ListOptions) (EventList, error)
 	// Get retrieves a single Event specified by its identifier.
 	Get(context.Context, string) (Event, error)
+	// Clones a pre-existing Event, removing the original's metadata and Worker
+	// config in the process. If the original Event references a Project by ID,
+	// one new Event will be created. Otherwise, the Event provided is treated
+	// as a template and zero or more discrete Events may be created-- one for
+	// each subscribed Project. An EventList is returned containing all newly
+	// created Events.
```
krancour, line 240: By the time an event has an ID, it's a discrete event, so copy(eventID) can't possibly be copying a template. I think everything is good and just the comment here has to change.

sdk/v2/core/events.go
```diff
@@ -324,6 +331,22 @@ func (e *eventsClient) Get(
...
+func (e *eventsClient) Clone(
+	ctx context.Context,
+	id string,
+) (EventList, error) {
+	events := EventList{}
+	return events, e.ExecuteRequest(
+		ctx,
+		rm.OutboundRequest{
+			Method:      http.MethodPut,
+			Path:        fmt.Sprintf("v2/events/%s/clone", id),
```
krancour, line 343: REST nit. Post to `.../clones` might be better. Given that an event could be cloned multiple times, and therefor _have_ multiple clones, "clones" is a collection of sorts. Adding something to a collection is, more conventionally, a post.

v2/apiserver/internal/core/events.go
```diff
@@ -532,6 +535,29 @@ func (e *eventsService) GetByWorkerToken(
 	return event, nil
 }
 
+func (e *eventsService) Clone(
+	ctx context.Context,
+	id string,
+) (EventList, error) {
+	events := EventList{}
+
+	if err := e.authorize(ctx, system.RoleReader(), ""); err != nil {
```
krancour, line 544: I think this is the wrong role. One would need to have `core.ProjectUser` (for the event's project) in order to be authorized to create any event at all.

v2/apiserver/internal/core/events.go
```diff
@@ -532,6 +535,29 @@ func (e *eventsService) GetByWorkerToken(
...
+	if err != nil {
+		return events, errors.Wrapf(err, "error retrieving event %q from store", id)
+	}
+
+	// Clone all event details *except* metadata and worker config
+	clone := event
+	clone.ObjectMeta = meta.ObjectMeta{}
+	clone.Worker = Worker{}
+
+	return e.Create(ctx, clone)
```
krancour, line 558: Nice re-use.

v2/apiserver/internal/core/rest/events_endpoints.go
```diff
@@ -40,6 +40,12 @@ func (e *EventsEndpoints) Register(router *mux.Router) {
 		e.AuthFilter.Decorate(e.get),
 	).Methods(http.MethodGet)
 
+	// Clone event
+	router.HandleFunc(
+		"/v2/events/{id}/clone",
```
krancour, line 45: See https://github.com/brigadecore/brigade/pull/1362/files#r616007108

krancour:
Overall looks great. There are just a few minor things to fix.

One additional thought I had is I think it might be useful to add a `cloneOf` field (or similar name) to the `Event` struct (both server-side and in the SDK) so that there's some kind of traceability back to the original event.

v2/apiserver/internal/core/events.go
vdice: Here I decided to remove the authz prior to the `store.Get` call -- and just defer to the authz checks in the subsequent `Create()` call.  I noticed that `UpdateSourceState`, `Cancel`, `Delete` all follow this pattern.  WDYT?

vdice:
> One additional thought I had is I think it might be useful to add a cloneOf field (or similar name) to the Event struct (both server-side and in the SDK) so that there's some kind of traceability back to the original event.

Oh that's an interesting idea.  I forgot to add this in the last commit.  It sounds like this addition has the potential to be useful, but I'm wary of adding until we have strong motivations to do so.  That being said, what are your thoughts of adding a label to the `Labels` field on a cloned `Event`, e.g. `cloneOf=<event id>` instead of adding a new field to the `Event` struct?

v2/apiserver/internal/core/events.go
krancour: Maybe just leave a comment that makes note of that?

krancour (approved):
LGTM

Issue #1371, opened 2021-04-21
errors occur if you try to get logs from a worker or job that isn't yet in a running state


krancour:

vdice:
Thanks for filing.  I wonder if this was a regression that snuck in.  I just hit it and wanted to paste in the apiserver logs for tracing down:

`2021/04/21 18:57:29 error retrieving log stream for event "cb69a8cb-4f3b-4761-89cf-d066d31652ad": error opening log stream for pod "cb69a8cb-4f3b-4761-89cf-d066d31652ad" in namespace "brigade-simple-pipeline-83f2ae00-5b12-4dfb-af95-c9e7bfddd698": container "worker" in pod "cb69a8cb-4f3b-4761-89cf-d066d31652ad" is waiting to start: ContainerCreating`

krancour:
I think somewhere when making the decision to go after warm v cool logs, we do a check (or a de-facto check) that the pod exists. I guess if, for instance, a pod is _starting_ but not fully running yet (maybe the git-initializer is running, for instance) then the pod exists but the _container_ we're looking for doesn't.

krancour:
This was closed by #1373

PR #1373, opened 2021-04-22
add --follow flag to cli event create command

### LINKED ISSUE #1329: Do we want a `--follow` option for `brig event create`

krancour:
Fixes #1329 and (incidentally) #1371

### DIFF
--- a/v2/apiserver/internal/core/kubernetes/logs_store.go
+++ b/v2/apiserver/internal/core/kubernetes/logs_store.go
@@ -11,6 +11,7 @@ import (
 	"github.com/brigadecore/brigade/v2/apiserver/internal/core"
 	"github.com/brigadecore/brigade/v2/apiserver/internal/meta"
 	myk8s "github.com/brigadecore/brigade/v2/internal/kubernetes"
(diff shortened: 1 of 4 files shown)


vdice (commented):
Looking good; may need a few updates from the last merge to get CI green.

vdice (approved):

PR #1375, opened 2021-04-22
minor clone cleanup

krancour:
This simplifies the client and service interfaces for cloning an `Event`. It was previously returning an `EventList` but cloning an `Event` can only ever produce _one_ new `Event`, so it seems simpler to return just an `Event` instead of an `EventList`.

This PR also removes an unused bit of the `MockEventStore`.


vdice (approved):
🧹 Good call; thanks for spiffing up!  LGTM.

Issue #1199, opened 2020-12-12
Add retry command


krancour:
Add a `brig event retry` command, where the actual retry logic is handled by the backend EventsClient.

Bubbling up some notes from discussion below:

I'm not tracking down related issues until after I've had my coffee, but we've seen requests before for a "retry" command that skips jobs that succeeded already and retries those that didn't. We've always said that's near impossible to do because our shared storage isn't a layered file system (which seems it would be a tough hill to climb) and that means we don't have the capacity to restore shared state to what it was prior to a job's failure before a job is retried. I think a compromise that we arrived at that did make it into the v2 proposal doc was that we could implement this sort of "retry" specifically for workers/jobs that don't use a volume to share state.

An Event struct has a Worker field and that is its blueprint for how the Event will be processed-- a battle plan if you will. It's constructed from the associated Project's WorkerTemplate. The Event's Worker also aggregates Job states.

We could think of retry as "copy the event and, provided there is no shared state involved, copy the battle plan also; clear the state on failed jobs." We could then amend the logic for Job creation to short circuit when a "new" job is created, but it already exists and is already in a success state.

vdice:
@krancour how does this command differ from the `clone` command in https://github.com/brigadecore/brigade/issues/1199 ? 

This would be for retrying an event run, right?  e.g. `brig event retry --id <event id>`

krancour:
I think we need to talk and map out clone, retry, etc.

I have thoughts, but nothing firm. Video chat tomorrow?

krancour:
We can still chat later, but here are some thoughts that came to me last night.

"Retry" and "clone," in my mind, are perhaps backed by different intentions.

"Clone" is, perhaps, "make a copy of this event and process it."

I'm not tracking down related issues until after I've had my coffee, but we've seen requests before for a "retry" command that skips jobs that succeeded already and retries those that didn't. We've always said that's near impossible to do because our shared storage isn't a layered file system (which seems it would be a tough hill to climb) and that means we don't have the capacity to restore shared state to what it was prior to a job's failure before a job is retried. I think a compromise that we arrived at that _did_ make it into the v2 proposal doc was that we could implement this sort of "retry" specifically for workers/jobs that don't use a volume to share state.

An Event struct has a Worker field and that is its blueprint for how the Event will be processed-- a battle plan if you will. It's constructed from the associated Project's WorkerTemplate. The Event's Worker also aggregates Job states.

[...]

PR #1376, opened 2021-04-23
feat(events): add retry command

vdice:

* Adds an event retry command
* Supports retrying all non-successful jobs, assuming they require no shared workspace

Closes https://github.com/brigadecore/brigade/issues/1199

**If applicable**:


v2/apiserver/internal/core/events.go
```diff
@@ -403,13 +406,21 @@ func (e *eventsService) createSingleEvent(
 
 	event.ID = uuid.NewV4().String()
 
+	jobs := []Job{}
 	workerSpec := project.Spec.WorkerTemplate
+	// If the event is a retry of another, defer to the Worker.Spec on the event
+	// itself.  Retain the Jobs slice as well.
+	if event.Labels != nil && event.Labels["retryOf"] != "" {
```
krancour, line 413: If this is going to be a label, we might consider using a key that's less prone to collision by prefixing with `brigade.sh/`. Come to think of it, we should have done the same for the `clone` command.

v2/apiserver/internal/core/jobs.go
```diff
@@ -245,6 +245,10 @@ func (j *jobsService) Create(
 	eventID string,
 	job Job,
 ) error {
+	var useWorkspace = job.Spec.PrimaryContainer.WorkspaceMountPath != ""
+	var usePrivileged = job.Spec.PrimaryContainer.Privileged
+	var useDockerSocket = job.Spec.PrimaryContainer.UseHostDockerSocket
```
krancour, line 250: I see you moved these lines up from farther down and that's fine, but there's a loop also that should have moved with it. It loops over the sidecar containers to determine if any of _them_ use these features.

v2/apiserver/internal/core/jobs.go
```diff
@@ -253,15 +257,44 @@ func (j *jobsService) Create(
...
+		// Else, deal with retry event job
+		//
+		// If the job has already succeeded, simply return, as we don't want to
+		// replace this job in the store or schedule again
+		if originalJob.Status != nil &&
+			originalJob.Status.Phase == JobPhaseSucceeded {
+			return nil
+		}
+		// We don't currently support retrying jobs that require a shared workspace
+		if useWorkspace && event.Worker.Spec.UseWorkspace {
```
krancour, line 283: I think the determination of whether an event can be retried needs to be made at the event-level when the retry is first requested. I have two reasons for this. The first being that it seems friendlier to inform the user up front if an event cannot be retried than to blindly attempt the retry and then have the retry fail.

The second reason is that it's not actually safe to, on a job-by-job basis, say "this job doesn't use the shared workspace so it can be retried." Here's a scenario to illustrate:

A worker is _supposed_ to run jobs A, B, and C in serial. A and C use the shared workspace. B does not.

A succeeds on the first attempt. B fails. The failed event has no record of a job named C because the worker never got as far as creating job C.

On retry, A is skipped because it succeeded. B is retried because it previously failed and _does not_ depend on the shared workspace. When B succeeds, the worker will create job C (for the first time). Job C, being dependent on the shared workspace, might have been dependent on things that job A should have done-- except job A didn't rerun, which means the shared workspace may not be in the expected state.

This is why there's a `UseWorkspace` field on the `WorkerSpec` type. We should probably just have a look at that field when retry is requested and refuse to retry if it's `true`.

krancour:
Apart from in-line comments, there's one other issue here that I know of-- logs.

When requesting logs for an event, you can optionally specify a particular job (instead of getting worker logs by default). If a job X didn't run for a particular event, but is just a cached result, the logs for that job X need to come from the original event's job X.

This probably requires a new field in Job (or JobStatus) that points back to the original logs. We can seamlessly act on that information when logs are requested.

That being said... this creates a slight problem if event B is a retry of event A and event A has been deleted. But as long as we at least know that event B job X logs should come from event A job X, we can _at least_ inform the user of the fact that the logs in question are no longer available.

vdice:
> Apart from in-line comments, there's one other issue here that I know of-- logs.

Looking at this next.  Thanks for raising.

v2/apiserver/internal/core/events.go
```diff
@@ -812,6 +835,49 @@ func (e *eventsService) DeleteMany(
...
+	if err != nil {
+		return Event{}, errors.Wrapf(
+			err,
+			"error retrieving event %q from store",
+			id,
+		)
+	}
+
+	// Only allow retry if the event Worker has reached a terminal phase
+	if !event.Worker.Status.Phase.IsTerminal() {
```
krancour, line 855: I think it would be appropriate to return a `*meta.ErrConflict` here. The REST machinery ultimately translates that to a 409.

v2/apiserver/internal/core/events.go
```diff
@@ -403,13 +417,21 @@ func (e *eventsService) createSingleEvent(
...
 	event.ID = uuid.NewV4().String()
 
+	jobs := []Job{}
 	workerSpec := project.Spec.WorkerTemplate
+	// If the event is a retry of another, defer to the Worker.Spec on the event
+	// itself.  Retain the Jobs slice as well.
+	if event.Labels != nil && event.Labels[RetryLabelKey] != "" {
+		workerSpec = event.Worker.Spec
+		jobs = event.Worker.Jobs
+	}
```
krancour, line 427: Don't we want to be more selective about which jobs we copy over? I think, offline, we had discussed copying over jobs/status only for jobs that succeeded _and_ didn't use the shared workspace. Anything else shouldn't be inherited because it needs to be re-run.

v2/apiserver/internal/core/jobs.go
```diff
@@ -286,6 +275,42 @@ func (j *jobsService) Create(
...
+			return &meta.ErrConflict{
+				Type: JobKind,
+				ID:   job.Name,
+				Reason: fmt.Sprintf(
+					"Event %q already has a job named %q.",
+					eventID,
+					job.Name,
+				),
+			}
+		}
```
krancour, line 290: This looks problematic because it during a single retry, you could get away with creating multiple jobs with the same name.

But see https://github.com/brigadecore/brigade/pull/1376#discussion_r619673271

If we used a strategy that decides up front what job results were re-usable in the retry, then most of this logic here goes away.

v2/apiserver/internal/core/events.go
vdice: Yup, good call.  Updated.

v2/apiserver/internal/core/jobs.go
vdice: Ok, logic here updated/simplified.  PTAL when convenient.

vdice:
> This probably requires a new field in Job (or JobStatus) that points back to the original logs. We can seamlessly act on that information when logs are requested.

For cached job logs, I went this route.  Good suggestion.  Let me know how this looks.

sdk/v2/core/events.go
```diff
@@ -249,6 +249,9 @@ type EventsClient interface {
 	// DeleteMany deletes multiple Events specified by the EventListOptions
 	// parameter.
 	DeleteMany(context.Context, EventsSelector) (DeleteManyEventsResult, error)
+	// Retry copies an Event, including Worker configuration and Jobs, and
+	// creates a new Event from this information.
```
krancour, line 253: May want to update the comment to reflect that "where possible" successful job results are reused and are not re-run.

sdk/v2/core/events.go
```diff
@@ -418,6 +421,22 @@ func (e *eventsClient) DeleteMany(
...
+func (e *eventsClient) Retry(
+	ctx context.Context,
+	id string,
+) (Event, error) {
+	event := Event{}
+	return event, e.ExecuteRequest(
+		ctx,
+		rm.OutboundRequest{
+			Method:      http.MethodPost,
+			Path:        fmt.Sprintf("v2/events/%s/retry", id),
```
krancour, line 433: REST nit. Since there could tentatively be more than one retry, perhaps this should end with `/retries`?

v2/apiserver/internal/core/events.go
```diff
@@ -290,6 +301,9 @@ type EventsService interface {
 		context.Context,
 		EventsSelector,
 	) (DeleteManyEventsResult, error)
+	// Retry copies an Event, including Worker configuration and Jobs, and
+	// creates a new Event from this information.
```
krancour, line 305: Same comment as https://github.com/brigadecore/brigade/pull/1376/files#r620520095

v2/apiserver/internal/core/events.go
```diff
@@ -403,13 +417,30 @@ func (e *eventsService) createSingleEvent(
...
+		for _, job := range event.Worker.Jobs {
+			if job.Status.Phase == JobPhaseSucceeded && !job.UsesWorkspace() {
+				// Capture event ID for logs, if not already set.
+				if job.Status.LogsEventID == "" {
+					job.Status.LogsEventID = event.Labels[RetryLabelKey]
+				}
+				jobs = append(jobs, job)
+			}
+		}
+	}
```
krancour, line 436: 👍

v2/apiserver/internal/core/jobs.go
```diff
@@ -254,6 +269,13 @@ func (j *jobsService) Create(
 		return errors.Wrapf(err, "error retrieving event %q from store", eventID)
 	}
 	if _, ok := event.Worker.Job(job.Name); ok {
+		// If we have a match and this is a retry, the job must be one pre-selected
+		// to be cached (has succeeded and does not use a shared workspace),
+		// therefore, return now to prevent rescheduling.
+		if event.Labels != nil || event.Labels[RetryLabelKey] != "" {
+			return nil
+		}
```
krancour, line 277: Suppose the job that is requested to be created varies from the original job in some way... we should detect that and return an error.

There are ways that this could happen...

Suppose a worker does something non-deterministic-- like creates a random string-- and uses it as input to a job (command argument or env var perhaps).

When we detect such a difference, we cannot reasonably use the cached result.

vdice: Ah, indeed... looking at this next.

vdice:
@krancour the latest commit adds logic to check if a job eligible for caching matches the original.  If not, we currently error out -- but do you have any thoughts around https://github.com/brigadecore/brigade/pull/1376#discussion_r620583015 ?

vdice:
Another Q, @krancour -- I added a new `LogsEventID` field to the `JobStatus` struct in `v2/apiserver/internal` and not to the `JobStatus` struct in the `sdk` pkg.  Therefore, I didn't think the job-status schema would need to be updated, as the field is only meant for internal use.  WDYT?

v2/apiserver/internal/core/jobs.go
krancour: I agree. I think the tricky thing in that scenario, however, will be how we invalidate the inherited result once and only once.

We _shouldn't_ replace a job that wasn't an inherited one or else a single worker could get away with creating job X over and over.

I think that should be doable, because I think you already added some field to the job status that points back to the original event. If that's non-empty, I suppose the job was inherited and is eligible to be replaced.

This could create a strange scenario where someone looking at a new event that's a retry of an existing event could see jobs that initially appear to have been successful (inherited status) and _then_ they "restart" because the worker re-submitted the jobs with different input than the originals...

It gets me thinking that maybe we need to add something along the lines of a `INHERITED_SUCCESS` job phase. Terrible name. We'll have to come up with something better. But the point would be for users who have retried an event to be able to easily distinguish whether a job just ran or whether its result was inherited from a previous run. I think this is probably something to think about in a follow-up.

v2/apiserver/internal/core/events.go
```diff
@@ -403,13 +419,30 @@ func (e *eventsService) createSingleEvent(
...
+	jobs := []Job{}
 	workerSpec := project.Spec.WorkerTemplate
+	// If the event is a retry of another, defer to the Worker.Spec on the event
+	// itself.  Retain/cache any successful jobs that do not use a shared
+	// workspace.
+	if event.Labels != nil && event.Labels[RetryLabelKey] != "" {
+		workerSpec = event.Worker.Spec
+		for _, job := range event.Worker.Jobs {
+			if job.Status.Phase == JobPhaseSucceeded && !job.UsesWorkspace() {
+				// Capture event ID for logs, if not already set.
```
krancour, line 431: "If not already set"... am I interpretting this correctly that this will do the right thing in the case of a retry of a retry?

krancour:
> I added a new LogsEventID field to the JobStatus struct in v2/apiserver/internal and not to the JobStatus struct in the sdk pkg. Therefore, I didn't think the job-status schema would need to be updated, as the field is only meant for internal use. WDYT?

Agreed.

v2/apiserver/internal/core/events.go
vdice: Indeed.  I'll make this clearer in the comment.

krancour: Glad you had already thought of that scenario. It hadn't occurred to me until late lastnight. Went to check and you had it covered already!

vdice:
I realize I need to shuffle around some unit tests now that logic has migrated between events.go and jobs.go...  will push an update soon.

v2/apiserver/internal/core/jobs.go
vdice: Hmmm, right.  For this PR, should we just leave as returning the conflict error?

v2/apiserver/internal/core/containers.go
```diff
@@ -23,3 +23,50 @@ type ContainerSpec struct {
 	// to be set within the OCI container.
 	Environment map[string]string `json:"environment,omitempty" bson:"environment,omitempty"` // nolint: lll
 }
+
+// DeepEquals returns a boolean value indicating whether or not a ContainerSpec
+// is equivalent to another.
+func (c ContainerSpec) DeepEquals(c2 ContainerSpec) bool {
```
krancour, line 29: I'm worried that we'll overlook updating this if we ever add a field.

I'm generally opposed to using reflection in production code because it's slow, but I wonder if `reflect.DeepEqual()` may be our best option for this.

v2/apiserver/internal/core/events.go
```diff
@@ -290,6 +301,11 @@ type EventsService interface {
 		context.Context,
 		EventsSelector,
 	) (DeleteManyEventsResult, error)
+	// Retry copies an Event, including Worker configuration and Jobs, and
+	// creates a new Event from this information.  Where possible, jobs are not
+	// re-scheduled and their results re-used, for example when a job has
+	// succeeded and does not make use of a shared workspace.
+	Retry(context.Context, string) (Event, error)
```
krancour, line 308: Nit: The word "cache" could be misleading. I wonder if it might be more accurate to say we "inherit" results where possible?

v2/apiserver/internal/core/events.go
```diff
@@ -403,13 +419,32 @@ func (e *eventsService) createSingleEvent(
...
+				// Capture event ID for tracing original logs.  Note that it may
+				// already be set (e.g. via a retry of another retry), in which case
+				// we do not want to override it.
+				if job.Status.LogsEventID == "" {
+					job.Status.LogsEventID = event.Labels[RetryLabelKey]
+				}
+				jobs = append(jobs, job)
+			}
+		}
+	}
```
krancour, line 440: I'm wondering if it may end up feeling cleaner to move retry-specific code into the `retry()` function instead of conditionally doing this stuff `createSingleEvent()` when it's determined that we're dealing with a retry... or you may have had reasons for doing it this way...

v2/apiserver/internal/core/jobs.go
```diff
@@ -84,6 +84,39 @@ type Job struct {
...
+	}
+	for container, spec := range j.Spec.SidecarContainers {
+		if j2Spec, ok := j2.Spec.SidecarContainers[container]; !ok ||
+			!spec.DeepEquals(j2Spec) {
+			return false
+		}
+	}
+	return true
+}
+
```
krancour, line 119: Same comment as [here](https://github.com/brigadecore/brigade/pull/1376/files#r622553704)

v2/apiserver/internal/core/jobs.go
```diff
@@ -142,6 +175,17 @@ type JobContainerSpec struct {
...
+// JobContainerSpec is equivalent to another.
+func (j JobContainerSpec) DeepEquals(j2 JobContainerSpec) bool {
+	return j.ContainerSpec.DeepEquals(j2.ContainerSpec) &&
+		j.WorkingDirectory == j2.WorkingDirectory &&
+		j.WorkspaceMountPath == j2.WorkspaceMountPath &&
+		j.SourceMountPath == j2.SourceMountPath &&
+		j.Privileged == j2.Privileged &&
+		j.UseHostDockerSocket == j2.UseHostDockerSocket
+}
+
```
krancour, line 188: Same comment as [here](https://github.com/brigadecore/brigade/pull/1376/files#r622553704)

v2/apiserver/internal/core/jobs.go
```diff
@@ -164,6 +208,9 @@ type JobStatus struct {
 	Ended *time.Time `json:"ended,omitempty" bson:"ended,omitempty"`
 	// Phase indicates where the Job is in its lifecycle.
 	Phase JobPhase `json:"phase,omitempty" bson:"phase,omitempty"`
+	// LogsEventID indicates which event ID the job logs are associated with.
+	// This is useful for looking up logs for a cached job for retry events.
```
krancour, line 212: Nit: Another spot where "inherited" might feel more accurate than "cached."

v2/apiserver/internal/core/jobs.go
```diff
@@ -253,7 +300,29 @@ func (j *jobsService) Create(
...
+					Type: JobKind,
+					ID:   job.Name,
+					Reason: fmt.Sprintf(
+						"Job %q is not eligible for caching, as it is not equivalent "+
+							"to the original job from event %q",
+						job.Name,
+						event.Labels[RetryLabelKey],
+					),
+				}
+			}
```
krancour, line 321: I'm ok with this behavior, at least near term, but I thought you had wanted to throw away the inherited result in this scenario and proceed with scheduling a new job.

v2/apiserver/internal/core/logs.go
```diff
@@ -168,6 +169,20 @@ func (l *logsService) Stream(
...
+		// as job may be cached and carried over on a retry event
+		if job.Status != nil && job.Status.LogsEventID != "" {
+			event, err = l.eventsStore.Get(ctx, job.Status.LogsEventID)
+			if err != nil {
+				return nil,
+					errors.Wrapf(
+						err,
+						"error retrieving original event %q from store",
+						job.Status.LogsEventID,
+					)
```
krancour, line 183: If the event is gone, the error we get back with be a `*meta.ErrNotFound`. Because that's the innermost of the nested errors, the client will get a 404 with a message body that indicates <event id of original event> not found. That could end up being pretty misleading...

I am wondering if we should trap `*meta.ErrNotFound`, specifically, here and raise an all new error that will be more informative.

krancour:
Looking good. Added a few comments. This is a complex feature. (I appreciate you tackling it!) Just want to make sure we've covered all bases.

v2/apiserver/internal/core/jobs.go
vdice: Oh yup, had forgotten.  Logic updated.

v2/apiserver/internal/core/events.go
vdice: Good call.  Job inheritance logic moved to Retry.

v2/apiserver/internal/core/containers.go
vdice: Good concern.  Ok, updated to use reflect.

v2/apiserver/internal/core/jobs.go
```diff
@@ -84,6 +85,18 @@ type Job struct {
 	Status *JobStatus `json:"status" bson:"status"`
 }
 
+// UsesWorkspace returns a boolean value indicating whether or not the job
+// uses a shared workspace.
+func (j Job) UsesWorkspace() bool {
+	var usesWorkspace = j.Spec.PrimaryContainer.WorkspaceMountPath != ""
```
krancour, line 91: Nit: You can short-circuit here and return right away if the above is true.

v2/apiserver/internal/core/jobs.go
```diff
@@ -84,6 +85,18 @@ type Job struct {
...
+	for _, sidecarContainer := range j.Spec.SidecarContainers {
+		if sidecarContainer.WorkspaceMountPath != "" {
+			usesWorkspace = true
```
krancour, line 94: Nit: Could also short circuit right here.

v2/apiserver/internal/core/jobs.go
```diff
@@ -253,15 +270,24 @@ func (j *jobsService) Create(
...
+			return &meta.ErrConflict{
+				Type: JobKind,
+				ID:   job.Name,
+				Reason: fmt.Sprintf(
+					"Event %q already has a job named %q.",
+					eventID,
+					job.Name,
+				),
+			}
+		}
```
krancour, line 285: Are we still missing a check somewhere in this vicinity?

I think even if the event itself is a try, creating a job with the same name as an existing one should raise a conflict if the job that would be replaced wasn't inherited.

v2/apiserver/internal/core/logs.go
krancour: This appears to still be an issue.

v2/apiserver/internal/core/jobs.go
vdice: Oh, in that case, I was thinking we'd just proceed with re-scheduling the job with the same name, instead of returning a conflict error.  WDYT?

krancour: But wouldn't that let a single worker create a job with the same name more than once as long as the even is a retry?

vdice: Ahh yes, thank you.  Took me a bit to grok.  Will update.

vdice: I've changed it back to just returning the ErrConflict for all cases except for when the job is a retry and its configuration matches the original.  I'd rather revisit the corner cases around this at a future date when needed.

krancour: sgtm

v2/apiserver/internal/core/logs.go
vdice: Updated here -- did I address the concern as envisioned?

v2/apiserver/internal/core/logs.go
```diff
@@ -168,6 +170,24 @@ func (l *logsService) Stream(
...
+
+		// Check to see if we need to look up logs via a specific event ID,
+		// as job may be cached and carried over on a retry event
+		if job.Status != nil && job.Status.LogsEventID != "" {
+			event, err = l.eventsStore.Get(ctx, job.Status.LogsEventID)
+			if err != nil {
+				if _, ok := err.(*meta.ErrNotFound); ok {
+					return nil,
+						fmt.Errorf("error retrieving logs for job %q", job.Name)
+				}
```
krancour, line 182: This is a tricky situation. If the user is looking for the logs of a particular job from a particular event Y that is a retry of event X, but that job in question was inherited from X and X is now gone, and we were to respond by saying 404 "couldn't find X" it's confusing to the user who wasn't actually _looking_ for X. You're now suppressing the confusing message. 👍 on that.

The down side now, however, is that the new error will surface to the end user as a 500 with no real detail.

Raising a `*meta.ErrConflict` with a detailed message might be our best bet here, for lack of a better option.

Also, I'm ok with this happening in a follow-up.

v2/apiserver/internal/core/jobs.go
krancour: But the logic looks wrong. Shouldn't it be this this (difference is `||` vs `&&`):

```
if event.Labels != nil || event.Labels[RetryLabelKey] != "" {
```

krancour: Also, I do think we can do something more here in a follow-up.

v2/apiserver/internal/core/logs.go
krancour: Opened a follow-up, so we can consider this resolved for now.

https://github.com/brigadecore/brigade/issues/1403

v2/apiserver/internal/core/jobs.go
vdice: Good catch, it should definitely be `&&` --  a mistake carried over when I had this setup to check the reverse.

v2/apiserver/internal/core/logs.go
vdice: Good call, thanks for creating the follow-up.

Issue #1233, opened 2021-01-21
Git Initializer tests: add ssh url/private repo case


vdice:
Add a test case with an ssh clone url associated with a private repo... something along the lines of:

```
{
	name:          "GitHub - Private Repo - ssh",
	project: core.Project{
		ObjectMeta: meta.ObjectMeta{
			ID: "private-ssh",
		},
		Spec: core.ProjectSpec{
			WorkerTemplate: core.WorkerSpec{
				Git: &core.GitConfig{
					CloneURL: "git@github.com:brigadecore/private-test-repo.git",
					Commit: "a292c2ffe0226f702519ec2833054953e82a816e",
				},
				DefaultConfigFiles: map[string]string{
					"brigade.js": defaultBrigadeJS,
				},
			},
		},
	},
	terminalPhase: core.WorkerPhaseSucceeded,
},
```

The tricky part will be how we want to wire in a private ssh key for use with the corresponding test repo.

See also the following idea for adding the ssh key as a secret to the test project: https://github.com/brigadecore/brigade/pull/1229#discussion_r561962839

Follow-up from https://github.com/brigadecore/brigade/pull/1229

vdice:
I understand how we can wire up a secret (the private ssh key for an existing private test repo) for running these in CI, but I also want these tests to always be runnable by devs locally.  What do we think about _only_ running this private repo test case when CI is running (on GitHub PR or merge/tag events)?

krancour:
Seems reasonable. Can there just be a boolean field in the test case struct that's something like `skipLocally`, and we can set that to `true` for tests of this nature?

vdice:
Added in https://github.com/brigadecore/brigade/pull/1383

PR #1383, opened 2021-04-27
ci(integration): add private repo test case

vdice:

* Adds a private repo test case to the integration test suite

Close https://github.com/brigadecore/brigade/issues/1233

**If applicable**:

### DIFF
--- a/v2/tests/integration_test.go
+++ b/v2/tests/integration_test.go
@@ -25,10 +25,11 @@ const (
 )
 
 type testcase struct {
-	name        string
-	project     core.Project
-	configFiles map[string]string
-	assertions  func(
+	name              string
+	postProjectCreate func(context.Context, sdk.APIClient) error
+	project           core.Project
+	configFiles       map[string]string
+	assertions        func(
 		t *testing.T,
 		ctx context.Context,
 		client sdk.APIClient,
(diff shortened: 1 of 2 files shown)


v2/tests/integration_test.go
```diff
@@ -301,6 +340,12 @@ func TestMain(t *testing.T) {
 	)
 
 	for _, tc := range testcases {
+		// Skip private repo test if required env var not present
+		if tc.name == "GitHub - private repo" &&
+			os.Getenv("BRIGADE_CI_PRIVATE_REPO_SSH_KEY") == "" {
+			continue
+		}
```
krancour, line 347: Is it worth having a more general mechanism that scales better as we (potentially) continue to add test cases that can _only_ run in CI?

I had been thinking we could add a `bool` field to the `testCase` struct like `skipLocally` or something that is `false` by default, but can be set to `true` for the few cases that can only run in CI.

wdyt?

krancour: But consider this non-blocking.

krancour (approved):

v2/tests/integration_test.go
vdice: I did consider that, but opted to just wait on adding the toggle until we had more than one test case to utilize it.  Should be quick to add when we get there!

vdice: (Actually found it handy to be able to run locally _if_ I had the req'd env var set, for testing 😄 )

PR #1384, opened 2021-04-28
log in using GitHub as identity provider

### LINKED ISSUE #1133: Add support for GitHub as identity provider
This is easier said than done because iirc, GitHub does not implement OpenID Connect, but uses some other authentication scheme (on top of OAuth2, similar to OpenID, but different?).

Implementing this might require a new abstraction for identity providers that doesn't assume all identity providers use OpenID Connect.

~~This is a nice to have and not a hard requirement for 2.0.~~

Support for this identity provider would probably garner more community satisfaction with the new authn system then either of the providers we currently support.

krancour:
Fixes #1133 

Originally, there was a binary choice between using an OpenID Connect identity provider or simply not supporting third-party identity providers at all, but GitHub doesn't implement OpenID Connect, and is a "high value" identity provider, since most would-be Brigade 2 users are very likely to already have GitHub accounts. GitHub _does_ permit authentication using OAuth2...

OpenID Connect, itself, is built _on top_ of OAuth2. So even if GitHub doesn't implement OpenID Connect, quite a bit of similarity remains between the process for authenticating using OpenID Connect and authenticating using GitHub.

What this PR does is backs off of the requirement that all third-party identity providers must implement OpenID Connect and instead assumes, more generally, that all third-party identity providers support OAuth2. The very few details that vary between the two are now abstracted behind a `ThirdPartyAuthenticationHelper` interface. An appropriate implementation of that interface is selected at startup on the basis of configuration found in environment variables.

To be a bit more specific-- for _all_ third-party auth, we generate a unique, unguessable "state" and pass that to the identity provider. Upon completion, the identity provider gives back a code, which can be exchanged for a token. In the case of OIDC, that token has user information embedded within it. In the case of GitHub, the token that can be used to retrieve the user's GitHub profile. The only two details that vary between OIDC-based auth and GitHub-based auth are 1. where we send the users to complete authentication and 2. how we extract / obtain the user identity information.

This PR looks bigger than it effectively is. Many of the changes are simply re-wording of comments or renaming of field to be less OIDC-specific and more generally oriented around "third party authentication."


sdk/v2/authn/sessions.go
```diff
@@ -12,45 +12,45 @@ import (
...
-// OIDCAuthOptions encapsulates user-specified options when creating a new
-// session using an OpenID Connect workflow.
-type OIDCAuthOptions struct {
-	// SuccessURL indicates where users should be redirected to after successful
-	// completion of the OpenID Connect authentication workflow. If this is left
-	// unspecified, users will be redirected to a default success page.
+// ThirdPartyAuthOptions encapsulates user-specified options when creating a new
+// Session that will authenticate using a third-party identity provider.
+type ThirdPartyAuthOptions struct {
+	// AuthSuccessURL indicates where users should be redirected to after
```
vdice, line 18:
```suggestion
	// SuccessURL indicates where users should be redirected to after
```

v2/apiserver/internal/authn/github/third_party_auth_helper_test.go
```diff
@@ -0,0 +1,98 @@
...
+	)
+	helper := NewThirdPartyAuthHelper(
+		ThirdPartyAuthHelperConfig{
+			ClientID: testClientID,
+		},
+	)
+	require.Equal(t, testAuthURL, helper.AuthURL(testState))
+}
+
+// TODO: Test this
```
vdice, line 36: Can be removed?

vdice (approved):
LGTM

PR #1390, opened 2021-04-29
streamline first-time login for admins 

### LINKED ISSUE #1386: post install setup needs to be more seamless
By design, new users who "just show up" after having authenticated with a third-party identity provider have no permissions. This is because not everyone known to whatever identity provider is in play necessary is authorized to do _anything_ in Brigade. Imagine a Brigade instance that uses GitHub for authentication. Anyone with a GitHub account can show up to the party, but naturally, we don't want randos showing up and automatically being granted permissions-- even read-only permissions.

This is all fine, except it creates a challenge for the very first user (presumably the person who installed Brigade) who must, currently, authenticate using the third-party identity provider, then re-authenticate as root, grant themselves admin privileges, then re-authenticate as themself.

This is further complicated by the fact that, imho, leaving root access enabled should be discouraged in a production cluster. This means the operator who handles the installation potentially has to install with root access enabled, then re-configure to disable root access after the first user has taken on the admin role.

We could make this much more seamless.

Many pieces of software I have worked with in the past automatically make the first user to register/login the admin. This would work, but I don't care for it because of the opportunity it presents for an unknown party to become the admin just by being first.

What I'd like to propose is that one or more admins can be specified by their ID (email address if using OpenID Connect or GitHub handle if using GitHub) at the time of installation (e.g. in `values.yaml`). When a new user authenticates, if they are matched to an ID in that list, they are automatically granted admin permissions.

krancour:
Fixes #1386

### DIFF
--- a/charts/brigade/values.yaml
+++ b/charts/brigade/values.yaml
@@ -63,6 +63,13 @@ apiserver:
     ## Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
     ## For example, "60s", "2h45m", "168h" (1 week)
     userSessionTTL: 168h
+    ## List user IDs from third-party identity provider (email address if using
+    ## OpenID Connect or GitHub handle if using GitHub) who will AUTOMATICALLY
+    ## be granted system admin permissions the first time they log in. This is
+    ## useful for automating post-install setup. Absent this option, one would
+    ## have to log in, log out, log back in as root, grant themselves
+    ## permissions, then log out, then log back in as themselves.
+    admins: []
 
   tls:
     ## Whether to enable TLS. If true then you MUST either set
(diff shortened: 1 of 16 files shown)


charts/brigade/values.yaml
```diff
@@ -35,7 +35,7 @@ apiserver:
   thirdPartyAuth:
     ## Valid values are "oidc" (for OpenID Connect), "github" (for OAuth2 with
     ## GitHub as the identity provider), or "disabled"
-    strategy: disabled
+    strategy: github
```
vdice, line 38: I don't think we want these changes to go in, do we 😄 ?

I'd suggest leaving with `disabled` as the default still, but definitely keeping the new `admins` section/comments, perhaps with a sample commented-out GH username.

vdice (commented):
I think this is a great way to address https://github.com/brigadecore/brigade/issues/1386.  Other than a comment around the chart values changes, looking good to me.

charts/brigade/values.yaml
krancour: Yep. That was an accident! Thanks for catching!

krancour: Fixed after force push.

vdice (approved):

PR #1391, opened 2021-04-29
Fix marshal job when status is empty

carolynvs:
When status is empty on a job, create fails when called through the rest API because it's marshaled as `"status": null`, and status is not an allowed field on the job document. I have marked it as omitempty so that a nil status isn't included in the json.

```
Bad request: Request body failed JSON validation:
  0. (root): Additional property status is not allowed
```

I'm not 100% sure why this is something that just I ran into implementing brigdrake. When brigade uses its own api, does it not go through the json schema validation?

**If applicable**:


krancour:
Nice catch. You're probably the first to create a real job using the Go SDK. The default worker is implemented in TS.

krancour (approved):
LGTM

PR #1392, opened 2021-04-30
ref(naming): increase/adjust limits on resource names

vdice:

* Update resource name max lengths to 63 where applicable (shared fields of label, url as used by event source, type and others)

* Project IDs now bumped from 18 to 63

* To support the bump for project IDs, I've removed the injection of project ID into the project namespace name itself and instead have added the project ID to the namespace in the form of a label

Users that need to interrogate K8s directly can still quickly see which namespace aligns with their project via:

```
 $ k get ns --show-labels
NAME                                           STATUS   AGE    LABELS
brigade                                        Active   163m   <none>
brigade-81c16560-2b21-4b6c-bc6b-c6d2908b4608   Active   51m    brigade.sh/project=shared-workspace
```
* Update job name schema to ensure max length is < 63 (and abides by K8s container naming restrictions, as a job's primary container gets the same name as the job itself)

Closes https://github.com/brigadecore/brigade/issues/1195

**If applicable**:


krancour:
@vdice I love the approach-- especially to increasing the allowable length for project IDs, which was simply too restrictive before.

My only reservation is the truncation of project ID labels. My worry is that, from a human k8s operator's perspective, resources associated with two projects having very long but similar names could end up being indistinguishable from one another.

Any thoughts on keeping all the other improvements in this PR and just dialing back max project ID length to 63 so the full ID can always fit in a label? Jumping from 18 to 63 is still a giant improvement.

vdice:
> Any thoughts on keeping all the other improvements in this PR and just dialing back max project ID length to 63 so the full ID can always fit in a label? Jumping from 18 to 63 is still a giant improvement.

Agreed, 63 seems like a reasonable max length for sure -- and a big improvement as mentioned.  We now know we _could_ go 253 in the future.  Sure, I'll update project ID to be 63.  

Shall we leave event source and type at 253?

krancour:
> Shall we leave event source and type at 253?

I guess we could leave them that way if we never need to put them in a k8s label, but then we lose the flexibility to do so if we ever find the need to do so.

Maybe 63 is the more cautious choice? That's where my vote goes, but I'm not outrageously passionate about that opinion either.

vdice:
> I guess we could leave them that way if we never need to put them in a k8s label, but then we lose the flexibility to do so if we ever find the need to do so.

I mis-spoke a little bit.  As seen from the changes, I bumped the common/shared definitions for `label` and `url` from 250 maxLength to 253.  Not exactly a huge bump and only did so because K8s does allow the higher max.  `label` is used for event type, labels and qualifiers while `url` is used for event source.  Shall we leave them at 253 for now and revisit when/if we need to create K8s labels from any of these in the future?

krancour:
> Shall we leave them at 253 for now and revisit when/if we need to create K8s labels from any of these in the future?

Again-- I don't feel super strongly, but I do imagine that being a hard thing to dial back in the future. There could be, by then, v2-compatible gateways already identifying themselves with a URL longer than 63 characters.

vdice:
Ok, those fields have now been scaled back to 63 chars max.

krancour (approved):
LGTM

PR #1400, opened 2021-05-05
ref(charts/ingress): update ingress api/usage

vdice:

* Updates the ingress resource in the Brigade chart per deprecation warnings: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122

I switched the apiVersion check to look for 1.19 and above for the final `networking.k8s.io/v1` and otherwise fall back to `networking.k8s.io/v1beta1`, which has been in since 1.14.

**If applicable**:


charts/brigade/templates/apiserver/ingress.yaml
```diff
@@ -19,13 +19,17 @@ spec:
...
+        path: /
+        backend:
+          service:
+            name: {{ include "brigade.apiserver.fullname" . }}
+            port:
+              {{- if .Values.apiserver.tls.enabled }}
+              number: 443
+              {{- else }}
+              number: 80
+              {{- end }}
```
krancour, line 32: Does this new schema work even with the v1beta1 api?

vdice: Oh right, I didn't think to try, but I assume not.  Shall we commit to the new api?  Or I suppose I can update the template to support both by toggling how the schema looks depending on version?

vdice: Straightforward enough to toggle the schema, so I've done so.  Will test against a 1.18.x cluster as well...

vdice: Looks good in my testing.

krancour (approved):
LGTM

Issue #1403, opened 2021-05-07
improve error handling when jobs logs are inherited and not found


krancour:
As a follow up to https://github.com/brigadecore/brigade/pull/1376, in the case where an event is a retry of another and someone goes looking for specific jobs on the new event that should be inherited from the original, but the original event has been deleted in the meantime, we should return a more informative error than the 500 that currently surfaces.

vdice:
I'm trying to recall, do we want to update [this](https://github.com/brigadecore/brigade/blob/v2/v2/apiserver/internal/core/logs.go#L181) to return an ErrConflict-typed error?  How about the error message as currently written?

krancour:
You're remembering correctly. Initially, I thought conflict might be our best option, for lack of a better one.

An alternative I just thought of would be to go with a 404, which I initially resisted, but only if we amend the `ErrNotFound` error type with a new field that allows for an (optional) detailed explanation. We'd eat any `ErrNotFound` we get when retrieving the original event (so the end user doesn't get a weird message about something they weren't directly looking for not being found) and issue a new `ErrNotFound` with a detailed explanation.

Any thoughts on that?

vdice:
I like that.  `ErrNotFound` indeed makes more sense in this context.  I'll look at what it might look like to add a new field to the current struct and how we can use it here...

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1410

PR #1410, opened 2021-05-14
feat(errors.go): add Reason field to ErrNotFound; use in job logs scenario

vdice:

* Adds a `Reason` field to the `ErrNotFound` struct for instances where further context/explanation is helpful
* Use this new field in the scenario when inherited job logs cannot be found due to the original event no longer existing

Closes https://github.com/brigadecore/brigade/issues/1403

**If applicable**:

### DIFF
--- a/sdk/v2/meta/errors.go
+++ b/sdk/v2/meta/errors.go
@@ -56,10 +56,15 @@ type ErrNotFound struct {
 }
 
 func (e *ErrNotFound) Error() string {
-	if e.Reason != "" {
+	if e.Type == "" && e.ID == "" && e.Reason != "" {
 		return e.Reason
 	}
-	return fmt.Sprintf("%s %q not found.", e.Type, e.ID)
+
+	msg := fmt.Sprintf("%s %q not found", e.Type, e.ID)
+	if e.Reason != "" {
+		return msg + fmt.Sprintf(": %s", e.Reason)
+	}
+	return msg + "."
 }
 
 // ErrConflict represents an error wherein a request cannot be completed because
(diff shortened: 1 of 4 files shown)


sdk/v2/meta/errors.go
```diff
@@ -50,9 +50,15 @@ type ErrNotFound struct {
...
 	ID string `json:"id,omitempty"`
+	// Reason is a natural language explanation around why the resource could not
+	// be located.
+	Reason string `json:"reason,omitempty"`
 }
 
 func (e *ErrNotFound) Error() string {
+	if e.Reason != "" {
+		return e.Reason
+	}
```
krancour, line 61: What do you think about offering the explanation in addition to the existing text as opposed to instead of?

vdice: That was indeed my first approach, but I found the output in this specific case slightly awkward... I was going to have Type be JobLogs and ID be the job name... so then the combined error messsage reads: "JobLogs myjobname not found: ....".  Granted, this is a specific case that shouldn't preclude others... but I also envisioned that there might be other cases where Type/ID aren't available or relevant and we just want to supply the reason.

Oh, we could also do it this way: if Type, ID and Reason are non-empty, combine 'em.  if only Type, ID exist, print the standard 'not found' message. If only Reason exists, just print that.  WDYT?

krancour: I understand the concern you're presenting and I like your proposed solution. 👍

krancour (approved):
LGTM

I think this adds a lot of flexibility for scenarios where it would be misleading to say "x doesn't exist" without any further explanation.

Issue #1378, opened 2021-04-23
enforce timeouts on workers and jobs


krancour:
Currently, nothing enforces timeouts on workers and jobs.

This should be the observer component's duty.

vdice:
The job object in brigadier looks to have a default timeout of 15 mins and is configurable via `job.timeout`.  Enforcement still needs to be added in the observer.

For the worker, is there a default timeout already defined?  If not, what should it be?  Do we want this value to be configurable by users (via the chart)?

krancour:
> For the worker, is there a default timeout already defined?

Doesn't look like we have that yet. (iirc, Brigade v1 didn't have it either?)

imho, we should make it user-configurable by adding it to the `WorkerSpec`, but also allow operators to set a max allowable ceiling. We should probably let operators set a max allowable ceiling for jobs also. Many operators might not be tolerant of their end-users submitting workers or jobs that run for days.

krancour:
@vdice you're working on this, right? What's the rough plan of attack? Can I assume it involves the observer starting a timer the first time it observes a new worker or job and using the api to forcefully terminate that worker or job if the timer goes off before it observes completion?

vdice:
@krancour I can switch things up to work that way.  My first approach was to run a calculation each time sync[Job|Worker]Pod was called: check the pod's start time plus timeoutSeconds value -- if less than current time, consider timed out and go from there.  Perhaps this is too naive and or costly?  https://github.com/vdice/brigade/blob/feat/worker-and-job-timeouts/v2/observer/jobs.go#L189-L208

krancour:
@vdice what you just described was actually my first thought as well, but I think there's the potentials for a worker or job to run quite long past the specified timeout if the corresponding pod goes a while without syncing.

vdice:
https://github.com/brigadecore/brigade/pull/1411 adds Job timeout enforcing.

Going to start work on adding the ability to specify a worker timeout (using the same `timeoutDuration` approach agreed-upon in #1411 ) plus operator defaults and of course logic to enforce these values.

vdice:
@krancour When the observer is enforcing a timeout on a worker pod, in addition to updating the worker status and deleting worker resources, what do we want to do about the worker's jobs, if anything?  Do we want to iterate through all of them and if a job has a non-terminal phase, update its status with JobPhaseTimedOut?

krancour:
[...]

PR #1411, opened 2021-05-20
feat(observer): enforce job timeout

vdice:

* Adds enforcement of timeout logic for jobs

This comprises the Job portion of https://github.com/brigadecore/brigade/issues/1378

**If applicable**:

### DIFF
--- a/sdk/v2/core/jobs.go
+++ b/sdk/v2/core/jobs.go
@@ -118,9 +118,12 @@ type JobSpec struct {
 	// sidecar container), then logic within those containers must account for
 	// these constraints.
 	SidecarContainers map[string]JobContainerSpec `json:"sidecarContainers,omitempty"` // nolint: lll
-	// TimeoutSeconds specifies the time, in seconds, that must elapse before a
-	// running Job should be considered to have timed out.
-	TimeoutSeconds int64 `json:"timeoutSeconds,omitempty"`
+	// TimeoutDuration specifies the time duration that must elapse before a
+	// running Job should be considered to have timed out. This duration string
+	// is a sequence of decimal numbers, each with optional fraction and a unit
+	// suffix, such as "300ms", "3.14s" or "2h45m".
+	// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+	TimeoutDuration string `json:"timeoutDuration,omitempty"`
 	// Host specifies criteria for selecting a suitable host (substrate node) for
 	// the Job. This is useful in cases where a Job requires a specific,
 	// non-default operating system (i.e. Windows) or specific hardware (e.g. a
@@ -239,6 +242,9 @@ type JobsClient interface {
 		status JobStatus,
 	) error
 	Cleanup(ctx context.Context, eventID, jobName string) error
+	// Timeout, given an Event identifier and Job name, executes timeout logic
+	// for a Job that has exceeded its timeout limit.
+	Timeout(ctx context.Context, eventID, jobName string) error
 }
 
 type jobsClient struct {
@@ -376,6 +382,25 @@ func (j *jobsClient) Cleanup(
 			Method: http.MethodPut,
 			Path: fmt.Sprintf(
 				"v2/events/%s/worker/jobs/%s/cleanup",
+				eventID,
+				jobName,
+			),
+			SuccessCode: http.StatusOK,
+		},
+	)
+}
+
+func (j *jobsClient) Timeout(
+	ctx context.Context,
+	eventID,
+	jobName string,
+) error {
+	return j.ExecuteRequest(
+		ctx,
+		rm.OutboundRequest{
+			Method: http.MethodPut,
+			Path: fmt.Sprintf(
+				"v2/events/%s/worker/jobs/%s/timeout",
 				eventID,
 				jobName,
 			),
--- a/v2/apiserver/internal/core/jobs.go
+++ b/v2/apiserver/internal/core/jobs.go
@@ -114,9 +114,12 @@ type JobSpec struct {
 	// sidecar container), then logic within those containers must account for
 	// these constraints.
 	SidecarContainers map[string]JobContainerSpec `json:"sidecarContainers,omitempty" bson:"sidecarContainers,omitempty"` // nolint: lll
-	// TimeoutSeconds specifies the time, in seconds, that must elapse before a
-	// running Job should be considered to have timed out.
-	TimeoutSeconds int64 `json:"timeoutSeconds,omitempty" bson:"timeoutSeconds,omitempty"` // nolint: lll
+	// TimeoutDuration specifies the time duration that must elapse before a
+	// running Job should be considered to have timed out. This duration string
+	// is a sequence of decimal numbers, each with optional fraction and a unit
+	// suffix, such as "300ms", "3.14s" or "2h45m".
+	// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+	TimeoutDuration string `json:"timeoutDuration,omitempty" bson:"timeoutDuration,omitempty"` // nolint: lll
 	// Host specifies criteria for selecting a suitable host (substrate node) for
 	// the Job. This is useful in cases where a Job requires a specific,
 	// non-default operating system (i.e. Windows) or specific hardware (e.g. a
@@ -231,6 +234,9 @@ type JobsService interface {
 	// Cleanup removes Job-related resources from the substrate, presumably
 	// upon completion, without deleting the Job from the data store.
 	Cleanup(ctx context.Context, eventID, jobName string) error
+	// Timeout updates a Job's status to indicate it has timed out and proceeds
+	// to cleanup Job-related resources from the substrate.
+	Timeout(ctx context.Context, eventID, jobName string) error
 }
 
 type jobsService struct {
@@ -641,6 +647,42 @@ func (j *jobsService) Cleanup(
 		)
 	}
 	return nil
+}
+
+func (j *jobsService) Timeout(
+	ctx context.Context,
+	eventID string,
+	jobName string,
+) error {
+	if err := j.authorize(ctx, RoleObserver, ""); err != nil {
+		return err
+	}
+
+	status, err := j.GetStatus(ctx, eventID, jobName)
+	if err != nil {
+		return errors.Wrapf(
+			err,
+			"error retrieving status for event %q job %q",
+			eventID,
+			jobName,
+		)
+	}
+
+	// Update job status
+	now := time.Now()
+	status.Phase = JobPhaseTimedOut
+	status.Ended = &now
+
+	if err := j.UpdateStatus(ctx, eventID, jobName, status); err != nil {
+		return errors.Wrapf(
+			err,
+			"error updating status for event %q job %q",
+			eventID,
+			jobName,
+		)
+	}
+
+	return j.Cleanup(ctx, eventID, jobName)
 }
 
 // JobsStore is an interface for components that implement Job persistence
--- a/v2/apiserver/internal/core/kubernetes/substrate.go
+++ b/v2/apiserver/internal/core/kubernetes/substrate.go
@@ -1172,6 +1172,9 @@ func (s *substrate) createJobPod(
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      myk8s.JobPodName(event.ID, jobName),
 			Namespace: project.Kubernetes.Namespace,
+			Annotations: map[string]string{
+				myk8s.AnnotationTimeoutDuration: fmt.Sprint(jobSpec.TimeoutDuration),
+			},
 			Labels: map[string]string{
 				myk8s.LabelComponent: myk8s.LabelKeyJob,
 				myk8s.LabelProject:   event.ProjectID,
--- a/v2/apiserver/schemas/job.json
+++ b/v2/apiserver/schemas/job.json
@@ -120,9 +120,11 @@
 						"$ref": "#/definitions/containerSpec"
 					}
 				},
-				"timeoutSeconds": {
-					"type": "integer",
-					"description": "Job timeout in seconds"
+				"timeoutDuration": {
+					"type": "string",
+					"description": "Job timeout string expressed as a sequence of decimal numbers, each with optional fraction and a unit suffix, such as '300ms', '3.14s' or '2h45m'",
+					"pattern": "^([0-9]+(\\.[0-9]*)?(ns|us|µs|ms|s|m|h))+$",
+					"minLength": 2
 				},
 				"host": {
 					"$ref": "#/definitions/host"
--- a/v2/brigadier-polyfill/src/jobs.ts
+++ b/v2/brigadier-polyfill/src/jobs.ts
@@ -39,7 +39,7 @@ export class Job extends BrigadierJob {
           spec: {
             primaryContainer: this.primaryContainer,
             sidecarContainers: this.sidecarContainers,
-            timeoutSeconds: this.timeout,
+            timeoutDuration: this.timeoutSeconds + "s",
             host: this.host
           }
         },
--- a/v2/internal/kubernetes/common.go
+++ b/v2/internal/kubernetes/common.go
@@ -7,6 +7,8 @@ import (
 )
 
 const (
+	AnnotationTimeoutDuration = "brigade.sh/timeoutDuration"
+
 	LabelComponent = "brigade.sh/component"
 	LabelEvent     = "brigade.sh/event"
 	LabelJob       = "brigade.sh/job"
--- a/v2/observer/jobs.go
+++ b/v2/observer/jobs.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/brigadecore/brigade/sdk/v2/core"
 	myk8s "github.com/brigadecore/brigade/v2/internal/kubernetes"
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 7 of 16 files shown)


v2/observer/jobs.go
```diff
@@ -46,6 +47,16 @@ func (o *observer) syncJobPods(ctx context.Context) {
 func (o *observer) syncJobPod(obj interface{}) {
 	pod := obj.(*corev1.Pod)
 
+	// Fetch the cancel func associated with a timed job pod
+	lookup := pod.Namespace + "/" + pod.Name
+	cancelTimer, exists := o.timedPodsSet[lookup]
```
krancour, line 52: Nit. If you _were_ to successfully find a `cancelTimer` in the map here, you don't ever end up using it, so the code is slightly easier to follow here if you use `_, exists := o.timedPodsSet[lookup]` and only define `cancelTimer` within the conditional below if/when it is necessary.

krancour: I think we also need to protect this map with a mutex.

The inner workings of the k8s lister/watcher is somewhat opaque to me, but I am at least questioning the possibility that the `syncJobPod()` function can be invoked more than once, concurrently, on the _same_ pod. For instance, syncing one given pod's state n _and_ state n + 1 concurrently.

(Or maybe the lister/watcher is built to avoid that condition, otherwise, there would be race conditions all of the place in k8s, right? Bottom line is I simply don't know. Perhaps we need to research this.)

At any rate... the reason I bring it up is because their _may_ be a need to protect this map with a mutex so that we never accidentally start multiple timer loops for the same pod.

v2/observer/jobs.go
```diff
@@ -176,3 +191,67 @@ func (o *observer) deleteJobResources(
...
+				duration,
+				pod.Name,
+			),
+		)
+		return
+	}
+
+	go func() {
+		ticker := time.NewTicker(timeout)
+		defer ticker.Stop()
```
krancour, line 220: I'd suggest a `time.Timer` over a `time.Ticker`. The timer goes off once like an alarm clock instead of repeatedly at intervals.

v2/observer/jobs.go
```diff
@@ -136,7 +150,8 @@ func (o *observer) syncJobPod(obj interface{}) {
 	}
 
 	if status.Phase == core.JobPhaseSucceeded ||
-		status.Phase == core.JobPhaseFailed {
+		status.Phase == core.JobPhaseFailed ||
+		status.Phase == core.JobPhaseTimedOut {
```
vdice, line 154: @krancour how do we feel about adding a check for `JobPhaseUnknown` here as well?  Then, the `cancelTimer()` call can be consolidated here as well.  Or, was it a design decision to _not_ delete job resources if/when the phase is unknown?

krancour: Good question. _In general_ I'm not quite sure what to do about unkown. It's a state we have only because I know it's _possible_ for the underlying pod to enter a state that k8s considers unkown. This is mostly a mystery to me. I've never seen it happen.

Let me think on this for a bit.

v2/observer/jobs.go
krancour: After a convo with @thomastaylor312, I withdraw any concern that a mutex may be needed to protect concurrent access to this map.

vdice:
@krancour ok, a handful of changes came through after testing this e2e:

- Needed to add `TIMED_OUT` as an acceptable job/worker phase in the corresponding schemas
- Since the `timeout` property of a `job` in brigadier is in milliseconds and our Job struct in go has `timeoutSeconds`, we must divide the former by 1000 to get a value for the latter.  To allow for non-integer results, I've updated the schema and golang values for these fields (`number` and `float64` respectively). 

For the latter updates, although they work perfectly fine, I feel like we should consider unifying one one unit for this timeout value.  Update the brigadier job timeout field to be in seconds?  Or, update the core API struct to be in ms?  Thoughts?

krancour:
> For the latter updates, although they work perfectly fine, I feel like we should consider unifying on one unit for this timeout value. Update the brigadier job timeout field to be in seconds? Or, update the core API struct to be in ms? Thoughts?

What would be great is if we could use `time.Duration` in the Go code instead of int or float64, but I'm _sure_ that comes with a lot of other complications-- like how that serializes and how one language might support that and another might not...

It _might_ be worth looking at, but I'm fully expecting we have to just settle on numeric units-- in which case, I think seconds is better (i.e. Brigadier should change to match).

vdice:
> What would be great is if we could use time.Duration in the Go code instead of int or float64

@krancour I had an epiphany both from this and from talking w/ @thomastaylor312 (who immediately spotted the inefficiencies around the conversions we/I were doing with different units of time to time.Duration, etc.)  Why not change `TimeoutSeconds` on the go struct to `TimeoutDuration` of type string?  Then, we can process as `time.Duration` when we get to actually starting the timer -- but we gain flexibility around whatever type of unit of time consumers want to specify, as long as they follow the time.Duration protocol.  If we like this, we can also add regex verification of the `timeoutDuration` property of the job schema, to catch invalid values at REST request time, all but ensuring the value will be parse-able/valid at runtime.  WDYT?

krancour:
@vdice so it stays a string right up until we use it? sgtm

vdice:
> @vdice so it stays a string right up until we use it? sgtm

Yup!  I can look at adding regex validation to this updated schema field...

krancour:
@vdice ya. That sounds really good.

sdk/v2/core/jobs.go
```diff
@@ -118,9 +118,12 @@ type JobSpec struct {
...
 	SidecarContainers map[string]JobContainerSpec `json:"sidecarContainers,omitempty"` // nolint: lll
-	// TimeoutSeconds specifies the time, in seconds, that must elapse before a
-	// running Job should be considered to have timed out.
-	TimeoutSeconds int64 `json:"timeoutSeconds,omitempty"`
+	// TimeoutDuration specifies the time duration that must elapse before a
+	// running Job should be considered to have timed out. This duration string
+	// is a possibly signed sequence of decimal numbers, each with optional
+	// fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m".
+	// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+	TimeoutDuration string `json:"timeoutSeconds,omitempty"`
```
thomastaylor312, line 126: Should probably have a new json tag now to match the new struct name, right?

sdk/v2/core/jobs.go
```diff
@@ -118,9 +118,12 @@ type JobSpec struct {
...
 	// these constraints.
 	SidecarContainers map[string]JobContainerSpec `json:"sidecarContainers,omitempty"` // nolint: lll
-	// TimeoutSeconds specifies the time, in seconds, that must elapse before a
-	// running Job should be considered to have timed out.
-	TimeoutSeconds int64 `json:"timeoutSeconds,omitempty"`
+	// TimeoutDuration specifies the time duration that must elapse before a
+	// running Job should be considered to have timed out. This duration string
+	// is a possibly signed sequence of decimal numbers, each with optional
+	// fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m".
+	// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
```
krancour, line 125: What does it mean to have a negative duration timeout?

v2/apiserver/internal/core/kubernetes/substrate.go
```diff
@@ -1172,6 +1172,9 @@ func (s *substrate) createJobPod(
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      myk8s.JobPodName(event.ID, jobName),
 			Namespace: project.Kubernetes.Namespace,
+			Annotations: map[string]string{
+				"timeoutDuration": fmt.Sprint(jobSpec.TimeoutDuration),
```
krancour, line 1176: Should this annotation key be prefixed with something like `brigade.sh`? iirc, we've done that in other places. I think that's a widely observed practice to avoid adding annotations that might accidentally be meaningful to some other, unspecified controller that is working in the same cluster.

v2/apiserver/schemas/job.json
```diff
@@ -120,9 +120,11 @@
 						"$ref": "#/definitions/containerSpec"
 					}
 				},
-				"timeoutSeconds": {
-					"type": "integer",
-					"description": "Job timeout in seconds"
+				"timeoutDuration": {
+					"type": "string",
+					"description": "Job timeout string expressed as a possibly signed sequence of decimal numbers, each with optional fraction and a unit suffix, such as '300ms', '-1.5h' or '2h45m'",
+					"pattern": "^[-+]?([0-9]+(\\.[0-9]*)?(ns|us|µs|ms|s|m|h))+$",
```
krancour, line 126: Same question about negative durations.

v2/apiserver/schemas/job-status.json
```diff
@@ -36,7 +36,7 @@
 		"phase": {
 			"type": "string",
 			"description": "The job's phase",
-			"enum": [ "ABORTED", "CANCELED", "FAILED", "PENDING", "RUNNING", "SCHEDULING_FAILED", "STARTING", "SUCCEEDED", "UNKNOWN" ]
+			"enum": [ "ABORTED", "CANCELED", "FAILED", "PENDING", "RUNNING", "SCHEDULING_FAILED", "STARTING", "SUCCEEDED", "TIMED_OUT", "UNKNOWN" ]
```
krancour, line 39: I think it wasn't an accident that `TIMED_OUT` was missing here.

More on this in another comment. Will link back to here...

v2/observer/jobs.go
```diff
@@ -176,3 +191,64 @@ func (o *observer) deleteJobResources(
...
+				o.errFn(
+					errors.Wrapf(
+						err,
+						"error updating status for event %q worker job %q",
+						eventID,
+						jobName,
+					),
+				)
+			}
+			go o.deleteJobResourcesFn(pod.Namespace, pod.Name, eventID, jobName)
```
krancour, line 250: Instead of updating the status visa API and then going directly to the substrate to delete job-related resources, imho, it would be preferable to take the same "API is the brains of this outfit" route that we have taken elsewhere-- invoke a new "timeout job" operation on the jobs client and let the API do all the work.

This is why `TIME_OUT` was initially missing from the [job status update schema](https://github.com/brigadecore/brigade/pull/1411/files#r639028280). It was left out with the assumption that the observer doesn't directly _observe_ a pod that has timed out and then have a need to report that status, rather it was assumed the observer would observe that time has elapsed for a running pod and can _request_ the API server to execute the timeout.

sdk/v2/core/jobs.go
vdice: Good question -- I can't think of a real-world case for this in tandem with how we use this field.  I'll remove that example and ensure the schema only allows non-negative numbers.

v2/observer/jobs.go
vdice: Ah, I see.  Sounds good.  I will transition this logic into a new API client func/operation.

vdice:
Pushed the smaller fixes/changes first.  Now will work on the API timeout operation, etc.

v2/apiserver/internal/core/jobs.go
```diff
@@ -643,6 +649,31 @@ func (j *jobsService) Cleanup(
...
+) error {
+	if err := j.authorize(ctx, RoleObserver, ""); err != nil {
+		return err
+	}
+
+	// Update job status
+	status := JobStatus{
+		Phase: JobPhaseTimedOut,
+	}
+	if err := j.UpdateStatus(ctx, eventID, jobName, status); err != nil {
```
krancour, line 665: Is there other information in the job's status that we're losing here? iirc, there's a start and end time. Do we want to preserve the start time and add in the end time?

v2/apiserver/internal/core/jobs.go
```diff
@@ -643,6 +649,31 @@ func (j *jobsService) Cleanup(
...
+		return errors.Wrapf(
+			err,
+			"error updating status for event %q job %q",
+			eventID,
+			jobName,
+		)
+	}
+
+	return j.Cleanup(ctx, eventID, jobName)
```
krancour, line 674: Oh wow. That was easy!

v2/brigadier-polyfill/src/jobs.ts
```diff
@@ -39,7 +39,7 @@ export class Job extends BrigadierJob {
           spec: {
             primaryContainer: this.primaryContainer,
             sidecarContainers: this.sidecarContainers,
-            timeoutSeconds: this.timeout,
+            timeoutDuration: this.timeout + "ms",
```
krancour, line 42: Is the `+ "ms"` here because the Brigadier `Job` type specifies timeout in ms?

```
/**
   * The duration, in milliseconds, after which Brigade should automatically
   * terminate and fail the job if it has not completed. The default is 15
   * minutes.
   */
  public timeout: number = defaultTimeout
```

Do you think we should consider changing that? It was a choice that, iirc, was inherited from Brigadier 1.x.

It does seem a bit onerous, to me, to ask people to think/work in milliseconds when that level of precision probably isn't required and also cannot be guaranteed.

v2/internal/kubernetes/common.go
```diff
@@ -7,6 +7,8 @@ import (
 )
 
 const (
+	AnnotationTimeoutDuration = "brigade.sh/timeoutDuration"
```
krancour, line 10: 👍

v2/observer/jobs.go
```diff
@@ -176,3 +190,49 @@ func (o *observer) deleteJobResources(
...
+				duration,
+				pod.Name,
+			),
+		)
+		return
+	}
+
+	go func() {
+		ticker := time.NewTimer(timeout)
+		defer ticker.Stop()
```
krancour, line 219: I still think this can be a timer instead of a ticker.

v2/observer/jobs.go
```diff
@@ -76,10 +87,13 @@ func (o *observer) syncJobPod(obj interface{}) {
...
 	case corev1.PodSucceeded:
 		status.Phase = core.JobPhaseSucceeded
+		cancelTimer()
 	case corev1.PodFailed:
 		status.Phase = core.JobPhaseFailed
+		cancelTimer()
 	case corev1.PodUnknown:
 		status.Phase = core.JobPhaseUnknown
+		cancelTimer()
 	}
```
krancour, line 97: Besides invoking `cancelTimer()`, we also need to remove the `cancelTimer` func from the map or else we'll slowly leak memory by indefinitely retaining a pointer to every cancel function for every job.

krancour: Probably the best thing to do is make the `cancelTimer` function remove _itself_ from the map when it exits.

v2/observer/jobs.go
```diff
@@ -176,3 +190,49 @@ func (o *observer) deleteJobResources(
 		)
 	}
 }
+
+func (o *observer) startJobPodTimer(ctx context.Context, pod *corev1.Pod) {
```
krancour, line 194: Somewhere around here, I would do `defer delete(o.timedPodsSet, namespacedPodName(pod.Namespace, pod.Name))`

See https://github.com/brigadecore/brigade/pull/1411/files?file-filters%5B%5D=.go#r639772237

krancour:
@vdice this is in great shape.

I did add a few comments about making sure we clean up the map of cancelation functions whenever a timeout timer is canceled.

Other than that, I think we should consider whether Brigadier shoudl really ask users to specify timeout in milliseconds (a choice inherited from 1.x).

These are all pretty minor things. This looks great.

v2/brigadier-polyfill/src/jobs.ts
vdice: I'd be in support of changing, for sure.  Seconds, perhaps?

krancour: That. I think seconds would be great.

Units like 1m or 24h seem like they'd be great too, but idk that JS/TS have an easy/convenient method of easily converting those to/from numeric values in a way that's consistent with Go and I definitely wouldn't want user to lose the ability to do their own calculations on that value. So sticking with a `Number` seems best and seconds seems like a reasonable unit.

Should the field's name maybe change to `timeoutSeconds` to eliminate all ambiguity?

krancour: Clarifying:

> Should the field's name maybe change to timeoutSeconds to eliminate all ambiguity?

I'm talking about the field in the TS Job type.

On the Go end, `TimeoutDuration` is fine.

v2/observer/jobs.go
vdice: Oh yes, thank you for identifying.

vdice:
Ok @krancour, should be ready to revisit when convenient.

v2/apiserver/internal/core/jobs.go
```diff
@@ -643,6 +649,42 @@ func (j *jobsService) Cleanup(
...
```
krancour, line 685: Non-blocking nit: There's room for a small optimization here. When you call `j.Cleanup(...)`, it's going to go back to the database to get the Event you've already got in hand. It's possible to refactor so that `j.Cleanup(...)` uses a private helper like `j.cleanup(...)` that accepts Event or Job, etc. as args. That helper could be invoked directly here to avoid the extra db call.

Again... this is non-blocking.

krancour:
Registered one non-blocking nit, but this LGTM.

krancour (approved):

v2/apiserver/internal/core/jobs.go
vdice: Good call.  I'll add/update in https://github.com/brigadecore/brigade/pull/1419

vdice:
Somewhere in there I forgot to sign a commit.  I was going to squash before merging anyways, so I just pushed the squashed/rebased commit now.

PR #1416, opened 2021-05-25
fix event schema

krancour:
Discovered this issue with schema that probably snuck in around the time we DRYed it up. It prevents events from including git cloneURL, commit, or ref.

This issue didn't surface sooner for lack of a brigade2-compatible GitHub (or other git-based SVN) gateway.

Hacking away on https://github.com/krancour/brigade-github-gateway exposed the issue.


vdice:
Ah, I remember this.  I think it was set up the previous way to allow for a Worker's Git section to have the additional property of `initSubmodules` (in addition to CloneURL, Commit, Ref) -- but _not_ allow additional properties for an Event's Git config.

The former aspect is covered in the integration tests and so we see the failure here:
```
=== RUN TestMain/GitHub-job_fails
integration_test.go:359:
Error Trace: integration_test.go:359
Error: Received unexpected error:
Bad request: Request body failed JSON validation:
0. spec.workerTemplate.git: Additional property initSubmodules is not allowed
1. spec.workerTemplate.git: Must validate all the schemas (allOf)
```

I wonder if there's another way to allow for what we want, but still retain common definitions.  Maybe instead of attempting to share the entire `Git` schema def, we break out the common CloneURL, Commit and Ref defs and compose these into separate Git configs for Event and Worker?

krancour:
Oh. Damn. I'm glad you knew the history of why it was like this!

I just had to refer back to JSON schema 7 specification to see if there was any such thing as inheritance and there is not and the authors recognize it as a shortcoming that they intend to address in a future release.

> Maybe instead of attempting to share the entire Git schema def, we break out the common CloneURL, Commit and Ref defs and compose these into separate Git configs for Event and Worker?

It seems then that this might be the best route.

krancour:
@vdice ok. Just took another crack at this using the approach you suggested.

vdice (approved):
Nice.  Looks great!

krancour:
Thanks for the help with this one @vdice !

PR #1419, opened 2021-05-25
Feat/worker timeout

vdice:

* Adds Worker timeout configurability and enforcement

Currently in draft form as it builds/depends on https://github.com/brigadecore/brigade/pull/1411.  Will rebase when 1411 is merged...

**If applicable**:


v2/apiserver/internal/core/workers.go
```diff
@@ -390,6 +402,85 @@ func (w *workersService) UpdateStatus(
...
+	}
+
+	// Update all applicable worker job statuses with JobPhaseAborted
+	// TODO: go func here?
+	// TODO: How about errors?  Do we want to compile and report only
+	// after all have been processed?
+	for _, job := range event.Worker.Jobs {
+		if job.Status != nil && !job.Status.Phase.IsTerminal() {
+			status := *job.Status
+			status.Phase = JobPhaseAborted
```
krancour, line 456: This might not always be right.

iirc, when we cancel workers, they go to ABORTED if they were RUNNING, but they go to CANCELED if they were PENDING.

We should probably do the same here. Timed out worker == RUNNING jobs ABORTED and PENDING jobs CANCELED.

krancour: Just gave myself a refresher on how we cancel events in the db and it involves some fairly complex updates.

I'm almost thinking the best thing to do here is add a dedicated timeout operation on the worker store so that we can quickly and efficiently change the status of all associated jobs that are either pending or starting/running.

vdice: > I'm almost thinking the best thing to do here is add a dedicated timeout operation on the worker store so that we can quickly and efficiently change the status of all associated jobs that are either pending or starting/running.

I'll give this approach a go.  I should look at the `Cancel` and `CancelMany` implementations in events_store for similar logic, right?

krancour: That sounds like a good starting point.

vdice: Great call on this one.  Updated in https://github.com/brigadecore/brigade/pull/1419/commits/3162975063ba74beaf30dfaeddb4f60d4005f920; PTAL when convenient.

v2/apiserver/internal/core/mongodb/workers_store.go
```diff
@@ -54,3 +57,95 @@ func (w *workersStore) UpdateStatus(
...
+	if err != nil {
+		return errors.Wrapf(
+			err,
+			"error updating status of event %q worker",
+			eventID,
+		)
+	}
+	if res.MatchedCount == 1 {
+		return nil
+	}
```
krancour, line 92: I don't think a pending worker could (or should) timeout, right? The clock only starts ticking after the worker's pod has been created, right? Or am I overlooking something?

vdice: Oh, yes, if a worker with phase WorkerPhasePending is guaranteed to still be waiting on a pod, we should remove this.  I wasn't sure if there'd be cases where a worker would have a pod during this phase.

krancour: >  I wasn't sure if there'd be cases where a worker would have a pod during this phase.

Shouldn't be possible because we move workers to a `STARTING` state before trying to create them in the substrate.

krancour (approved):
LGTM!

vdice:
(Rebased and addressed merge conflicts with last push)

PR #1421, opened 2021-05-26
Log access fix

### LINKED ISSUE #1420: Bug: Creator of an event isn't guaranteed access to that event's logs, but should be
Hacking away on a Brigade2-compatible GitHub gateway, I've discovered that my gateway-- the source of certain events-- isn't authorized to retrieve the logs for the events it created.

As things currently stand, accessing the logs for any project's events requires the `PROJECT_USER` role for the project the event belongs to.

I'd propose that for the purposes or reporting to upstream event sources (GitHub in this case), logs should be accessible to anyone having `EVENT_CREATOR` with scope equal to the event's source field. (i.e. the same conditions that allowed creation of the event in the first place.)

I'm counting this as a bug because the way I'm saying it _should_ work is the way it was _supposed_ to work in the first place.

krancour:
Fixes #1420

### DIFF
--- a/v2/apiserver/internal/core/logs.go
+++ b/v2/apiserver/internal/core/logs.go
@@ -6,7 +6,9 @@ import (
 	"fmt"
 	"time"
 
+	libAuthz "github.com/brigadecore/brigade/v2/apiserver/internal/lib/authz"
 	"github.com/brigadecore/brigade/v2/apiserver/internal/meta"
+	"github.com/brigadecore/brigade/v2/apiserver/internal/system"
 	myk8s "github.com/brigadecore/brigade/v2/internal/kubernetes"
 	"github.com/brigadecore/brigade/v2/internal/retries"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


v2/apiserver/internal/core/logs.go
```diff
@@ -150,7 +155,13 @@ func (l *logsService) Stream(
 	// in order to stream logs.
 	if err =
 		l.projectAuthorize(ctx, event.ProjectID, RoleProjectUser); err != nil {
-		return nil, err
+		// The principal wasn't an authorized Project user, but we also allow log
+		// access for the creator of the event so that gateways can send logs
+		// upstream.
+		if err =
+			l.authorize(ctx, system.RoleEventCreator, event.Source); err != nil {
```
vdice, line 162: Would it be prudent to add a couple finer-grained unit tests around this?  (Say, role is right but event source doesn't match (error), both match (no error))?

krancour: The authorizers all have their own tests, so they are known to function correctly.

[This case](https://github.com/brigadecore/brigade/blob/v2/v2/apiserver/internal/lib/authz/role_assignments_test.go#L32-L41) covers correct role, incorrect source.

vdice (approved):
LGTM

Issue #1402, opened 2021-05-07
improve job creation logic for retried events


krancour:
As a follow-up to https://github.com/brigadecore/brigade/pull/1376, we should improve what happens when an event is a retry of another and a job is created for the new event with configuration that differs from the job that was inherited from the original event.

Currently this results in a 409, but we could, instead, discard the inherited job and create the job as requested.

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1422

PR #1422, opened 2021-05-26
ref(jobs.go): re-create retry job if config differs

vdice:

* Add logic to re-create a job on a retried event if the configuration has changed.

There's one outstanding corner case that would remain.  Currently, this would allow multiple jobs with the same name but differing configuration to be re-created (on a retry event).  I'm wondering if we remove this possibility by putting the responsibility of ensuring there are no duplicate entries in the Worker.Jobs list to the underlying data store.  Currently, we use mongodb and utilize a [$push](https://github.com/brigadecore/brigade/blob/v2/v2/apiserver/internal/core/mongodb/jobs_store.go#L37) directive to add a job to a worker in the store (which allows duplicates).  We could instead change to [$addToSet](https://docs.mongodb.com/manual/reference/operator/update/addToSet/) to ensure only one is ever added (first one wins).  WDYT @krancour ?

Closes https://github.com/brigadecore/brigade/issues/1402

**If applicable**:

### DIFF
--- a/v2/apiserver/internal/core/jobs.go
+++ b/v2/apiserver/internal/core/jobs.go
@@ -279,7 +279,9 @@ func (j *jobsService) Create(
 		return errors.Wrapf(err, "error retrieving event %q from store", eventID)
 	}
 	if originalJob, ok := event.Worker.Job(job.Name); ok {
-		// If this is not a retry event, return ErrConflict.
+		// If this is not a retry event or the original job hasn't been inherited
+		// (if it had, the LogsEventID field would be non-empty), return
+		// ErrConflict.
 		if event.Labels == nil || event.Labels[RetryLabelKey] == "" {
 			return &meta.ErrConflict{
 				Type: JobKind,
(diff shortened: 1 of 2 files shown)


vdice:
> We could instead change to $addToSet to ensure only one is ever added (first one wins). WDYT @krancour ?

Actually, thinking through it more, that wouldn't be what we want.  (We'd more want a 'last one wins' enforcement).  I'll keep brainstorming but still interested in the approach you'd prefer @krancour.

krancour:
Let me a think for a bit and review some older threads. We've definitely talked about this at some point and I can't recall how far we got toward a solution.

krancour:
@vdice the `JobStatus` type has a `LogsEventID` field that only has a value if the Job was inherited as part of a retry.

We can look at that to determine if this is an invalid attempt to replace a recently-created job or a _valid_ attempt to replace an inherited job.

vdice:
Thanks @krancour, how does it look now?

v2/apiserver/internal/core/jobs.go
```diff
@@ -279,24 +279,39 @@ func (j *jobsService) Create(
...
-		// If we're dealing with a job intended for retry, inspect to be sure the
-		// provided job configuration matches the original.  If so, we will skip
-		// re-scheduling and inherit the original's results.
-		if event.Labels != nil && event.Labels[RetryLabelKey] != "" {
-			if reflect.DeepEqual(originalJob, job) {
-				return nil
+		// If this is not a retry event or the original job hasn't been inherited
+		// (if it had, the LogsEventID field would be non-empty), return
+		// ErrConflict.
+		if event.Labels == nil || event.Labels[RetryLabelKey] == "" {
```
krancour, line 285: Nit: The comment doesn't quite match the conditional below it.

v2/apiserver/internal/core/jobs.go
```diff
@@ -279,24 +279,39 @@ func (j *jobsService) Create(
...
-				eventID,
-				job.Name,
-			),
+
+		// Else, check to see if the job configurations match.
+		// If they do, return and inherit the original's results.
+		// Otherwise, proceed with the re-creation of the job *if* it has
+		// been inherited previously (LogsEventID field non-empty) -- return
+		// ErrConflict if it hasn't been inherited.
+		if reflect.DeepEqual(originalJob, job) {
```
krancour, line 302: Should we maybe only compare the job specs?

krancour: I ask because I think the `originalJob` is likely to have status and the new `job` is not, which would make these never equal.

vdice: Ah, indeed.  Updated.

krancour (approved):
LGTM

PR #1431, opened 2021-05-28
don't vendor node dependencies

krancour:
It was my own doing a year or two ago that we started vendoring node dependencies. I won't rehash my reasons for doing it, but it did prove to be a somewhat controversial decisions that for some reason never did get rolled back.

Fast forward to the present day and GitHub, via dependabot, is frequently offering to do minor or patch-level version bumps to remediate known dependencies. For as long as we vendor our node dependencies, the contents of our `node_modules/` dir end up stale every time we merge one of those PRs from the dependabot...

For me this is the final straw, because if my own controversial decision from a year or two ago is now making it more onerous than necessary to promptly remediate known vulnerabilities, that can't stand.

This PR should make things right.

Note: v2 already doesn't vendor JS deps (or Go deps either, for that matter).


radu-matei (approved):
❤️

PR #1437, opened 2021-06-03
brig init

willie-yao:
This PR adds the init command to the brig CLI. The init command initializes a base project.yaml file that users can easily get started on using Brigade with. The command has three flags:

- ID (-i, --id): Specifies the ID of the project (required)
- Language (-i, --language): Specifies the language of the Brigade project file (js or ts)
- Git (-g, --git): Specifies a Git repo to pull an existing Brigade project file from

The init command creates the project.yaml file under the .brigade directory. If the directory does not exist, the command should create the directory and place a project.yaml file in the directory. If a project.yaml file already exists inside an existing .brigade directory, then the command will prompt the user to ask if they want to overwrite it.


v2/cli/init_command.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/urfave/cli/v2"
+)
+
+type InitFields struct {
+	ID       string
+	Language string
+	Git      string
+}
+
```
krancour, line 21: This type is only used in one place, for one very specific purpose, so let's move it closer to where it's used-- down around what is now line 82-- and yes, you can define types inside a function. Those types can only be used _within_ said function, but that's just fine.

v2/cli/init_command.go
```diff
@@ -0,0 +1,149 @@
...
+spec:
+	workerTemplate:
+		logLevel: DEBUG
+		git:
+			cloneURL: {{ .Git }}
+`)
+
+// Handle init CLI flags
+func initialize(c *cli.Context) error {
+	filename := ".brigade/project.yaml"
```
krancour, line 80: Nit: Make this a const:

`const filename = ".brigade/project.yaml"`

v2/cli/init_command.go
```diff
@@ -0,0 +1,149 @@
...
+	// allowedLanguages := []string{"js", "ts"}
```
krancour, line 81: I know this is a draft still, but I assume you'll validate that the value provided is on of the allowed ones?

I would also suggest tolerating the words "javascript" and "typescript". Also, we don't need to be picky about case.

v2/cli/init_command.go
```diff
@@ -0,0 +1,149 @@
...
+	buf := new(bytes.Buffer)
+
+	tmpl.Execute(buf, fields)
+
+	// Check if the .brigade/project.yaml path exists. Confirm overwrite if so.
+	if fileExists(filename) {
+		response := false
+		prompt := &survey.Confirm{
+			Message: "A project.yaml file already exists in your .brigade directory, would you like to overwrite it?",
+		}
```
krancour, line 116: Not sure if this is already the case, but make sure No is the default.

v2/cli/init_command.go
```diff
@@ -0,0 +1,149 @@
...
+
+// fileExists checks if a file exists and is not a directory before we
+// try using it to prevent further errors.
+func fileExists(filename string) bool {
+	info, err := os.Stat(filename)
+	if os.IsNotExist(err) {
+		return false
+	}
+	return !info.IsDir()
+}
```
krancour, line 149: We already have this in the `v2/internal/file` package. Reuse that instead.

v2/cli/init_command.go
krancour: Nevermind. I see this check does occur further on down below. May want to delete this comment then.

v2/cli/init_command.go
```diff
@@ -0,0 +1,149 @@
...
+	language := strings.ToLower(c.String(flagLanguage))
+	gitRepo := c.String(flagGit)
+
+	// Check if inputted language is valid
+	if language == "" {
+		language = "ts"
+	} else if language != "js" && language != "ts" {
+		fmt.Println("Language flag must be set to either js (JavaScript) or ts (TypeScript)")
+		return nil
+	}
```
krancour, line 93: Suggest rewriting as follows:

```
switch (string.ToLower(language)) {
  case: "ts", "typescript":
    // whatever
  case: "js", "javascript":
    // whatever
  default:
    return errors.Errorf("Unrecognized value %q for --language flag", language)
}
```
The most important things here are:

* Not case sensitive
* Easier to amend when we add support for other languages (2.1)
* Error is returned and existing global error handling takes care of it

v2/cli/init_command.go
```diff
@@ -0,0 +1,149 @@
...
+
+	// Initialize fields to fill in the template yaml files
+	fields := InitFields{projectID, language, gitRepo}
```
krancour, line 96: It's always better to be explicit about what fields get what values:

```
fields := InitFields{
  ProjectID: projectID,
  Language: language,
  GitRepo: gitRepo,
}
```

This helps avoids accidents that are easily made by simply putting these values in the wrong positions.

v2/cli/init_command.go
```diff
@@ -0,0 +1,149 @@
...
+kind: Project
+metadata:
+	id: {{ .ID }}
+description: Your newly-initialized Brigade project!
+spec:
+	workerTemplate:
+		logLevel: DEBUG
+		git:
+			cloneURL: {{ .Git }}
+`)
```
krancour, line 76: Let's try to consolidate these into a single template.

I say this because right now you have one template for a project that uses git and another for a project that doesn't. Suppose, however, I told you that the default script need to be a little more complex than just `console.log("Hello, World!")`. Now you could be in a position of having a template with git support + a javascript (with JS style imports and such) and a template with git support + a typescript (with TS style imports and such). The number of permutations grows exponentially with the number of options we expose, so it's better to have one template that has some conditional logic in it.

There is what I might describe as an "expression language" called [sprig](https://github.com/Masterminds/sprig) that you can use in order to do conditionals and other complex things within your templates.

v2/cli/init_command.go
```diff
@@ -0,0 +1,149 @@
...
+	// Initialize fields to fill in the template yaml files
+	fields := InitFields{projectID, language, gitRepo}
+
+	if gitRepo != "" {
+		newProjectFileBytes = newGitProjectFileBytes
+	}
+
+	tmpl, err := template.New("newProjectYaml").Parse(string(newProjectFileBytes))
+	if err != nil {
+		log.Fatal(err)
```
krancour, line 104: Don't do this here. Return the error and let the CLI's global error handling take care of it.

v2/cli/init_command.go
```diff
@@ -0,0 +1,148 @@
...
+		logLevel: DEBUG
+		{{- if eq .Git ""}}
+		defaultConfigFiles:
+			brigade.{{ .Language }}: |
+				console.log("Hello, World!");
+		{{- else}}
+		git:
+			cloneURL: {{ .Git }}
+		{{- end}}
+`)
```
krancour, line 62: Let's spike this template out into its own file.

v2/cli/init_command.go
```diff
@@ -0,0 +1,148 @@
...
+			Name:    flagGit,
+			Aliases: []string{"g"},
+			Usage:   "Initialize the brigade project from an existing git repo ",
+		},
+	},
+	Action: initialize,
+}
+
+// Template project.yaml file for a non-git project
+var newProjectFileBytes = []byte(
```
krancour, line 43: Drop a `// nolint: lll` about this line. It will excuse line length issues for this whole large string we're building.

v2/cli/init_command.go
```diff
@@ -0,0 +1,158 @@
...
+
+// Handle init CLI flags
+func initialize(c *cli.Context) error {
+	var err error
+
+	projectID := c.String(flagID)
+	language := strings.ToLower(c.String(flagLanguage))
+	gitRepo := c.String(flagGit)
+
+	// Check if inputted language is valid
```
radu-matei, line 56: nit: `input language`

v2/cli/init_command.go
```diff
@@ -0,0 +1,158 @@
...
+	switch language {
+	case "ts", "typescript", "":
+		language = "ts"
+	case "js", "javascript":
+		language = "js"
+	default:
+		return fmt.Errorf("unrecognized value %q for --language flag", language)
```
radu-matei, line 63: Would it be helpful to print the expected values for the language flag?

willie-yao: `return fmt.Errorf("unrecognized value %q for --language flag (ts or js expected)", language)`

Good suggestion, how does this look?

v2/cli/init_command.go
```diff
@@ -0,0 +1,248 @@
...
+	"path"
+	"regexp"
+	"strings"
+
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/brigadecore/brigade/v2/internal/file"
+	"github.com/urfave/cli/v2"
+)
+
+type InitFields struct {
```
krancour, line 17: Let's downcase the name of this type so it's not exported. There's probably no harm in it being exported, but as a general rule, keep things _unexported_ until there's a reason not to.

v2/cli/init_command.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+var initCommand = &cli.Command{
+	Name:  "init",
+	Usage: "Initialize a template Brigade project",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:    flagID,
+			Aliases: []string{"i"},
+			Usage: "Name of the Brigade project to create " +
```
krancour, line 30: I would suggest changing this to "Name of the Brigade project to initialize". I believe that will improve clarity on what this command does. As it's currently worded, a user could pretty easily be fooled into thinking that `brig project create` takes place automatically when they do `brig init`

The notes will eventually add further clarity, but this will help.

v2/cli/init_command.go
```diff
@@ -0,0 +1,248 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/brigadecore/brigade/v2/internal/file"
+	"github.com/urfave/cli/v2"
+)
+
+type InitFields struct {
+	ID       string
```
krancour, line 18: I would suggest changing this to `ProjectID` to make it a bit more clear what this ID is for.

v2/cli/init_command.go
```diff
@@ -0,0 +1,248 @@
...
+	Language string
+	Git      string
```
krancour, line 20: Changing this field name to something like `GitCloneURL` would make its purpose more obvious.

v2/cli/init_command.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+var initCommand = &cli.Command{
+	Name:  "init",
+	Usage: "Initialize a template Brigade project",
```
krancour, line 25: Suggest changing to "Initialize a new Brigade project"

v2/cli/init_command.go
```diff
@@ -0,0 +1,248 @@
...
+			Usage: "Name of the Brigade project to create " +
+				"(required)",
+			Required: true,
+		},
+		&cli.StringFlag{
+			Name:    flagLanguage,
+			Aliases: []string{"l"},
+			Usage: "Set the language of the brigade project file" +
+				"to either JavaScript or TypeScript (default)",
+		},
```
krancour, line 39: This can be defaulted to "js" right here and that means a little less work later...

krancour: Also suggest modified usage text:

"Specify the scripting language to use for handling events -- JavaScript (js) or TypeScript (ts; default)"

v2/cli/init_command.go
```diff
@@ -0,0 +1,248 @@
...
+		&cli.StringFlag{
+			Name:    flagGit,
+			Aliases: []string{"g"},
+			Usage:   "Initialize the brigade project from an existing git repo",
```
krancour, line 43: This usage text is a bit misleading. Try something like:

"Optionally specify a git repository where the event handling script can be found"

v2/cli/init_command.go
```diff
@@ -0,0 +1,248 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Detect if .gitignore exists. If it does, add .brigade/secrets.yaml to it
+// If not, create it and add .brigade/secrets.yaml to it
+func addToGitignore(filepath string) error {
```
krancour, line 156: `filepath` may easily be misinterpreted below as the path of the `.gitignore` file. It takes careful reading of the code to realize otherwise. A name like `pathToIgnore` might be less ambiguous.

v2/cli/init_command.go
```diff
@@ -0,0 +1,248 @@
...
+			return err
+		}
+		s := string(b)
+
+		// Only add to gitignore if it's not in the gitignore already
+		if !strings.Contains(s, filepath) {
+			if _, err = f.WriteString("\n" + filepath); err != nil {
+				return err
+			}
+		}
```
krancour, line 179: I think there are probably a lot of weird cases here that weren't anticipated.

If this function were used to ignore "foo" and "foobar" were already in the list, we'd end up _not_ ignoring "foo". Another weird example is that if "foo/" were in the list already, "foo" wouldn't get added and those two things aren't equivalent ("foo/" matches everything under a _directory_ named "foo/").

I think an acceptable route is to unconditionally add the indicated entry and have the comment at the top of the function explicitly note that it's unconditional.

krancour: Which simplifies the logic for this whole function to "create .gitignore if it doesn't exist then, either way, open that file in append mode to add to it."

v2/cli/init_command.go
```diff
@@ -0,0 +1,248 @@
...
+
+func idMatch(id string) error {
+	idMatch, err := regexp.MatchString("^[a-z][a-z\\d-]*[a-z\\d]$", id)
+	if !idMatch || len(id) < 3 || len(id) > 63 {
+		return fmt.Errorf("invalid value %q for --id flag"+
+			" (3-63 characters, first char must be"+
+			" a letter, lowercase letters only)", id)
+	} else if err != nil {
+		return err
+	}
```
krancour, line 202: Few things here. As a general rule, check error _before_ doing anything else. Here you could possibly end up telling the user that the id they supplied was invalid, when the real problem is that the regular expression itself didn't compile.

Look into `regexp.MustCompile`. It's a way to create a regular expression that will panic if it doesn't compile, so anywhere you use it, you can assume the regex itself is error free, so it will make the whole chunk of code above a little neater.

v2/cli/init_command.go
```diff
@@ -0,0 +1,248 @@
...
+	return nil
+}
+
+func normalizeLanguage(language string) (string, error) {
+	switch language {
+	case "ts", "typescript", "":
```
krancour, line 208: I've suggested specifying the default [here](https://github.com/brigadecore/brigade/pull/1437/files#r647448934) so, I'd suggest _not_ accepting `""` here as being indicative of TS. If the default ever were to change (unlikely), we'd only want to change it in one place and the other place is probably the more obvious place where someone would look.

v2/cli/init_command.go
```diff
@@ -0,0 +1,248 @@
...
+		return "", fmt.Errorf("unrecognized value %q for --language flag"+
+			"(ts or js expected)", language)
+	}
+}
+
+// Check if the given file exists, and provide
+// confirmation to overwrite that file
+func checkFileOverwrite(
+	filepath string,
+	filename string,
```
krancour, line 222: This asks the caller to provide redundant information (the filename is in the path, after all). And this function will behave really weirdly if the caller provides a filename that's different from what's on the end of the path...

So just ask them for the path and pick out the filename on your own if you need it.

See https://golang.org/pkg/path/filepath/#Base

v2/cli/init_command_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+func TestNormalizeLanguage(t *testing.T) {
+	testCases := []struct {
+		name     string
+		expected string
+	}{
+		{"typescript", "ts"},
+		{"ts", "ts"},
+		{"javascript", "js"},
+		{"js", "js"},
```
krancour, line 21: I think I noticed this in other places, too. This is a good habit to break. Be explicit about what field receive what values instead of implicitly relying on field order. If someone swapped lines 15 and 16 above, your tests would break.

If your test cases looked like this, that swap wouldn't hurt you:

```
{name: "typescript", expected: "ts"},
{name: "ts", expected: "ts"},
{name: "javascript", expected: "js"},
{name: "js", expected: "js"},
```

Not to mention this is more readable / less ambiguous for someone who comes along later.

v2/cli/init_command_test.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+}
+
+func TestIDMatch(t *testing.T) {
+	testCases := []struct {
+		id string
+	}{
+		{"abc"},
+		{"a11"},
+		{"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"},
```
krancour, line 39: Same comment as [here](https://github.com/brigadecore/brigade/pull/1437/files#r647490516)

v2/cli/init_command_test.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.id, func(t *testing.T) {
+			err := idMatch(testCase.id)
+			require.NoError(t, err)
+		})
+	}
+
+	wrongTestCases := []struct {
```
krancour, line 49: I'd rather see one set of test cases than divide it up into "good cases and bad cases."

I can pair with you later to show you the pattern for incorporating expected results that vary by case.

v2/cli/init_command_test.go
```diff
@@ -0,0 +1,98 @@
...
+	err = ioutil.WriteFile(".gitignore", []byte(``), 0644)
+	require.NoError(t, err)
+	err = addToGitignore(".brigade/secrets.yaml")
+	require.NoError(t, err)
+	verifyContents(t, ".gitignore", ".brigade/secrets.yaml")
+
+	// Remove the created .gitignore file
+	err = os.Remove(".gitignore")
+	require.NoError(t, err)
+}
```
krancour, line 91: I think this test case ends up modifying the project's own files which can end up being problematic in unexpected ways down the road.

We can pair together today on how to test this functionality _without_ adding / modifying / deleting the project's own files.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,59 @@
...
+// nolint: lll
+var newProjectFileBytes = []byte(`# yaml-language-server: $schema=https://raw.githubusercontent.com/brigadecore/brigade/v2/v2/apiserver/schemas/project.json
+apiVersion: brigade.sh/v2-alpha.5
+kind: Project
+metadata:
+  id: {{ .ID }}
+description: Your newly-initialized Brigade project!
+spec:
+  workerTemplate:
+    logLevel: DEBUG
```
krancour, line 18: Let's comment this line. I don't want debug level logging on by default, but I do want people to very clearly see how it's done. So if it's there and commented, they can just uncomment it if they choose.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,59 @@
+package main
+
+import (
+	"bytes"
+	"text/template"
+)
+
+// Template project.yaml file for a non-git project
+// nolint: lll
+var newProjectFileBytes = []byte(`# yaml-language-server: $schema=https://raw.githubusercontent.com/brigadecore/brigade/v2/v2/apiserver/schemas/project.json
```
krancour, line 10: For clarity, I suggest renaming this to something like `projectTemplate`.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,59 @@
...
+    defaultConfigFiles:
+      brigade.{{ .Language }}: |
+        console.log("Hello, World!");
+    {{- else}}
+    git:
+      cloneURL: {{ .Git }}
+    {{- end}}
+`)
+
+var newGitBrigadeFileBytes = []byte(
```
krancour, line 29: For clarity, I suggest renaming this to `scriptTemplate`

v2/cli/init_templates.go
```diff
@@ -0,0 +1,59 @@
...
+`)
+
+var newNotesFileBytes = []byte(
+	`This is a template notes file! Welcome to project {{.ID}}
+`)
+
+var newSecretsFileBytes = []byte(`apiVersion: brigade.sh/v2-alpha.5
+kind: Secret
+key: hello
+value: world
```
krancour, line 40: I'll give you content for this file later today.,

v2/cli/init_templates.go
```diff
@@ -0,0 +1,59 @@
...
+      cloneURL: {{ .Git }}
+    {{- end}}
+`)
+
+var newGitBrigadeFileBytes = []byte(
+	`console.log("Hello, {{.ID}}!");
+`)
+
+var newNotesFileBytes = []byte(
+	`This is a template notes file! Welcome to project {{.ID}}
```
krancour, line 34: I'll give you content for this file later today.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,59 @@
...
+      brigade.{{ .Language }}: |
+        console.log("Hello, World!");
+    {{- else}}
+    git:
+      cloneURL: {{ .Git }}
+    {{- end}}
+`)
+
+var newGitBrigadeFileBytes = []byte(
+	`console.log("Hello, {{.ID}}!");
```
krancour, line 30: I'll give you content for this file later today.

v2/go.mod
```diff
@@ -7,6 +7,7 @@ replace github.com/brigadecore/brigade/sdk/v2 => ../sdk/v2
 require (
 	github.com/AlecAivazis/survey/v2 v2.0.7
 	github.com/Azure/go-amqp v0.13.1
+	github.com/Netflix/go-expect v0.0.0-20180615182759-c93bf25de8e8 // indirect
```
krancour, line 10: Any idea how this new dependency got introduced?

willie-yao: Oops yeah I remember. I was attempting to have a test confirm the prompt automatically but couldn't figure it out. I will delete this import now

v2/cli/init_command.go
```diff
@@ -0,0 +1,241 @@
...
+				"(required)",
+			Required: true,
+		},
+		&cli.StringFlag{
+			Name:    flagLanguage,
+			Aliases: []string{"l"},
+			Usage: "Specify the scripting language to use for handling events" +
+				" -- JavaScript (js) or TypeScript (ts; default)",
+			DefaultText: "ts",
+			Value:       "ts",
```
krancour, line 41: I had to research the `DefaultText` field. I would omit that, please. The help text looks the way it should with just `Value` set, so I don't want to be redundant.

willie-yao: My mistake, I left that in at first because I thought that was the field for setting the default value

v2/cli/init_command.go
```diff
@@ -0,0 +1,243 @@
...
+	// Check if the .brigade/project.yaml path exists.
+	// Confirm overwrite if so.
+	if err = checkFileOverwrite(
+		path.Join(".brigade", "project.yaml"),
+		buf,
+	); err != nil {
+		return err
+	}
+
+	// TODO: Right now the notes doesn't have template fields, add later!
```
krancour, line 119: Is this todo still relevant? The line below looks good to me.

v2/cli/init_command.go
```diff
@@ -0,0 +1,243 @@
...
+
+	// Generate js/ts buffer
+	scriptBuf, scriptErr :=
+		execTemplate(scriptTemplate, fields)
+	if scriptErr != nil {
+		return scriptErr
+	}
+	fields.Script = scriptBuf.String()
+
+	gitFileName := fmt.Sprintf("brigade.%s", fields.Language)
```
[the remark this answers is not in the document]
willie-yao: How does `brigadeScriptFileName` sound?

v2/cli/init_command.go
```diff
@@ -0,0 +1,243 @@
...
+			gitIgnorePath,
+			[]byte(pathToIgnore),
+			0644,
+		); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
krancour, line 189: I think this whole function can be condensed down to just this:

```
file, err := os.OpenFile(".gitignore", os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0600)
if err != nil {
	return err
}
defer file.Close()
if _, err = file.WriteString(fmt.Sprintf("\n%s\n", pathToIgnore)); err != nil {
	return err
}
```

v2/cli/init_command.go
```diff
@@ -0,0 +1,243 @@
...
+	idMatch := regex.MatchString(id)
+	if !idMatch || len(id) < 3 || len(id) > 63 {
+		return fmt.Errorf("invalid value %q for --id flag"+
+			" (3-63 characters, first char must be"+
+			" a letter, lowercase letters only)", id)
+	}
+	return nil
+}
+
+func normalizeLanguage(language string) (string, error) {
```
krancour, line 202: Nit: "canonicalizeLanguage" might more accurately describe what this does.

v2/cli/init_command.go
```diff
@@ -0,0 +1,243 @@
...
+		if !response {
+			return nil
+		}
+	}
+
+	if err := ioutil.WriteFile(filepath, buf.Bytes(), 0644); err != nil {
+		return err
+	}
+	return nil
+}
```
krancour, line 243: I think this function does too much stuff, which makes it difficult to test and even more difficult to reuse bits of it in the future.

Consider having this function only be responsible for determining if the file already exists, prompting for overwrite confirmation if necessary, and either way, returning a boolean that simply tells the caller whether they can proceed with writing (or overwriting, as the case may be) the file.

This way the concerns of validating user intent and actually writing the file are cleanly separated from one another and can be tested independently of one another.

v2/cli/init_command_test.go
```diff
@@ -0,0 +1,132 @@
...
+			result, err := normalizeLanguage(testCase.name)
+			require.NoError(t, err)
+			require.Equal(t, result, testCase.expected)
+		})
+	}
+}
+
+func TestIDMatch(t *testing.T) {
+	testCases := []struct {
+		id         string
```
krancour, line 35: Nit: Mostly for the sake of consistency, use `name` instead of `id` for your test case structs. That's how it's been done throughout the code base and consistency is best when possible.

krancour: Ah. Nevermind. I see now that this is meant to be the `id` of a project. That's fine.

But, let's also add a `name` field that describes what's being tested.

By example, `id` might be `aa`, but `name` might be `id is too short`.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,130 @@
...
+      brigade.{{ .Language }}: |
+        {{ nindent 8 .Script }}
+    {{- else }}
+    git:
+      cloneURL: {{ .GitCloneURL }}
+    {{- end }}
+`)
+
+// nolint: lll
+var scriptTemplate = []byte(`{{- if eq .Language "js" }}const { events } = require("@brigadecore/brigadier");
```
krancour, line 36: Can this be broken into two lines? That would be more readable.

But I know templates are weird. There could be reasons...

v2/cli/init_templates.go
```diff
@@ -0,0 +1,130 @@
...
+
+events.on("brigade.sh/cli", "exec", async event => {
+	let job = new Job("hello", "debian:latest", event);
+	job.primaryContainer.command = ["echo"];
+	job.primaryContainer.arguments = ["Hello, World!"];
+	await job.run();
+});
+
+events.process(); 
+{{- else }}import { events, Job } from "@brigadecore/brigadier"
```
krancour, line 46: Another spot to break up the line if possible.

v2/cli/init_command.go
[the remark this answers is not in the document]
krancour: Sure. Or even just scriptFilename would be well understood.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,130 @@
...
+##
+# foo: bar
+# bat: baz
+`)
+
+// Creates content from a given template and struct
+func execTemplate(
+	templateBytes []byte,
+	fields interface{},
+) (bytes.Buffer, error) {
```
krancour, line 116: Nit: There might be circumstances where returning (and subsequently passing around) a buffer makes sense for reasons of memory management, but I can't find a good reason to do it here. Returning `[]byte` instead of a `bytes.Buffer` might be more straightforward.

v2/cli/init_command.go
willie-yao: Pushed a change that makes this function return a boolean for recording the user response, and moved the logic for writing the file out of the function. Let me know if that's the right idea!


	if overwrite, err := checkFileOverwrite(
		path.Join(".brigade", "NOTES.txt"),
		notesBuf,
	); err != nil {
		return err
	} else if overwrite {
		if err := ioutil.WriteFile(
			path.Join(
				".brigade", "NOTES.txt"),
			notesBuf,
			0644,
		); err != nil {
			return err
		}
	}

v2/cli/init_command.go
```diff
@@ -0,0 +1,258 @@
...
+		fmt.Println("Creating .brigade directory...")
+		err = os.Mkdir(".brigade", 0755)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Generate js/ts buffer
+	scriptBuf, scriptErr :=
+		execTemplate(scriptTemplate, fields)
```
krancour, line 85: I like that you're returning a byte array now. It's more straightforward. In light of this, the variable you assign that result to probably should no longer have "buf" in its name. Something like `scriptBytes` would make sense. This applies to other invocations of `execTemplate()` as well.

krancour: It's out of the ordinary to not use/reuse simply `err` for errors. I suspect you chose this alternative name because the linter gave you grief about a "shadow declaration?" If so, I'll show you how to resolve that more cleanly.

v2/cli/init_templates.go
krancour: You can also try a `-` before the closing curlies to see if that helps.

v2/cli/init_command.go
willie-yao: Yup exactly, I had to change the name of the error because of that reason!

v2/cli/init_command_test.go
krancour: Well... you just taught me something cause I've used tempdir for this stuff often and never realized T had any built in support for that. Neat.

v2/cli/init_command.go
krancour: This is, indeed the right idea. Try to do the path. Join stuff just once tho... save it in a variable. It's always good to write things in a way that if you have to change something someday, it's done in just one place.

v2/cli/init_command.go
krancour: The short an answer to explicitly declare the _other_ variable separately above so that err can be reused without redeclaring (i.e. without the colon in the assignment). This stops you from making a new err variable with the same name as the old one, which has the potential to generate confusion-- which is why the linter complains.

v2/cli/init_templates.go
willie-yao: That fixed it, thanks!

v2/cli/init_command.go
```diff
@@ -0,0 +1,255 @@
...
+		return err
+	} else if !dirExists {
+		fmt.Println("Creating " + dirPath + " directory...")
+		err = os.Mkdir(dirPath, 0755)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
krancour, line 188: Please rename this to `ensureDirectory`, as that succinctly and accurately denotes what it does-- ensures the named directory exists, either because it already did or because you are creating it here and now.

Also, please move this to the `file` package since I anticipate this could one day be useful to other parts of Brigade than just the CLI.

v2/cli/init_command.go
```diff
@@ -0,0 +1,255 @@
...
+func idMatch(id string) error {
+	regex := regexp.MustCompile(`^[a-z][a-z\d-]*[a-z\d]$`)
+	idMatch := regex.MatchString(id)
+	if !idMatch || len(id) < 3 || len(id) > 63 {
+		return fmt.Errorf("invalid value %q for --id flag"+
+			" (3-63 characters, first char must be"+
+			" a letter, lowercase letters only)", id)
+	}
+	return nil
+}
```
krancour, line 216: This has, weirdly, been the first time we've cared about validating a project ID client side, but it's unlikely to be the last. Let's move this to a function named `ValidateProjectID` in the `sdk/v2/core` package.

v2/cli/init_command.go
```diff
@@ -0,0 +1,255 @@
...
+				scriptPath,
+				scriptBytes,
+				0644,
+			); err != nil {
+				return err
+			}
+		}
+	}
+
+	// Script is added to project.yaml otherwise
```
krancour, line 106: This comment is misleading because it talks about something that happens conditionally, but there's no obvious conditional logic here. I _do_ happen to know that the conditional is inside the template, but don't assume someone new who comes along and reads this code will posses that degree of knowledge or context.

It think, in this case, it's clearer to just remove this comment.

krancour: Or rewrite the comment to just say something like "Generate the project.yaml file," without mention of conditions that are actually buried within the template.

v2/cli/init_command.go
```diff
@@ -0,0 +1,255 @@
...
+	if err != nil {
+		return err
+	}
+	fields.Script = string(scriptBytes)
+
+	brigadeScriptFileName := fmt.Sprintf("brigade.%s", fields.Language)
+
+	// If git repo is provided, create a separate js/ts file for script
+	var overwrite bool
+	scriptPath := path.Join(".brigade", brigadeScriptFileName)
```
krancour, line 91: As a general rule, try to give most variables the narrowest scope possible. In this case, these can be moved into the condition block that follows.

v2/cli/init_command.go
```diff
@@ -0,0 +1,255 @@
...
+		}
+	}
+	return nil
+}
+
+// Detect if .gitignore exists. If it does, add .brigade/secrets.yaml to it
+// If not, create it and add .brigade/secrets.yaml to it unconditionally
+func addToGitignore(pathToIgnore string, gitIgnorePath string) error {
+	file, err :=
+		os.OpenFile(gitIgnorePath, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0600)
```
krancour, line 194: 600 seems unnecessarily restrictive. Let's go with 644, please.

v2/cli/init_command.go
willie-yao: Would I be moving this into the projects.go file? That would make the most sense, but it seems like most functions in that file are related to a ProjectsClient

v2/cli/init_command.go
willie-yao: `overwrite` is used again on line 116 in another conditional, so would it make more sense to leave the declaration of `overwrite` alone?

v2/internal/file/file.go
```diff
@@ -14,3 +17,17 @@ func Exists(filename string) (bool, error) {
 	}
 	return false, err
 }
+
+// Check if the dirPath directory exists. If not, create it
+func EnsureDirectory(dirPath string) error {
```
krancour, line 22: I'm surprised the linter didn't flag this. The comment above an exported function needs to start with the function's name. i.e. "// EnsureDirectory does...."

v2/internal/file/file_test.go
```diff
@@ -17,3 +18,18 @@ func TestExists(t *testing.T) {
 	require.NoError(t, err)
 	require.False(t, exists)
 }
+
+func TestCheckCreateDirectory(t *testing.T) {
```
krancour, line 22: Name should change to `TestEnsureDirectory`.

v2/cli/init_command.go
krancour: Sure. But move the other one.

v2/cli/init_command.go
krancour: `projects.go` sounds like a good place. That file is all things project-related and not limited to the API client.

sdk/v2/core/projects.go
```diff
@@ -85,6 +86,18 @@ func (p ProjectList) MarshalJSON() ([]byte, error) {
...
+	regex := regexp.MustCompile(`^[a-z][a-z\d-]*[a-z\d]$`)
+	idMatch := regex.MatchString(id)
+	if !idMatch || len(id) < 3 || len(id) > 63 {
+		return fmt.Errorf("invalid value %q for --id flag"+
+			" (3-63 characters, first char must be"+
+			" a letter, lowercase letters only)", id)
+	}
+	return nil
+}
+
```
krancour, line 100: Let's move this down to the bottom of the file, please, so it's set apart from API-related stuff.

sdk/v2/core/projects_test.go
```diff
@@ -23,6 +23,81 @@ func TestProjectListMarshalJSON(t *testing.T) {
 	metaTesting.RequireAPIVersionAndType(t, ProjectList{}, "ProjectList")
 }
 
+func TestValidateProjectID(t *testing.T) {
```
krancour, line 26: Same as [this](https://github.com/brigadecore/brigade/pull/1437/files#r660822823)

sdk/v2/core/projects_test.go
```diff
@@ -23,6 +23,81 @@ func TestProjectListMarshalJSON(t *testing.T) {
...
+	testCases := []struct {
+		name       string
+		id         string
+		assertions func(err error)
```
krancour, line 30: This is a common pattern that I like and use a lot, but we can do things a little more simply here. Instead of having each test case furnish a function that can make assertions, just have a bool and each test case can say whether the example is valid or not.

v2/cli/init_command.go
```diff
@@ -0,0 +1,243 @@
...
+			Name:    flagGit,
+			Aliases: []string{"g"},
+			Usage: "Optionally specify a git repository where the" +
+				" event handling script can be found",
+		},
+	},
+	Action: initialize,
+}
+
+// Handle init CLI flags
```
krancour, line 51: I think I mentioned that for exported functions, the comment above should start with the name of the function. While that's not enforced for unexported functions, it's still a good thing to adhere to-- that way comments don't have to be significantly re-written if something that was once unexported becomes exported in the future. In this case, it would also be fine to just drop the comment.

v2/cli/init_command.go
```diff
@@ -0,0 +1,243 @@
...
+// 		return err
+// 	} else if !dirExists {
+// 		fmt.Println("Creating " + dirPath + " directory...")
+// 		err = os.Mkdir(dirPath, 0755)
+// 		if err != nil {
+// 			return err
+// 		}
+// 	}
+// 	return nil
+// }
```
krancour, line 188: Let's remove this commented code.

willie-yao: Wow I can't believe I forgot to remove that after testing, I'll be more careful next time!

v2/cli/init_command.go
```diff
@@ -0,0 +1,243 @@
...
+	"github.com/brigadecore/brigade/v2/internal/file"
+	"github.com/urfave/cli/v2"
+)
+
+type initFields struct {
+	ProjectID   string
+	Language    string
+	GitCloneURL string
+	Script      string
+}
```
krancour, line 21: I usually recommend against creating types that are only used in one place because it "clutters" things.

Also, not every template that you render needs all of this information.

I would suggest using an anonymous struct type having only relevant fields anywhere you need to pass stuff to `execTemplate`.

v2/cli/init_command.go
krancour: No worries. Fixed easily enough.

v2/cli/init_command.go
willie-yao: Sounds good. If so, would it be good to just save each value to a local variable and pass them into each execTemplate as needed?

willie-yao: For now, I just have fields saved as an anonymous struct, and pass them into execTemplate in the same way. Although there are unnecessary fields for some templates, I feel like it might be too cluttered to declare an entire different anonymous struct for each execTemplate. I could be wrong in my thinking though

krancour: > For now, I just have fields saved as an anonymous struct, and pass them into execTemplate in the same way

> I feel like it might be too cluttered to declare an entire different anonymous struct for each execTemplate

This sounds like a reasonable compromise.

v2/cli/init_command.go
```diff
@@ -187,7 +188,8 @@ func addToGitignore(pathToIgnore string, gitIgnorePath string) error {
 	return nil
 }
 
-// Given a language, check if it is an allowed language and canonicalize it
+// canonicalizeLanguage checks if a given lagnuage is allowed. If it is,
+// shorten it to its file extension. Otherwise, return an error.
 func canonicalizeLanguage(language string) (string, error) {
 	switch language {
```
krancour, line 194: Should probably convert the input to lower case first. As it stands now, there's an unwritten assumption that the caller has.

v2/cli/init_command.go
```diff
@@ -0,0 +1,230 @@
...
+		if err = ioutil.WriteFile(
+			secretsPath,
+			secretsBytes,
+			0644,
+		); err != nil {
+			return err
+		}
+	}
+
+	err = addToGitignore(".brigade/secrets.yaml", "./.gitignore")
```
krancour, line 166: There's a second path we should add here. It's `.brigade/node_modules/`

v2/cli/init_command.go
```diff
@@ -0,0 +1,230 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// addToGitIgnore detects if .gitignore exists. If it does, add
+// .brigade/secrets.yaml to it. If not, create it and add .brigade/secrets.yaml
+// to it unconditionally
+func addToGitignore(pathToIgnore string, gitIgnorePath string) error {
```
krancour, line 177: I want to suggest making this a [variadic function](https://gobyexample.com/variadic-functions)

This will allow multiple paths to be specified for addition to `.gitignore`. You can loop over those to ignore them all.

The advantage of this approach is that it's more efficient than what you'd have to do currently to ignore multiple paths-- i.e. it's more efficient than calling `addToGitignore(...)` multiple times for each path, which would unnecessarily result in checking for the existence of the file, opening, and closing it every time. It would be better to do all of that just once.

krancour: Related to [this comment](https://github.com/brigadecore/brigade/pull/1437/files#r661587236), btw.

v2/cli/init_command.go
willie-yao: I don't actually create a node_modules folder, so are we just assuming the user will install some node dependencies with their project? That should be a safe assumption to make

sdk/v2/core/projects.go
```diff
@@ -340,3 +341,15 @@ func (p *projectsClient) Authz() AuthzClient {
...
 }
+
+// ValidateProjectID checks if a given id is valid.
+func ValidateProjectID(id string) error {
+	regex := regexp.MustCompile(`^[a-z][a-z\d-]*[a-z\d]$`)
+	idMatch := regex.MatchString(id)
+	if !idMatch || len(id) < 3 || len(id) > 63 {
+		return fmt.Errorf("invalid value %q for --id flag"+
+			" (3-63 characters, first char must be"+
+			" a letter, lowercase letters only)", id)
```
krancour, line 352: Since this has now moved out of the CLI and into the SDK, it could be used in any number of ways that have nothing to do with the CLI, so this error message probably needs to be more generic now-- i.e. no mention of a specific flag.

sdk/v2/core/projects_test.go
```diff
@@ -254,3 +254,62 @@ func TestProjectsClientDelete(t *testing.T) {
...
+		{
+			name:     "id is too short by 2 characters",
+			id:       "a",
+			expected: false,
+		},
+		{
+			name:     "id is too short by 1 character",
+			id:       "aa",
+			expected: false,
+		},
```
krancour, line 290: These can be collapsed into a single case. "id is too short"

v2/cli/init_command.go
```diff
@@ -0,0 +1,256 @@
...
+	fields := struct {
+		ProjectID   string
+		Language    string
+		GitCloneURL string
+		Script      string
+	}{
+		c.String(flagID),
+		strings.ToLower(c.String(flagLanguage)),
+		c.String(flagGit),
+		"",
```
krancour, line 57: You should always include keys when initializing fields in a struct type. All it would take is for the order of some fields in lines 49 - 52 to be changed and we'd be in deep trouble. If we're in the habit of always using keys, that same change would have no impact.

v2/cli/init_command.go
```diff
@@ -0,0 +1,256 @@
...
+		return err
+	}
+
+	// Check if the .brigade directory exists
+	err = file.EnsureDirectory(".brigade")
+	if err != nil {
+		return err
+	}
+
+	// Generate js/ts buffer
```
krancour, line 77: Nit: Change the comment to "Generate script". iirc "buffer" is an artifact of one of the previous states of this PR and is no longer relevant.

v2/cli/init_command.go
```diff
@@ -0,0 +1,256 @@
...
+
+	brigadeScriptFileName := fmt.Sprintf("brigade.%s", fields.Language)
+
+	// If git repo is provided, create a separate js/ts file for script
+	var overwrite bool
+	if fields.GitCloneURL != "" {
+		scriptPath := path.Join(".brigade", brigadeScriptFileName)
+		if overwrite, err = checkFileOverwrite(scriptPath); err != nil {
+			return err
+		} else if overwrite {
```
krancour, line 93: You don't need the else here. If there was an error, we've returned already. The only way we get to here is if there was no error.

v2/cli/init_command.go
```diff
@@ -0,0 +1,256 @@
...
+		var packageBytes []byte
+		packageBytes, err = execTemplate(packageTemplate, fields)
+		if err != nil {
+			return err
+		}
+
+		packagePath := path.Join(".brigade", "package.json")
+		if overwrite, err = checkFileOverwrite(packagePath); err != nil {
+			return err
+		} else if overwrite {
```
krancour, line 113: Another case where you can skip the else.

v2/cli/init_command.go
```diff
@@ -0,0 +1,256 @@
...
+			secretsPath,
+			secretsBytes,
+			0644,
+		); err != nil {
+			return err
+		}
+	}
+
+	err = addToGitignore(
+		"./.gitignore",
```
krancour, line 187: This tripped me up initially. At a quick glance, it looks so much like `./.gitignore` is being added to `.gitignore`.

I know that's not what's actually happening. I know you've parameterized the file path _to make testing easier and that is a great move._

What can be done to make this easier to understand, though, is this:

Create _more_ generic function like `addLinesToFile(...)`. That will be the thing you can easily test.

`addToGitIgnore(...)` won't be tested, and it can invoke the well-tested `addLinesToFile(...)` function using the hard-coded `./.gitignore` path.

The end result is that when someone in the future looks at _this_ bit of code right here, they'll see:

```
err = addToGitIgnore(".brigade/secrets.yaml", ".brigade/node_modules/")
```

And that will be less likely to confuse them. They won't stop to ask themselves why `.gitignore` is being added to `.gitignore`.

v2/cli/init_command.go
```diff
@@ -0,0 +1,256 @@
...
+		".brigade/secrets.yaml",
+		".brigade/node_modules",
+	)
+	if err != nil {
```
krancour, line 191: Can move this conditional up to line 186

v2/cli/init_command.go
```diff
@@ -0,0 +1,256 @@
...
+			0644,
+		); err != nil {
+			return err
+		}
+	}
+
+	err = addToGitignore(
+		"./.gitignore",
+		".brigade/secrets.yaml",
+		".brigade/node_modules",
```
krancour, line 189: I'm fuzzy on this, but I am almost positive that the trailing `/` needs to be present to ignore the entire contents of a directory (which is what we're doing here).

Even if it turns out not to be strictly required, it is at least a frequently applied convention.

v2/cli/init_command.go
```diff
@@ -0,0 +1,256 @@
...
+			file.WriteString(fmt.Sprintf("\n%s", path)); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// canonicalizeLanguage checks if a given lagnuage is allowed. If it is,
+// shorten it to its file extension. Otherwise, return an error.
+func canonicalizeLanguage(language string) (string, error) {
```
krancour, line 219: I know I suggested the name `canonicalizeLanguage`. After understanding this PR much better, what we're not merely converting non-canonical input to something canonical, what we're _really_ doing here is converting input (like ts / typescript / js / javascript) to a file extension.

Suggesting another rename for clarity: `fileExtensionForLanguage(...)`

v2/cli/init_command.go
willie-yao: There are actually three more cases of this same thing happening for three other files in `initialize`, so I'll go ahead and apply the same changes to those

v2/cli/init_command.go
willie-yao: Should `addLinesToFile` be moved to the `file` package? Since it's now no longer specific to `init`

willie-yao:
Quick question: if a user doesn't specify a git url, should we still generate a `package.json`? The embedded script files still have brigadier listed as a dependency, so I might need to change `package.json` to always be generated rather than only when the `-g` flag is specified.

v2/cli/init_command.go
krancour: That's a good question. imho, that's not a common enough operation that we should bother at this point. We can later if it crops up again.

krancour:
> Quick question: if a user doesn't specify a git url, should we still generate a package.json? The embedded script files still have brigadier listed as a dependency, so I might need to change package.json to always be generated rather than only when the -g flag is specified.

That's a good question. I had already toyed with the idea that we should embed a package.json and I ended up opting against it.

Here's the rationale:

Brigade never _requires_ you to have a package.json if the only library you use happens to be brigadier. The worker is smart enough to generate a minimal package.json for you if none exists and adds brigadier (or more accurately, a polyfill-- alternative implementations of the brigadier interfaces) to it.

The only real reasons to _have_ a package.json are if you have additional dependencies you need to spell out _or_ because you want context help in your editor...

If you're writing a brigade.js or brigade.ts file, you're likely to want that context help.

If your brigade.js or brigade.ts is embedded in a yaml file, all bets are off. No editor's going to give you context help for that chunk of js/ts that you embed. So this takes us back to the only reason you'd do this being that you have additional dependencies.

I'm making a judgement call that it's probably rare to begin with that someone is going to embed their script in their project file and that it will be rarer still that such a script isn't pretty trivial.

So... bottom line... don't add package.json to the project.yaml file.

vdice (approved):
LGTM!  A few minor typos to fix before merging and then some potential follow-up(s) to consider.

Awesome test coverage!

v2/cli/init_templates.go
```diff
@@ -0,0 +1,141 @@
...
+	"bytes"
+	"text/template"
+
+	"github.com/Masterminds/sprig"
+)
+
+// Template project.yaml file for a non-git project
+// nolint: lll
+var projectTemplate = []byte(`# yaml-language-server: $schema=https://raw.githubusercontent.com/brigadecore/brigade/v2/v2/apiserver/schemas/project.json
+apiVersion: brigade.sh/v2-alpha.5
```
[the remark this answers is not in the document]
krancour: I think fighting with versions is a short term problem. imho, we shouldn't worry about it.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,141 @@
...
+## Example keys/values:
+##
+# foo: bar
+# bat: baz
+`)
+
+var packageTemplate = []byte(`{
+  "name": "{{ .ProjectID }}",
+  "dependencies": {
+    "@brigadecore/brigadier": "^2.0.0-alpha.5"
```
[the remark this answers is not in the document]
willie-yao: This is a good suggestion. Since @krancour mentioned it's a short term problem, I think we can leave it as a follow-up.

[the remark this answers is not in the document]
vdice: 👍 sgtm!

sdk/v2/core/projects.go
```diff
@@ -340,3 +341,15 @@ func (p *projectsClient) Authz() AuthzClient {
 func (p *projectsClient) Secrets() SecretsClient {
 	return p.secretsClient
 }
+
+// ValidateProjectID checks if a given id is valid.
+func ValidateProjectID(id string) error {
+	regex := regexp.MustCompile(`^[a-z][a-z\d-]*[a-z\d]$`)
```
krancour, line 347: This regex will get recompiled every time this function is called. It would be better to move this to the package level where it will be compiled once when the package is initialized. That was there's less overhead if and when this function is called repeatedly.

sdk/v2/core/projects_test.go
```diff
@@ -254,3 +254,57 @@ func TestProjectsClientDelete(t *testing.T) {
...
+				" allowed characters and numbers",
+			id:       "a11",
+			expected: true,
+		},
+		{
+			name: "Valid id with maximum amount of allowed characters",
+			id: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" +
+				"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+			expected: true,
+		},
```
krancour, line 280: Let's collapse these into just once success case. It's interesting and useful to test lots of failure cases-- make validation fail for different reasons-- but it's less interesting to test different success cases. There's _really_ only one success case-- you succeed when you don't fail.

v2/cli/init_command_test.go
```diff
@@ -0,0 +1,65 @@
...
+	// Verify .gitignore contains .brigade/secrets.yaml
+	verifyContents(t, gitIgnorePath, ".brigade/secrets.yaml")
+
+	// Clear .gitignore, test functionality of writing to existing .gitignore
+	err = ioutil.WriteFile(gitIgnorePath, []byte(``), 0644)
+	require.NoError(t, err)
+	err = addLinesToFile(gitIgnorePath, ".brigade/secrets.yaml")
+	require.NoError(t, err)
+	verifyContents(t, gitIgnorePath, ".brigade/secrets.yaml")
+}
```
krancour, line 58: Since the `addLinesToFile` is generic now, let's make the test case also less `.gitignore`-specific.

v2/cli/init_command_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+func verifyContents(t *testing.T, filename string, content string) {
+	b, err := ioutil.ReadFile(filename)
+	require.NoError(t, err)
+	s := string(b)
+	require.True(t, strings.Contains(s, content))
+}
```
krancour, line 65: Nit: Rename to `verifyFileContents` just because it makes things more clear.

v2/cli/init_command_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	// Clear .gitignore, test functionality of writing to existing .gitignore
+	err = ioutil.WriteFile(gitIgnorePath, []byte(``), 0644)
+	require.NoError(t, err)
+	err = addLinesToFile(gitIgnorePath, ".brigade/secrets.yaml")
+	require.NoError(t, err)
+	verifyContents(t, gitIgnorePath, ".brigade/secrets.yaml")
+}
+
+func verifyContents(t *testing.T, filename string, content string) {
```
krancour, line 60: Nit: Suggest changing `filename` to `filePath` because `filename` is often misinterpreted as being only what follows the last path delimiter (e.g. `/`). If this arg is called `filePath` instead, there's less room for misinterpretation.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,141 @@
+package main
+
+import (
+	"bytes"
+	"text/template"
+
+	"github.com/Masterminds/sprig"
+)
+
+// Template project.yaml file for a non-git project
```
krancour, line 10: This comment isn't accurate. This template is used for projects that involve git as well as the ones that don't.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,141 @@
...
+apiVersion: brigade.sh/v2-alpha.5
+kind: Project
+metadata:
+  id: {{ .ProjectID }}
+description: My new Brigade project
+spec:
+  eventSubscriptions:
+  - source: brigade.sh/cli
+    types:
+    - exec
```
krancour, line 22: Let's comment this section and treat it as an example. Let's also add more comments above to explain the example.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,141 @@
...
+events.process(); 
+{{- else -}}
+import { events, Job } from "@brigadecore/brigadier"
+
+events.on("brigade.sh/cli", "exec", async event => {
+	let job = new Job("hello", "debian:latest", event)
+	job.primaryContainer.command = ["echo"]
+	job.primaryContainer.arguments = ["Hello, World!"]
+	await job.run()
+})
```
krancour, line 55: Let's split this into two separate templates and decide which template to use in the code that renders a script from the template.

This isn't terrible as is, but I'm aware of the probability of adding support for a third "script" style (declarative YAML) that will likely land in 2.1. Based on that, I'd say _one_ script template like we have now is going to become unwieldy quickly if the template keeps growing with additional conditional logic.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,141 @@
...
+
+events.process()
+{{- end }}
+`)
+
+// nolint: lll
+var notesTemplate = []byte(`This file was created by brig init and outlines next steps.
+
+1. Edit project.yaml as you see fit and submit it to Brigade using:
+     brig project create -f .brigade/project.yaml
```
krancour, line 65: Let's add a lint break before this. It makes it a bit easier to read.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,141 @@
...
+
+2. If your project needs to make use of any secret/sensitive values, set them
+   one at a time using:
+     brig project secrets set --id {{ .ProjectID }} --set <key>=<value>
```
krancour, line 69: Line break.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,141 @@
...
+
+   Or set secrets in bulk by modifying secrets.yaml, then submit them to
+   Brigade using:
+     brig project secrets set --id {{ .ProjectID }} --file .brigade/secrets.yaml
```
krancour, line 73: Line break.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,141 @@
...
+
+   Secrets become available within your brigade.{{ .Language }} script using:
+     event.project.secrets.<key>
```
krancour, line 76: Line break.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,141 @@
...
+    location.
+{{- end }}
+`)
+
+var secretsTemplate = []byte(`## This file was created by brig init.
+##
+## Specify values for project secrets using key/value pairs below.
+##
+## You can load all these secrets into your project using
+## brig project secrets set --id {{ .ProjectID }} --file .brigade/secrets.yaml
```
krancour, line 103: Let's add a line break and indent to make this more readable.

v2/cli/init_templates_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+func TestProjectFileTemplate(t *testing.T) {
+	testCases := []struct {
+		name        string
+		projectID   string
+		language    string
+		gitCloneURL string
+	}{
+		{"nonGit", "hello", "ts", ""},
+		{"git", "hello", "ts",
```
krancour, line 30: Reminder not to initialize things without being explicit about what values go in which fields. That will be more resilient to change.

v2/cli/init_templates_test.go
```diff
@@ -0,0 +1,71 @@
...
+				execTemplate(projectTemplate, struct {
+					ProjectID   string
+					Language    string
+					GitCloneURL string
+					Script      string
+				}{
+					testCase.projectID,
+					testCase.language,
+					testCase.gitCloneURL,
+					"",
```
krancour, line 46: Reminder not to initialize things without being explicit about what values go in which fields. That will be more resilient to change.

v2/cli/init_templates.go
willie-yao: I might be misunderstanding your suggestion. Do you mean I should leave this section commented out like so? Also I'm not sure if my description is accurate.
```
spec:
  # This section defines any arbitrary event for your project to listen
  # and respond to.
  # eventSubscriptions:
  # - source: brigade.sh/cli
  #   types:
  #   - exec
```

krancour: Try this:

```
## Subscribe below to any events that should trigger your script.
## The commented example depicts a subscription to "exec"
## events originating from the Brigade CLI. 
# eventSubscriptions:
# - source: brigade.sh/cli
#   types:
#   - exec
```

Note that the double # are deliberate to distinguish between a comment that is _commentary_ and the comments below it which are commented YAML. This is a widely used convention in the Kubernetes community.

sdk/v2/core/projects.go
```diff
@@ -14,6 +15,8 @@ import (
 // ProjectKind represents the canonical Project kind string
 const ProjectKind = "Project"
 
+var regex = regexp.MustCompile(`^[a-z][a-z\d-]*[a-z\d]$`)
```
krancour, line 18: Can we change the name of this so it's more obvious what it's for?

v2/cli/init_templates.go
```diff
@@ -0,0 +1,149 @@
...
+    {{- else }}
+    git:
+      cloneURL: {{ .GitCloneURL }}
+    {{- end }}
+`)
+
+// nolint: lll
+var typeScriptTemplate = []byte(`import { events, Job } from "@brigadecore/brigadier"
+
+events.on("brigade.sh/cli", "exec", async event => {
```
krancour, line 41: Another case where I think a comment could be instructive.  Let's try something like:

```
// Use events.on() to define how your script responds to different events. The examples below depicts
// handling of "exec" events originating from the Brigade CLI.
```

The handler itself can also be commented.

The JS template will require the same treatment.

v2/cli/init_command.go
```diff
@@ -0,0 +1,274 @@
...
+		return err
+	}
+
+	// Check if input language is valid
+	fields.Language, err = fileExtensionForLanguage(fields.Language)
+	if err != nil {
+		return err
+	}
+
+	// Check if the .brigade directory exists
```
krancour, line 71: Change "check" to "ensure".

v2/cli/init_command.go
```diff
@@ -0,0 +1,274 @@
...
+	// Generate script based on given language
+	var scriptBytes []byte
+	switch fields.Language {
+	case "ts":
+		scriptBytes, err = execTemplate(typeScriptTemplate, fields)
+	case "js":
+		scriptBytes, err = execTemplate(javaScriptTemplate, fields)
+	default:
+		return fmt.Errorf("unrecognized value %q for --language flag"+
+			"(ts or js expected)", fields.Language)
```
krancour, line 86: Wouldn't we already have failed if the language specified by the user didn't map to js or ts?

I think this default case cannot occur.

willie-yao: You're right, I added this default case because I thought that it was convention to always have a default case, but it's unnecessary in this case

v2/cli/init_templates.go
willie-yao: I added a few more comments in the handler itself. Do these make sense and does the wording look okay?

```
// Use events.on() to define how your script responds to different events. 
// The examples below depicts handling of "exec" events originating from
// the Brigade CLI.
  // Use the Job() constructor to define a new job for the event.
  let job = new Job("hello", "debian:latest", event)

  // Define the commands and arguments for the job.
  job.primaryContainer.command = ["echo"]
  job.primaryContainer.arguments = ["Hello, World!"]

  // Use job.run() to run the job you have just defined.
  await job.run()
})
```

willie-yao: Oh wait, do you mean that I should add comments to the handler, or comment it out?

willie-yao: I'm assuming that I'm supposed to comment it out, since the exec event in `project.yaml` is commented out. I also only used one set of `//` for the comment, as I'm not sure if that's the standard for script files in the same way as yaml files

krancour: Ya... comment _out_ the handler.

krancour:
Up until now I've only reviewed the code. Now I've built this and have been getting hands on.

When I used this feature, I see output like this:

```
Creating .brigade directory...
```

That's useful, but I'd really like to see more... I'd like to see some output for every file that gets created or overwritten.

krancour:
I also seem to be getting an extra blank line at the start of the script if a script is embedded into the project.yaml file.

krancour:
Also, if we add to brigade.js, can you add a blank line to the end? It's actually fairly conventional that all files end with an empty line.

willie-yao:
> Also, if we add to brigade.js, can you add a blank line to the end?

The template already has a blank line at the end of the file for both js and ts. Is it different on your end?

krancour:
> The template already has a blank line at the end of the file for both js and ts. Is it different on your end?

I'm sorry. I misspoke.

I meant `.gitignore`

v2/cli/init_command.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+		if overwrite {
+			if err = ioutil.WriteFile(
+				scriptPath,
+				scriptBytes,
+				0644,
+			); err != nil {
+				return err
+			}
+			fmt.Println("Creating " + brigadeScriptFileName + "...")
```
krancour, line 107: Little bit of an issue here. This is going to say "creating" even if what we're doing is overwriting. We really should differentiate between those to things. This comment is applicable in several other spots as well.

v2/internal/file/file.go
```diff
@@ -14,3 +17,18 @@ func Exists(filename string) (bool, error) {
...
 	return false, err
 }
+
+// EnsureDirectory checks if the dirPath directory exists. If not, it
+// creates it.
+func EnsureDirectory(dirPath string) error {
+	if dirExists, err := Exists(dirPath); err != nil {
+		return err
+	} else if !dirExists {
+		fmt.Println("Creating " + dirPath + " directory...")
```
krancour, line 27: Sorry for not catching this one sooner.

As a general rule, we want to keep output like this _out_ of functions that have the potential for reuse. Supposing for instance that we ended up reusing this function somewhere server side-- this isn't something I want/need to be seeing show up in the server's logs.

What you might consider doing is returning a `bool` that indicates whether the directory was created (`true`) or not (`false`, since it already existed). Then the caller can act on that information and display such a message if it's contextually appropriate to do so.

v2/cli/init_command.go
willie-yao: I had to throw in the print statements directly inside `checkFileOverwrite` due to the logic I used for checking for whether to write the file or not. Is that okay since `checkFileOverwrite` isn't being exported?

willie-yao: I understand that your next comment addresses this problem a bit. Would it be better to return two booleans? One for indicating whether a file is being written at all, and whether it is being overwritten or created

v2/cli/init_command.go
```diff
@@ -95,19 +95,26 @@ func initialize(c *cli.Context) error {
...
+		if fileWrite {
 			if err = ioutil.WriteFile(
 				scriptPath,
 				scriptBytes,
 				0644,
 			); err != nil {
 				return err
 			}
+			if overwrite {
+				fmt.Println("Overwriting " + brigadeScriptFileName + "...")
```
krancour, line 114: Can we use `scriptPath` here instead of `brigadeScriptFileName`? Same comment applies on line 116.

v2/cli/init_command.go
```diff
@@ -118,17 +125,23 @@ func initialize(c *cli.Context) error {
...
+		if fileWrite {
 			if err = ioutil.WriteFile(
 				packagePath,
 				packageBytes,
 				0644,
 			); err != nil {
 				return err
 			}
+			if overwrite {
+				fmt.Println("Overwriting .brigade/package.json...")
```
krancour, line 141: Can we use `packagePath` instead of writing out the path again here? (Especially since forward slash may be inaccurate on a Windows system.) Same comment applies numerous other places.

v2/cli/init_command.go
krancour: Also, let's opt for `fmt.Printf("Overwriting %s...\n", scriptPath)` over using `+` for string concat. Also applies in similar spots.

krancour:
@willie-yao I'm pretty happy with this code.

I'm going to build from this branch and get hands on with it one more time to check things out before LGTM'ing.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,157 @@
...
+    git:
+      cloneURL: {{ .GitCloneURL }}
+    {{- end }}
+`)
+
+// nolint: lll
+var typeScriptTemplate = []byte(`import { events, Job } from "@brigadecore/brigadier"
+
+// Use events.on() to define how your script responds to different events. 
+// The examples below depicts handling of "exec" events originating from
```
krancour, line 42: Change "examples" to "commented example".

v2/cli/init_templates.go
```diff
@@ -0,0 +1,157 @@
...
+// })
+
+events.process()
+`)
+
+// nolint: lll
+var javaScriptTemplate = []byte(`const { events } = require("@brigadecore/brigadier");
+
+// Use events.on() to define how your script responds to different events. 
+// The examples below depicts handling of "exec" events originating from
```
krancour, line 59: Change "examples" to "commented example".

krancour:
```
Creating .brigade directory...
Creating .brigade/project.yaml...
Creating .brigade/NOTES.txt...
Creating .brigade/secrets.yaml...
Adding .brigade/secrets.yaml to .gitignore...
Adding .brigade/node_modules/ to .gitignore...
```

This created .gitignore, but never told me it was doing so.

I think the `addLinesToFile()` function might need to be decomposed a bit. Let's make some kind of reusable `EnsureFile()` function in the `files` package that will be similar to the `EnsureDirectory()` function. You'll use that first to make sure the file exists and _then_ add lines to it.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,157 @@
...
+  #   - exec
+  workerTemplate:
+    # logLevel: DEBUG
+    {{- if eq .GitCloneURL "" }}
+    defaultConfigFiles:
+      brigade.{{ .Language }}: |
+        {{- nindent 8 .Script }}
+    {{- else }}
+    git:
+      cloneURL: {{ .GitCloneURL }}
```
krancour, line 34: Under `cloneURL`, let's add the following:

```
ref: refs/heads/master
```

Especially given that more and more people are using `main` instead of `master`, this will make it much more obvious how to change this if desired.

v2/cli/init_templates.go
```diff
@@ -0,0 +1,157 @@
...
+  #   types:
+  #   - exec
+  workerTemplate:
+    # logLevel: DEBUG
+    {{- if eq .GitCloneURL "" }}
+    defaultConfigFiles:
+      brigade.{{ .Language }}: |
+        {{- nindent 8 .Script }}
+    {{- else }}
+    git:
```
krancour, line 33: Let's add a comment above this line that indicates that these `git` details can be overridden by individual events. This will explain away behavior that might otherwise surprise people.

Something like this:

```
## Your brigade.{{ .Language }} script will be loaded from the following git repository by default, but note that individual events CAN override these details. The Brigade GitHub Gateway, for instance, overrides these details for relevant events so the events themselves reference applicable pull branches, tags, pull requests, etc.
```

willie-yao:
> Let's make some kind of reusable `EnsureFile()` function in the `files` package

I used the existing `Ensure()` function that is already in the `files` package, and added the check to `addToGitIgnore` since that is specific to the `.gitignore` check. Does that work?

v2/cli/init_command.go
```diff
@@ -241,6 +241,13 @@ func initialize(c *cli.Context) error {
 // addToGitIgnore invokes addLinesToFile, and adds the given paths to
 // the root directory's .gitignore
 func addToGitIgnore(pathsToIgnore ...string) error {
+	fileExists, err := file.Exists(".gitignore")
+	if !fileExists {
+		fmt.Printf("Creating .gitignore...\n")
+	}
+	if err != nil {
+		return err
+	}
```
krancour, line 250: This isn't how I'd been thinking of doing this, but it's clean. I like it.

v2/cli/init_templates.go
```diff
@@ -30,17 +30,24 @@ spec:
       brigade.{{ .Language }}: |
         {{- nindent 8 .Script }}
     {{- else }}
+    ## Your brigade.{{ .Language }} script will be loaded from the 
+    ## following git repository by default, but note that individual 
+    ## events CAN override these details. The Brigade GitHub Gateway, 
+    ## for instance, overrides these details for relevant events so the 
+    ## events themselves reference applicable pull branches, tags, pull
```
krancour, line 37: Tiny mistake in this line (which probably came from me). Remove "pull" from "pull branches".

PR #1442, opened 2021-06-07
feat(operator):  add max worker + job lifetime configurability

vdice:

* Adds the ability for Brigade operators to specify a max lifetime for workers and jobs

Last bit of work for https://github.com/brigadecore/brigade/issues/1378

~~Relies on https://github.com/brigadecore/brigade/pull/1419 and can rebase once 1419 is merged.  Meanwhile, all of the changes specific to this feature are currently in https://github.com/brigadecore/brigade/commit/d8397fa38e91d77816c302e545d8fbc08fb2aae0~~

I've added a default of 24 hours for max lifetime for both workers and jobs -- should we even have a default if left unspecified?

**If applicable**:

### DIFF
--- a/v2/observer/jobs.go
+++ b/v2/observer/jobs.go
@@ -204,26 +204,10 @@ func (o *observer) startJobPodTimer(ctx context.Context, pod *corev1.Pod) {
 		return
 	}
 
-	// Attempt to set the timeout per the annotation on the pod itself
-	duration := pod.Annotations[myk8s.AnnotationTimeoutDuration]
-	timeout, err := time.ParseDuration(duration)
-	// Fallback to the max if we are unable to parse timeout value or if it
-	// exceeds the max.
-	if err != nil || timeout > o.config.maxJobLifetime {
-		o.errFn(
-			fmt.Errorf(
-				"unable to parse timeout duration %q for pod %q; "+
-					"using configured maximum of %q",
-				duration,
-				pod.Name,
-				o.config.maxJobLifetime,
-			),
-		)
-		timeout = o.config.maxJobLifetime
-	}
-
 	go func() {
-		timer := time.NewTimer(timeout)
+		timer := time.NewTimer(
+			o.getPodTimeoutDuration(pod, o.config.maxJobLifetime),
+		)
 		defer timer.Stop()
 
 		select {
--- a/v2/observer/jobs.go
+++ b/v2/observer/jobs.go
@@ -203,22 +203,25 @@ func (o *observer) startJobPodTimer(ctx context.Context, pod *corev1.Pod) {
 		pod.Status.Phase != corev1.PodRunning {
 		return
 	}
-	if pod.Annotations[myk8s.AnnotationTimeoutDuration] == "" {
-		return
-	}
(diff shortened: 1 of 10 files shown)


vdice:
Looking into the unit test failure(s); hadn't seen them locally.

v2/observer/jobs.go
```diff
@@ -203,22 +203,25 @@ func (o *observer) startJobPodTimer(ctx context.Context, pod *corev1.Pod) {
...
+				errors.Wrapf(
+					err,
+					"unable to parse timeout duration %q for pod %q",
+					duration,
+					pod.Name,
+				),
+			)
+			return
+		}
 	}
```
krancour, line 225: Shouldn't we be looking at the pod-specified timeout and reducing it if it exceeds the max?

vdice: Oh, yes!  Thanks.  Will update.

vdice: Updated in https://github.com/brigadecore/brigade/pull/1442/commits/0561129841bf1e9c850cd6c16c7794c93de1a21a

v2/observer/jobs.go
```diff
@@ -222,6 +222,17 @@ func (o *observer) startJobPodTimer(ctx context.Context, pod *corev1.Pod) {
...
+			o.errFn(
+				fmt.Errorf(
+					"timeout %q for pod %q exceeds the configured maximum %q",
+					duration,
+					pod.Name,
+					o.config.maxJobLifetime,
+				),
+			)
+			return
+		}
```
krancour, line 235: I had been thinking that if the pod-specified timeout exceeded the max, we'd just reduce the timeout to the max.

If we return an error here, don't we prevent this pod from _ever_ timing out?

vdice: Ah, yup, as you originally requested, but I missed :)

Updated the logic to fallback to the config'd max if we can't parse the pod duration (including if it is unset/empty) or it exceeds the max.  We log an error in this case, for tracking in the observer logs.

v2/observer/jobs.go
```diff
@@ -203,22 +203,23 @@ func (o *observer) startJobPodTimer(ctx context.Context, pod *corev1.Pod) {
...
+	// Fallback to the max if we are unable to parse timeout value or if it
+	// exceeds the max.
+	if err != nil || timeout > o.config.maxJobLifetime {
 		o.errFn(
-			errors.Wrapf(
-				err,
-				"unable to parse timeout duration %q for pod %q",
+			fmt.Errorf(
+				"unable to parse timeout duration %q for pod %q; "+
+					"using configured maximum of %q",
```
krancour, line 216: Should we use a different message if the issue is that the max timeout allowed by the operator was exceeded? This message will always look like a parsing failure.

vdice: Fair point.  Was trying to simplify logic but at the expense of clarity in logging, which is never a good idea 😂 

Consolidated logic with this latest iteration... lemme know how it's looking when convenient.

krancour (approved):
LGTM!

PR #1444, opened 2021-06-08
make scheduler and observer use new sdk mock clients

krancour:
Follows up on #1440 to make use of the SDK's new mock clients.

This is a modest increase in LOC, but imho, improves the clarity of the scheduler and observer code quite a bit because it requires a lot less thought about the indirection that is taking place.


vdice (approved):
👍 Yup, I really like the clarity/consolidation added in this refactor.  LGTM!

Issue #1438, opened 2021-06-03
Proposal: projects should have to subscribe to events, even it the event has a project id


krancour:
Fundamentally, brigade connects arbitrary events with arbitrary actions.

Brigade 2, especially, is based on a _subscription model_ wherein projects _subscribe_ to events of interest. This allows gateways to create event templates that lack any explicit project identifier, then if multiple projects appear to be subscribed to that event, a discrete event is made for _each_ by copying that template and filling in the projectID field.

But currently, we _also_ allow an events to be created that _directly_ identify which project they're for-- i.e. the projectID field is already filled in. In these cases, it's taken for granted that the event is of interest to the project and the project receives the event-- regardless of what subscriptions it has.

A great example of seeing this in action is in the CLI where one can `brig event create -p <projectID> --source <source> --type <type>` and directly create an event for the indicated project and whether the project is actually subscribed to that event source/type or not ends up being irrelevant-- and that feels off.

I would hereby propose that _even when_ an event directly identifies what project it is for, that the project only receives it if it is explicitly subscribed to it.

This complicates "hello world" only ever-so-slightly because the project definition file goes from this:

```yaml
apiVersion: brigade.sh/v2-alpha.5
kind: Project
metadata:
  id: hello-world
description: The simplest possible example
spec:
  workerTemplate:
    logLevel: DEBUG
    defaultConfigFiles:
      brigade.js: |
        console.log("Hello, World!");
```

To this:

```yaml
apiVersion: brigade.sh/v2-alpha.5
kind: Project
metadata:
  id: hello-world
description: The simplest possible example
spec:
  eventSubscriptions:
  - source: brigade.sh/cli
    types:
    - exec
  workerTemplate:
    logLevel: DEBUG
    defaultConfigFiles:
      brigade.js: |
        console.log("Hello, World!");
```

But the benefit of doing this is it becomes much easier to rationalize about the subscription based model. The notion of there being _two_ ways for a project to receive an event goes away and is replaced by just one.

Also, because we're adding a `brig init` command (see draft #1437) that basically spits out a "hello world" project annotated with comments about further customizations and next steps, the "hello world" project would now actually involve a subscription (for `brigacde.sh/cli:exec`) that it wouldn't have had before, which makes the next steps for getting from "hello world" to something more useful all the more obvious.

cc @vdice

krancour:
If we change this, it really ought to happen before beta.1

vdice:
[...]

PR #1445, opened 2021-06-09
ref(*): projects only receive events if subscribed

vdice:

* Updates logic to ensure that a project only receives events according to their event subscriptions.  The main breaking change is thus: even if an event targets a particular project, it will only be delivered if that particular project is confirmed to be a subscriber.

Closes https://github.com/brigadecore/brigade/issues/1438

**If applicable**:


krancour (approved):
💥 LGTM

Issue #1291, opened 2021-02-19
Job creation errors are inadequately descriptive


krancour:
When the worker gets a non-200 response code from the API when creating a new Job, the error message isn't adequately descriptive.

For example:

```
2021-02-19T16:39:30.507Z DEBUG: writing default brigade.ts to /var/vcs/.brigade/brigade.ts
2021-02-19T16:39:30.511Z DEBUG: no existing package.json found at /var/vcs/.brigade/package.json
2021-02-19T16:39:30.511Z DEBUG: creating a minimal package.json
2021-02-19T16:39:30.511Z DEBUG: writing package.json to /var/vcs/.brigade/package.json
2021-02-19T16:39:30.512Z DEBUG: installing dependencies
2021-02-19T16:39:33.871Z DEBUG: compiling brigade.ts
2021-02-19T16:39:36.599Z INFO: brigade-worker version: 5be6835-dirty
2021-02-19T16:39:36.608Z [job: job1] INFO: Creating job job1
2021-02-19T16:39:36.610Z [job: jobA] INFO: Creating job jobA
2021-02-19T16:39:36.729Z ERROR: Error creating job "jobA": Request failed with status code 400
2021-02-19T16:39:39.153Z [job: job1] DEBUG: Current job phase is SUCCEEDED
2021-02-19T16:39:39.154Z [job: job2] INFO: Creating job job2
2021-02-19T16:39:41.755Z [job: job2] DEBUG: Current job phase is SUCCEEDED
```

I happen to know that the problem here is that `jobA` is invalid because it contains an uppercase character. (Job names are supposed to be lower alphas, numerals, and hyphens only.)

But the error message that _was_ included in the response from the server has been obscured.

krancour:
I think this can be corrected by making the worker use the brigade-sdk-for-js.

That had not existed that the time the new worker code was written.

itowlson:
I had a brief look at this.  Taking a dependency on the Brigade SDK is conceptually ideal, because it means that _all_ clients of the API (including Brigade's own internal components) go through the same paths.  But it also means that we have to coordinate across two repositories and means that we can't do a release of Brigade until _after_ the release of the SDK - a reversal of the normal order of releases - unless we have a mechanism for doing local builds of the JS SDK and pulling those into Brigade (and taking them out again before committing).  That might be worth it but it would be a faff (unless we found and implemented a good refactoring of dependencies).

[...]

PR #1447, opened 2021-06-15
ref(jobs.ts): update brigadier-polyfill's Job.run() to use sdk

vdice:

* Updates Job.run() in brigadier-polyfill to use the [js/ts SDK](https://github.com/brigadecore/brigade-sdk-for-js)
* Had to update brigadier's `ImagePullPolicy` to match the SDK's, so that container specs could be compatible.

Fixes https://github.com/brigadecore/brigade/issues/1291

Willing to re-do the seemingly unrelated yarn.lock changes if y'all can point me how to _only_ add the brigade-sdk dependency :)  I just used `yarn install`.

**If applicable**:

### DIFF
--- a/v2/brigadier-polyfill/package.json
+++ b/v2/brigadier-polyfill/package.json
@@ -27,9 +27,10 @@
     "typescript": "^4.1.2"
   },
   "dependencies": {
+    "@brigadecore/brigade-sdk": "^v2.0.0-alpha.5",
+    "@brigadecore/brigadier": "../brigadier",
     "@types/node": "^14.14.11",
     "axios": "^0.21.0",
-    "@brigadecore/brigadier": "../brigadier",
     "eventsource": "^1.0.7",
     "winston": "^3.3.3"
   }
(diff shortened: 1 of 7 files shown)


v2/brigadier-polyfill/package.json
```diff
@@ -27,9 +27,10 @@
     "typescript": "^4.1.2"
   },
   "dependencies": {
+    "@brigadecore/brigade-sdk": "^v2.0.0-alpha.5",
+    "@brigadecore/brigadier": "../brigadier",
     "@types/node": "^14.14.11",
     "axios": "^0.21.0",
```
krancour, line 33: I think we might be able to dump axios as a direct dependeny.

vdice: Good call; removed.

After running `make lint-js`, it looks like the `yarn.lock` file in `v2/worker` was updated as well -- and they look like identical updates to those in `v2/brigadier-polyfill`, I suppose b/c the latter is a dep for the former.  So, I've added these updates as well.

krancour (approved):
LGTM!

vdice:
(Last push == rebase off of v2)

PR #1449, opened 2021-06-16
add branding to login splash page

### LINKED ISSUE #1131: Login confirmation splash page
In 2.0, with OpenID Connect enabled, the login workflow for a CLI user is this:

1. `brig login --server https://location`
2. API server sends back unactivated token and URL to third party identity provider
3. User completes authentication in their browser using the provided URL
4. Identity provider redirects to API server-- token gets activated
5. API server redirects user to page that confirms their login-- informs them they may resume using the CLI

Currently, no. 5 above is plain text, but it deserves to be spruced up.

This issue requests an attractive "splash page" to confirm login and inform the user they may resume using the CLI.

Near term, assets like HTML, CSS, and JS (if applicable) can be dropped in project path `v2/apiserver/assets`. Longer term, a separate PR can incorporate those assets into the API server.

krancour:
Fixes #1131 

This replaces the plaintext login confirmation page with a branded splash page.

Most credit here goes to @flynnduism for developing the logo, favicon, and other thematic elements like colors and typeface.

@flynnduism I believe I've faithfully adhered to your style guide. Please let me know if you see any issues.

Side note: This is the first HTML or CSS I've touched in _years_. So please also lmk if I've done anything wonky on that front. afaik, everything I've used should be compatible with all modern browsers-- which is good enough, imho.

Screen Shot 2021-06-16 at 2 54 12 PM


vdice (commented):
Rad, looks great to me!  Want to give @flynnduism a chance to take a look.

flynnduism (approved):
This is great! Neat and simple, perfect for the intended scenario. 

:shipit:

PR #1456, opened 2021-06-21
Added k8s version compatibility note to README

willie-yao:
Signed-off-by: willie-yao <t-williamyao@microsoft.com>

Added a note to the README that specifies the minimum compatible version of kubernetes that Brigade 2 supports

**If applicable**:


README.md
```diff
@@ -24,6 +24,8 @@ release. In the meantime, here is a little to get you started.
 
 ### Installing Brigade 2 on a _Private_ Kubernetes Cluster
 
+__Note: Brigade 2 is compatible only with Kubernetes versions 1.16.0+__
```
krancour, line 27: I would remove "only" because having that there subtly implies a larger constraint than this really is. 1.16 is pretty old, so we shouldn't make a big deal out of this constraint.

krancour:
I believe there are also other docs that need to be updated in the `docs/` dir.

willie-yao:
> I believe there are also other docs that need to be updated in the `docs/` dir.

Just updated `developers.md` with a note that kind or minikube need to have versions 1.16.0+. Is there another doc I should add this note to?

krancour (approved):
LGTM

PR #1457, opened 2021-06-21
Limit authn by org

### LINKED ISSUE #1455: Restrict who can register
When OpenID Connect or GitHub is used as the authentication method, anyone known to the identity provider in question can create a Brigade account, as the local Brigade account is created transparently for all new users.

In practice, this isn't _entirely_ terrible because, for instance, a given Azure Active Directory tenant only knows of principals _in your organization_. Complete unknowns cannot come along and create Brigade accounts. Even if they could, they don't receive _any_ permissions within Brigade until an administrator grants them one or more roles.

When using GitHub as the identity provider, however, people unrelated to your organization can authenticate successfully and get a Brigade account. Again, with no permissions being assigned out of the gate, they can't _do_ anything.

Because of the above, Brigade was designed _without_ functionality to delete a user. _If you delete a user, they can always just come back (and start fresh with no permissions). _ Instead, the functionality that was implemented was for _locking_ a user. If you want someone gone, you _lock_ them out, but you remember forever that they're unwelcome.

While all of this works and isn't inherently insecure, it's also a bit rough around the edges.

What I propose is implementing an option that restricts authentication via third-party identity providers to _just_ principals that are members of some operator-specified org or team. This would enable a few key things:

* The option to automatically-grant read-only privileges to new users --> lessen's administrator burden

* We could finally implement user delete (as an alternative to lock) because a deleted principal's ability to come back and re-register can be restricted on the identity provider end by removal from a group / team / tenant.

krancour:
Fixes #1455

### DIFF
--- a/charts/brigade/templates/apiserver/deployment.yaml
+++ b/charts/brigade/templates/apiserver/deployment.yaml
@@ -80,6 +80,8 @@ spec:
             secretKeyRef:
               name: {{ include "brigade.apiserver.fullname" . }}
               key: github-client-secret
+        - name: GITHUB_ALLOWED_ORGANIZATIONS
+          value: {{ join "," .Values.apiserver.thirdPartyAuth.github.allowedOrganizations }}
         {{- end }}
         - name: SCHEDULER_TOKEN
           valueFrom:
--- a/charts/brigade/values.yaml
+++ b/charts/brigade/values.yaml
@@ -59,6 +59,14 @@ apiserver:
       ## itself to GitHub.
       # clientID: ""
       # clientSecret: ""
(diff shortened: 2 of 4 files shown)


vdice (approved):
LGTM!

Any thoughts on adding a similar `allowedUsers: []` section under `github` auth as well?  Perhaps useful for setups that would want to limit usage to a small number of specific GH users.

krancour:
> Any thoughts on adding a similar allowedUsers: [] section under github auth as well? Perhaps useful for setups that would want to limit usage to a small number of specific GH users.

I'm unopposed.

PR #1461, opened 2021-06-28
start dogfooding brigade 2

krancour:
At long last, this PR causes the v2 branch to start dogfooding itself.

The one thing we're still leaning on v1.x for is integration tests because our current approach to integration tests involves a number of host mounts and in v2, for security reasons, thus far, the only host mount we permit is the Docker socket.

We will either have to find a new approach to integration testing or reconsider the merits (an demerits) of allowing arbitrary host paths to be mounted into a pod.

In the meantime, however, this PR transitions most of what we do pre/post-merge to Brigade 2 and imho, it's better to do that sooner rather than later as it will help us cut beta.1 with a higher level of confidence.

Note to the reviewer: Except for the integration tests, don't expect checks formerly run by v1.x to run. After this is merged (using administrator privilege), we will have to adjust the `v2` branch protections to require the new checks to pass in lieu of the old ones.


.brigade/brigade.ts
```diff
@@ -0,0 +1,292 @@
...
+      pushGitInitializerJob(event),
+      pushLoggerLinuxJob(event),
+      pushObserverJob(event),
+      pushSchedulerJob(event),
+      pushWorkerJob(event),
+    ).run()
+  }
+}
+
+// Either of these events should initiate execution of the entire test suite.
```
vdice, line 248: Can we also add handlers for `issue_comment:created` and `issue_comment:edited` for `/brig run` and `/brig check` support?

.brigade/project.yaml
```diff
@@ -0,0 +1,18 @@
...
+description: Brigade 2 built with Brigade 2!
+spec:
+  eventSubscriptions:
+  - source: brigade.sh/github
+    qualifiers:
+      repo: brigadecore/brigade
+    types:
+    - check_run:rerequested
+    - check_suite:requested
+    - check_suite:rerequested
```
vdice, line 15: (see request to handle `issue_comment` events in the `brigade.ts` file; if we do so, we'd add the corresponding types here)

brigade.js
```diff
@@ -326,11 +71,6 @@ function run(e, p, job) {
 events.on("check_suite:requested", runSuite);
```
vdice, line 71: Should we keep the issue comment checks?  Then, while we're in a hybrid state (v1 running integration tests, the rest running via v2), `/brig run|check` would trigger re-running these as well, potentially.

vdice (commented):
Looks great!  Just questions around handling issue comment events.

As a follow-up after this is merged, we'd update the required checks list, right? (test-integration from v1, the rest are new from v2)

.brigade/brigade.ts
krancour: There's no need. The new gateway does all of that. In the same way that it automatically talks back to github to request a check_suite when an eligible PR is opened or synchronized, it now automatically talks back to github to request a check_suite for eligible issue comment events.

krancour: So what's nice is users of that gateway, in most cases, will just want to subscribe to various check events, without concern for _how_ or why the check was requested.

.brigade/project.yaml
krancour: [Same answer](https://github.com/brigadecore/brigade/pull/1461#discussion_r660995142). Gateway takes care of it.

brigade.js
krancour: Here yes. Perhaps we should. Fixing.

krancour: Fixed.

.brigade/brigade.ts
vdice: Ahh, nice!  I could've/should've realized that from reviewing the v2 GH gateway 😄 

Super slick.

vdice (approved):
LGTM

.brigade/brigade.ts
krancour: No worries. It was a blink and you miss it sorta thing.

PR #1472, opened 2021-07-01
ci(logger-windows): add CD

vdice:
* Adds/updates CI/CD for the Windows logger agent

Closes https://github.com/brigadecore/brigade/issues/1277

PR #1473, opened 2021-07-01
fix(.github): set docker registry on push of windows logger image

vdice:
Quick fix/follow-up for https://github.com/brigadecore/brigade/pull/1472

Fixes failure seen in https://github.com/brigadecore/brigade/runs/2966940164

### DIFF
--- a/.github/workflows/merge.yaml
+++ b/.github/workflows/merge.yaml
@@ -20,4 +20,4 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - name: Push Windows logger agent
-        run: make push-logger-windows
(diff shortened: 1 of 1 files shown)


.github/workflows/merge.yaml
```diff
@@ -20,4 +20,4 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - name: Push Windows logger agent
-        run: make push-logger-windows
\ No newline at end of file
+        run: DOCKER_REGISTRY=brigadecore make push-logger-windows
```
krancour, line 23: I think this should be `DOCKER_ORG= brigadecore`

vdice: Ah, you're right.  While both would work here, DOCKER_ORG is definitely more proper.  Fixed!

krancour (approved):

PR #1474, opened 2021-07-01
fix(.github): use action env stanza for env vars

vdice:
We set it at the job level so that the built images will have the correct org as well.

### DIFF
--- a/.github/workflows/merge.yaml
+++ b/.github/workflows/merge.yaml
@@ -10,6 +10,8 @@ on:
 jobs:
   build:
     runs-on: windows-latest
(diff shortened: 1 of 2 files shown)


krancour (approved):

vdice:
Most recent commit: I wanted to add the same `DOCKERHUB_ORG` env var to the pr.yaml as well, both to make sure it works as intended and to have alignment with merge.yaml, even though the former isn't strictly necessary as we don't currently push on PRs...

krancour:
Still LGTM

Issue #1460, opened 2021-06-25
Unit test failure for ThirdPartyAuthHelper


vdice:
(for Brigade v2)

It looks like a failure started to crop up today around the following thirdPartyAuth test: https://github.com/brigadecore/brigade/blob/v2/v2/apiserver/config_test.go#L308-L324

```
=== CONT  TestThirdPartyAuthHelper
    config_test.go:321:
        	Error Trace:	config_test.go:321
        	            				config_test.go:406
        	Error:      	"400 Bad Request: {\"error\":\"invalid_tenant\",\"error_description\":\"AADSTS90002: Tenant 'cc18ecf3-7acb-4d14-ba43-fc25dc310191' not found. This may happen if there are no active subscriptions for the tenant. Check to make sure you have the correct tenant ID. Check with your subscription administrator.\\r\\nTrace ID: 1bd2113a-777d-499c-aba8-3a2ad1ea4600\\r\\nCorrelation ID: e186699d-216d-40bb-a281-54eca6691df8\\r\\nTimestamp: 2021-06-25 19:31:43Z\",\"error_codes\":[90002],\"timestamp\":\"2021-06-25 19:31:43Z\",\"trace_id\":\"1bd2113a-777d-499c-aba8-3a2ad1ea4600\",\"correlation_id\":\"e186699d-216d-40bb-a281-54eca6691df8\",\"error_uri\":\"https://login.microsoftonline.com/error?code=90002\"}" does not contain "value not found for"
        	Test:       	TestThirdPartyAuthHelper
=== CONT  TestThirdPartyAuthHelper/OIDC_CLIENT_ID_required_but_not_set
    testing.go:1048: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
```

See https://brigadecore.github.io/kashti/jobs/test-unit-go-01f91zsm0r1k5w07d087c2cqnp for full logs

It looks like we are actually hitting the msft/azure auth endpoint?  cc @krancour

vdice:
Fixed by https://github.com/brigadecore/brigade/pull/1479

PR #1479, opened 2021-07-08
fix(config_test.go): update ThirdPartyAuthHelper tests to hit test server

vdice:
* Updates the ThirdPartyAuthHelper tests to hit a test server instead of an external provider

Closes https://github.com/brigadecore/brigade/issues/1460


krancour (approved):
LGTM

Issue #1137, opened 2020-10-20
Fix flaky TestListenAndServe test


krancour:
This test is flaky. I _think_ there's a race condition in the test code (not in the code _under_ test).

radu-matei:
Running all the tests in `v2/apiserver/internal/lib/restmachinery`, sometimes I get:

```
=== RUN   TestListenAndServe
=== RUN   TestListenAndServe/TLS_enabled;_missing_cert_path
=== RUN   TestListenAndServe/TLS_enabled;_missing_key_path
=== RUN   TestListenAndServe/TLS_enabled;_cert_not_found
=== RUN   TestListenAndServe/TLS_enabled;_key_not_found
=== RUN   TestListenAndServe/TLS_enabled;_invalid_cert
2021/01/25 12:55:43 API server is listening with TLS enabled on 0.0.0.0:8080
=== RUN   TestListenAndServe/TLS_enabled
2021/01/25 12:55:50 API server is listening with TLS enabled on 0.0.0.0:8080
=== RUN   TestListenAndServe/TLS_not_enabled
2021/01/25 12:55:50 API server is listening without TLS on 0.0.0.0:8080
=== CONT  TestListenAndServe
    server_test.go:197:
        	Error Trace:	server_test.go:197
        	            				server_test.go:208
        	Error:      	Not equal:
        	            	expected: <nil>(<nil>)
        	            	actual  : *net.OpError(&net.OpError{Op:"listen", Net:"tcp", Source:net.Addr(nil), Addr:(*net.TCPAddr)(0xc00016cc60), Err:(*os.SyscallError)(0xc00046dd40)})
        	Test:       	TestListenAndServe
=== CONT  TestListenAndServe/TLS_not_enabled
    testing.go:1052: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
--- FAIL: TestListenAndServe (6.10s)
    --- PASS: TestListenAndServe/TLS_enabled;_missing_cert_path (0.00s)
    --- PASS: TestListenAndServe/TLS_enabled;_missing_key_path (0.00s)
    --- PASS: TestListenAndServe/TLS_enabled;_cert_not_found (0.00s)
    --- PASS: TestListenAndServe/TLS_enabled;_key_not_found (0.00s)
    --- PASS: TestListenAndServe/TLS_enabled;_invalid_cert (0.00s)
    --- PASS: TestListenAndServe/TLS_enabled (6.10s)
[...]

PR #1480, opened 2021-07-08
fix(server_test.go): use different ports for ListenAndServe test variants

vdice:
**Update:** thanks to @krancour's suggestion of just using different ports for the test variants in question, this is what I've done.  These 808x ports seem to always be available in a test context and the race condition of one being in use when the next test runs is thus averted.

~~* Adds allowance for "address already in use" errors to pop up in the ListenAndServe tests~~

~~I figured this would be acceptable as it doesn't commonly occur and we ensure any other type of error does fail the test. Thoughts?~~

~~We'll know when it pops up:~~
```
2021/07/08 21:29:01 API server is listening with TLS enabled on 0.0.0.0:8080
=== RUN   TestListenAndServe/TLS_enabled
2021/07/08 21:29:10 API server is listening with TLS enabled on 0.0.0.0:8080
=== RUN   TestListenAndServe/TLS_not_enabled
2021/07/08 21:29:10 API server is listening without TLS on 0.0.0.0:8080
    server_test.go:217: Received error "address already in use"
--- PASS: TestListenAndServe (9.52s)
```

~~I also figured it wasn't worth the hassle to add in logic to try to rerun the test when we hit this, since most of the test runs in CI/locally will not encounter this behavior.  But open to ways to do so if we wish!~~

Closes https://github.com/brigadecore/brigade/issues/1137

### DIFF
--- a/v2/apiserver/internal/lib/restmachinery/server_test.go
+++ b/v2/apiserver/internal/lib/restmachinery/server_test.go
@@ -145,6 +145,7 @@ func TestListenAndServe(t *testing.T) {
 				keyFile, err := ioutil.TempFile("", "tls-*.key")
 				require.NoError(t, err)
 				return &ServerConfig{
+					Port:        8082,
 					TLSEnabled:  true,
 					TLSCertPath: certFile.Name(),
 					TLSKeyPath:  keyFile.Name(),
@@ -177,6 +178,7 @@ func TestListenAndServe(t *testing.T) {
 				require.NoError(t, err)
 
 				return &ServerConfig{
+					Port:        8081,
 					TLSEnabled:  true,
 					TLSCertPath: certFile.Name(),
 					TLSKeyPath:  keyFile.Name(),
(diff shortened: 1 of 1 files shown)


krancour:
@vdice nice work diagnosing this issue!

Now that we know it's all about the port used by the first of two success cases still being in use when the second of the two success cases runs, I'm wondering if we can fix this more easily by just making those two success cases use two different ports. That could be done very easily by just modifying the `setup` for both of those cases to return `ServerConfig` with the `Port` field filled in.

Thoughts?

vdice:
> That could be done very easily by just modifying the setup for both of those cases to return ServerConfig with the Port field filled in.

Yea, this would be the better way to go.  I briefly experimented w/ this approach but didn't know how to do it besides creation of a `net.Listener` to find an open/avail port.  Is there a better/easier way to do this?

vdice:
Probably if we just use `8082` for the first tls test, `8081` for the next and `8080` for the non-tls test, we should be good.  Running loops of this now... Perhaps we don't need to get clever with picking a random avail port... these seem to be available in the test context.

vdice:
Updated PR per https://github.com/brigadecore/brigade/pull/1480#issuecomment-877349649 -- good call @krancour !

krancour:
> Perhaps we don't need to get clever with picking a random avail port... these seem to be available in the test context.

Given that these tests usually run in a (clean room) container, it's probably safe to assume those ports are available.

krancour (approved):
LGTM

PR #1486, opened 2021-07-13
brig init validation of git clone urls

willie-yao:
Currently, there is no validation check for a Git link provided through the `--git` flag for `brig init`.  This PR adds this functionality by validating that the provided link begins with `http://`, `https://`, or `git@` and  ends with `.git`.


v2/cli/init_command.go
```diff
@@ -310,3 +316,17 @@ func checkFileOverwrite(
...
+			`((git|ssh|http(s)?)|(git@[\w\.]+))` +
+				`(:(//)?)([\w\.@\:/\-~]+)(\.git)(/)?`)
+	urlMatch := gitCloneURLRegex.MatchString(url)
+	if !urlMatch {
+		return fmt.Errorf("invalid value %q for git clone URL"+
+			" (must start with http://, https://, or git@"+
+			" and must end with .git)", url)
+	}
+	return nil
+}
```
krancour, line 332: I can see this being useful in other contexts. Let's move this to `sdk/v2/core` like we did with `ValidateProjectID()`.

v2/cli/init_command.go
```diff
@@ -310,3 +316,17 @@ func checkFileOverwrite(
 	}
 	return true, false, nil
 }
+
+func validateGitCloneURL(url string) error {
+	gitCloneURLRegex :=
+		regexp.MustCompile(
+			`((git|ssh|http(s)?)|(git@[\w\.]+))` +
+				`(:(//)?)([\w\.@\:/\-~]+)(\.git)(/)?`)
```
krancour, line 324: Don't break the regex across two lines. It makes it harder to read. Use a `// nolint: lll` to ignore the error from the `lll` linter if you have to.

v2/cli/init_command.go
willie-yao: Which file would this go in? I don't see a specific file in core that fits this function.

v2/cli/init_command.go
krancour: Also, you can use https://regoio.herokuapp.com/ to test your expression.

v2/cli/init_command.go
krancour: Let's add a new file called `validations.go`. You can move this _and_ `ValidateProjectID()` to there. Move tests accordingly to `validations_test.go`.

v2/cli/init_command.go
krancour: This regex might be more concise, readable, and performant:

```
`^(?:(?:https?://)|(?:git@))[\w:/\-\.\?=@&]*\.git$`
```

Not the use of `?:` following opening parenthesis. In regular expressions, parenthesis are used not only for grouping things, but also defining "capture groups" which are useful when you want to extract a segment of what was matched. We don't care about doing so here, so we can use `?:` to _stop_ the creation of a capture group. It will make the regex perform better.

krancour: Further refining this, this regex is slightly better even:

```
`^(?:(?:https?://)|(?:git@))[\w:/\-\.\?=@&]+\.git$`
```

willie-yao: This regex doesn't allow for an empty string for when a user doesn't use the `--git` flag. Should I just add a conditional to check for an empty string instead?

krancour: Ya. This deliberately doesn't validate the empty string. Only check for validity if the value isn't the empty string.

sdk/v2/core/validations.go
```diff
@@ -0,0 +1,35 @@
...
+// nolint: lll
+var gitCloneURLRegex = regexp.MustCompile(`^(?:(?:https?://)|(?:git@))[\w:/\-\.\?=@&]+\.git$`)
+
+// ValidateProjectID checks if a given id is valid.
+func ValidateProjectID(id string) error {
+	idMatch := projectIDRegex.MatchString(id)
+	if !idMatch || len(id) < 3 || len(id) > 63 {
+		return fmt.Errorf("invalid value %q for project id"+
+			" (3-63 characters, first char must be"+
+			" a letter, lowercase letters only)", id)
```
krancour, line 19: Nit: To enforce a consistent style, I'd prefer this were formatted like this:

```
return fmt.Errorf(
	"invalid value %q for project id"+
		" (3-63 characters, first char must be"+
		" a letter, lowercase letters only)",
	id,
)
```

The rationale for this is twofold.

1. Once a line gets so long that we need to break it up across a few lines, our brains swap from horizontal mode to vertical mode. Instead of looking across to see comma delimited args, we look down to see line break delimited args. As such, we should try to keep one arg per line because it makes everything more clear to someone who's now scanning down. Don't make people scan down _and_ across. idk if that makes sense.

2. If the closing parenthesis is on its own outdented line, it's visually much more obvious what opening paren it corresponds to.

sdk/v2/core/validations.go
```diff
@@ -0,0 +1,35 @@
...
+
+// ValidateProjectID checks if a given git clone URL is valid by ensuring it
+// begins with "http://", "https://", or "git@" and ends with ".git"
+func ValidateGitCloneURL(url string) error {
+
+	urlMatch := gitCloneURLRegex.MatchString(url)
+	if !urlMatch {
+		return fmt.Errorf("invalid value %q for git clone URL"+
+			" (must start with http://, https://, or git@"+
+			" and must end with .git)", url)
```
krancour, line 32: Same comment as [here](https://github.com/brigadecore/brigade/pull/1486/files#r671296503)

v2/cli/init_command.go
```diff
@@ -62,6 +62,13 @@ func initialize(c *cli.Context) error {
 		return err
 	}
 
+	if fields.GitCloneURL != "" {
+		err = core.ValidateGitCloneURL(fields.GitCloneURL)
+		if err != nil {
+			return err
+		}
```
krancour, line 69:
```
if fields.GitCloneURL != "" {
	if err = core.ValidateGitCloneURL(fields.GitCloneURL); err != nil {
		return err
	}
}
```

krancour: Combine assignments and error-checking conditionals whenever you can. To be clear, the first version wasn't bad, but the revision I am suggesting is very clear and concise that we're handling errors related to a particular function call.

krancour:
A few stylistic nits... functionally, this looks good.

krancour (approved):

PR #1493, opened 2021-07-14
docs(overview.md): update for v2

vdice:
* Updates the overview doc per v2

I ran with the existing structure, but also open to changing it up if others feel strongly.  I am slightly worried that I got too 'wordy' for the existing bulleted-list format... WDYT?


krancour:
Seems to me like a good start. LGTM

krancour (approved):

PR #1503, opened 2021-07-19
add creation date/time stamp to jobs

krancour:
In Kubernetes terms (since our API is heavily inspired by theirs), jobs are "synthetic resources" because they aren't their own thing than can exist independently of something else-- namely an event. This quirk means they're not composed of an `ObjectMeta` and as such were lacking a `Created` date/time stamp field. This info turns out to be useful, however, so this PR adds it.


radu-matei (approved):
LGTM.

Non-blocking: do we want to add a job test for this?

krancour:
@radu-matei 

> Non-blocking: do we want to add a job test for this?

Wanted to, but job creation doesn't return the job so I don't have any way to assert that field go filled in.

PR #1510, opened 2021-07-20
force cleanup for manually deleted worker and job pods

### LINKED ISSUE #1506: Worker resource cleanup if/when a worker pod manually deleted
When a worker pod is manually deleted, will all associated K8s resources be successfully cleaned up?  @krancour thinks there may be some hardening we can do here.

krancour:
Fixes #1506

### DIFF
--- a/v2/observer/jobs_test.go
+++ b/v2/observer/jobs_test.go
@@ -34,7 +34,7 @@ func TestSyncJobPods(t *testing.T) {
 			defer mu.Unlock()
 			syncJobPodFnCallCount++
 		},
-		syncDeletedPodFn: func(_ interface{}) {
+		syncDeletedJobPodFn: func(_ interface{}) {
 			mu.Lock()
 			defer mu.Unlock()
 			syncDeletedPodFnCalled = true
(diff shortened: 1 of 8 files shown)


v2/observer/jobs_test.go
```diff
@@ -495,6 +495,39 @@ func TestDeleteJobResources(t *testing.T) {
...
+		},
+		syncMu: &sync.Mutex{},
+		jobsClient: &coreTesting.MockJobsClient{
+			UpdateStatusFn: func(
+				context.Context,
+				string,
+				string,
+				core.JobStatus,
+			) error {
+				return nil
```
vdice, line 513: Not sure if overkill... do we want to assert on expected status (ABORTED) here?

Similarly, do we want to add assertions on strings passed in to `CleanupFn`?  

(If either/both are added, these comments apply to the worker pod test as well.)

krancour: > Not sure if overkill.

Not sure either, but happy to add.

vdice: Ok, might as well add when convenient.  Thanks!

krancour: All set.

vdice (approved):
LGTM!

Issue #1509, opened 2021-07-20
Brigade Roles: The main roles of interaction within Brigade


vdice:
Fill out the main `topics/roles` doc and the overview (index) doc for each role section (operators, administrators, project developers)

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1511

PR #1511, opened 2021-07-21
docs(*): fill out roles index and overviews for each

vdice:
* Adds documentation around roles in Brigade
* ~~Also snuck in a placeholder for a [Metrics](https://github.com/brigadecore/brigade-metrics) doc 😄~~

TODO
  - [x] Get linking working right 😂 

Closes https://github.com/brigadecore/brigade/issues/1509

### DIFF
--- a/docs/config.toml
+++ b/docs/config.toml
@@ -168,11 +168,6 @@ relativeURLs = "true"
 
     # Other Topics
     [[menu.index]]
-    name = "Metrics"
-    url = "/topics/metrics/"
-    weight = 74
-    parent = "topics"
-    [[menu.index]]
     name = "Developer Guide"
     url = "/topics/developers/"
     weight = 90
--- a/docs/config.toml
+++ b/docs/config.toml
@@ -168,6 +168,11 @@ relativeURLs = "true"
 
     # Other Topics
     [[menu.index]]
+    name = "Metrics"
+    url = "/topics/metrics/"
+    weight = 74
+    parent = "topics"
+    [[menu.index]]
     name = "Developer Guide"
     url = "/topics/developers/"
     weight = 90
--- a/docs/content/topics/administrators/_index.md
+++ b/docs/content/topics/administrators/_index.md
@@ -19,6 +19,9 @@ They'll select a supported third-party authentication strategy that an
 configuring Service Account, User and Role authorization strategies within
 Brigade.
 
+On account of this overlap of responsibilities between an Administrator and
+Operator, a common scenario is that both roles are filled by the same user.
+
   * [Authentication]: Here we discuss the supported authentication strategies
     and configuration in Brigade
 
--- a/docs/content/topics/administrators/_index.md
+++ b/docs/content/topics/administrators/_index.md
@@ -5,9 +5,24 @@ section: topics
 weight: 4
 aliases:
   - /index.md
+  - /administrators
   - /topics/administrators/index.md
 ---
 
 # Brigade Administrators
 
-TODO
+Users who are in charge of authentication and authorization concerns within the
+Brigade system are known as Administrators.
+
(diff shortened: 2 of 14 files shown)


docs/config.toml
```diff
@@ -158,6 +158,11 @@ relativeURLs = "true"
     weight = 70
     parent = "scripting"
     [[menu.index]]
+    name = "Metrics"
+    url = "/topics/metrics/"
+    weight = 74
+    parent = "topics"
+    [[menu.index]]
```
krancour, line 165: Brigade Metrics, for now, is a separate project and basically a bolt-on. Does that influence whether we should or shouldn't cover it here in these docs? It's not strongly held, but I think my opinion is that it's only worthy of a brief mention in a section on other things in the extended Brigade 2 ecosystem.

docs/content/topics/administrators/_index.md
```diff
@@ -3,9 +3,24 @@ title: Administrators
...
 ---
 
 # Brigade Administrators
 
-TODO
\ No newline at end of file
+Users who are in charge of authentication and authorization concerns within the
+Brigade system are known as Administrators.
+
+They'll select a supported third-party authentication strategy to plug into
```
krancour, line 15: They strategy would have had to be selected by the operator at deploy time.

docs/config.toml
vdice: Fair point; removed for now.

docs/content/topics/administrators/authentication.md
```diff
@@ -1,6 +1,10 @@
 ---
 title: Authentication
 description: Authentication configuration for Brigade
+aliases:
+  - /authentication
+  - /topics/authentication.md
+  - /topics/administrators/authentication.md
```
krancour, line 7: Isn't this an operator concern?

docs/content/topics/administrators/_index.md
```diff
@@ -3,9 +3,26 @@ title: Administrators
...
+  - /topics/administrators
   - /topics/administrators/index.md
 ---
 
 # Brigade Administrators
 
-TODO
\ No newline at end of file
+Users who are in charge of authentication and authorization concerns within the
+Brigade system are known as Administrators.
```
krancour, line 13: I'm worried this definition isn't general enough.

What if we present admins as users who are responsible for _managing_ an instance of Brigade?

Maybe also worth noting that this role may often, but not always, be fulfilled by the same person/people fulfilling the operator role?

krancour: The concern about this not being a general enough description repeats in a few other spots.

docs/content/topics/administrators/authentication.md
vdice: This is one that definitely straddles both roles.  I would think admins would want a choice over which third-party auth system is deployed (by an operator), as it relates to users which they'll set up authz around... but I'm down to move it if we feel strongly.

docs/content/topics/administrators/_index.md
vdice: Indeed, I added the following text to the roles doc:

```
Naturally, there may be overlap between these roles, and for development
setups, one or two users might cover all of them.  However, they also serve
nicely as categories of context for documentation, so without further ado,
let's dive in.
```

But I can certainly add role-specific notes to each index, such as the one you've pointed out here, where the admin role might naturally be filled by the operator.

docs/content/topics/administrators/authentication.md
krancour: Interesting point. I can imagine admins wanting some say in the matter. Let's roll with it as is for now and we can change it later if necessary.

krancour (approved):

Issue #1498, opened 2021-07-15
Docs: Make sure nav links are accurate


vdice:
Looking for help around the bottom nav links.  They seem to be auto-configured to follow alphabetical order.  I was under the impression that the `weight`s specified in the [config.toml](https://github.com/brigadecore/brigade/blob/21e67dbe7fdaec44f62cb86c360fe519e5feb87f/docs/config.toml) would dictate ordering, but it doesn't seem to be.

In the screenshot below, taken of the Topics index, I was hoping for the next link to be to the Design doc (`/topics/design.md`) as opposed to the Administrators section.

<img width="991" alt="Screen Shot 2021-07-15 at 9 42 37 AM" src="https://user-images.githubusercontent.com/9322017/125817060-32a53ca3-a1fc-4ef3-8425-82410350ef95.png">

_Originally posted by @vdice in https://github.com/brigadecore/brigade/issues/1487#issuecomment-880806232_

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1514

PR #1514, opened 2021-07-22
docs(*): align prev/next pagination according to config.toml layout

vdice:
* Aligns the prev/next pagination with the layout prescribed in the config.toml

Basically, if a `weight` isn't declared in a doc, the pagination will fall back to alphabetical ordering.

Fixes https://github.com/brigadecore/brigade/issues/1498

### DIFF
--- a/docs/content/_index.md
+++ b/docs/content/_index.md
@@ -18,7 +18,7 @@ writing advanced Brigade scripts. [Dive deep into what makes Brigade, Brigade](t
 
 ### Releasing Brigade
 
-For maintainers of this project, steps on cutting a new release of Brigade can be found [here](releasing).
+For maintainers of this project, steps on cutting a new release of Brigade can be found [here](topics/releasing).
 
 ### Related Resources
 
--- a/docs/config.toml
+++ b/docs/config.toml
@@ -58,134 +58,134 @@ relativeURLs = "true"
     [[menu.index]]
     name = "Operators"
     url = "/topics/operators/"
-    weight = 45
+    weight = 50
     identifier = "operators"
     parent = "topics"
     [[menu.index]]
     name = "Deployment"
     url = "/topics/operators/deploy/"
-    weight = 46
+    weight = 51
     parent = "operators"
     [[menu.index]]
     name = "Securing Brigade"
     url = "/topics/operators/security/"
-    weight = 47
+    weight = 52
     parent = "operators"
     [[menu.index]]
     name = "Storage"
     url = "/topics/operators/storage/"
-    weight = 48
+    weight = 53
     parent = "operators"
     [[menu.index]]
     name = "Gateways"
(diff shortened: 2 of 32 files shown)


docs/content/_index.md
```diff
@@ -18,7 +18,7 @@ writing advanced Brigade scripts. [Dive deep into what makes Brigade, Brigade](t
 
 ### Releasing Brigade
 
-For maintainers of this project, steps on cutting a new release of Brigade can be found [here](releasing).
+For maintainers of this project, steps on cutting a new release of Brigade can be found [here](topics/releasing).
```
krancour, line 21:
```suggestion
For maintainers of this project, steps for cutting a new release of Brigade can be found [here](topics/releasing).
```

krancour:
I'm thinking back to when I was about 11 years old and coding in GW-BASIC-- which had line numbers. 🤮 

It was a common practice to number lines by 10s because that way if you had to insert a few lines of code somewhere, you didn't have to renumber _everything_ that followed...

I'm wondering if we should do something similar with the weights here. (Also wondering if we _can_ do that, as I don't profess to know the inner workings of the doc system very well.)

vdice:
> I'm wondering if we should do something similar with the weights here.

Definitely, we could use/follow the practice of numbering in 10s to allow for further doc additions.

krancour:
@vdice I'm confused by weights, I think.

I see weights assigned to each page in the config.toml, but I also see weights (different weights) in the "front matter" on each page. The weights in the config.toml don't match the weights in each page's front matter.

I'm less concerned about the mismatch and more concerned about the confusion this presents. Are both weights used? Do they represent different things? Or are the weights in one location or the other superfluous.

I apologize, as much of this question is simply fueled by me not understanding Hugo.

vdice:
> I'm less concerned about the mismatch and more concerned about the confusion this presents. Are both weights used? Do they represent different things? Or are the weights in one location or the other superfluous.

Yes, the multiple weight methods are confusing.  I can add a note around this somewhere... maybe in `docs/readme.md`?

The weights for the config.toml specify ordering in the sidebar menu:

Screen Shot 2021-07-27 at 10 09 02 AM

...while the weight on each doc specifies ordering within a sub-section (intro, topics, topics/operators, etc.) for pagination (prev/next) ordering:

Screen Shot 2021-07-27 at 10 09 13 AM

I'm not yet savvy enough with the docs infra to know how to simplify this approach or get rid of the individual doc weight approach... perhaps as we grow more familiar it'll get nicer :)

krancour:
@vdice. I see. Thanks for the explanation!

krancour (approved):
LGTM

vdice:
Last commit adds a note around weights: https://github.com/brigadecore/brigade/pull/1514/commits/bfbf41c525fb252157b0ccc29564b99b4981da13

Issue #1516, opened 2021-07-25
Deployment: How to deploy and manage Brigade


vdice:
Add full, production-grade deployment steps under the Operator's deploy doc.

Go over setup for ingress + DNS, third-party certificate provisioning and third-party auth configuration.

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1518

PR #1518, opened 2021-07-27
docs(operator): add deploy doc

vdice:
* Adds the deployment doc under the Operator section, demonstrating setup with ingress/TLS, cert management and third-party auth

* Added a few modifications to the security doc, but intend for a full update to occur in a later issue

Closes https://github.com/brigadecore/brigade/issues/1516


docs/content/topics/operators/deploy.md
```diff
@@ -11,4 +11,429 @@ aliases:
...
+  - [Choosing a third-party authentication provider for managing user auth](#third-party-authentication)
+
+(Alternatively, if you're interested in using Brigade in a local development
+environment or trial capacity, take a look at the [QuickStart].)
+
+[QuickStart]: /intro/quickstart.md
+
+## Prerequisites
+
+* A [Kubernetes cluster], version 1.18.8 or higher.
```
krancour, line 28: I believe we determined that we're compatible with 1.16.0+.

docs/content/topics/operators/deploy.md
```diff
@@ -11,4 +11,429 @@ aliases:
 
 # Deploying and managing Brigade
 
-TODO
\ No newline at end of file
+In this doc, we'll go over steps for setting up and managing a full,
+production-grade Brigade deploy.  This includes:
```
krancour, line 15:
```suggestion
production-grade Brigade deployment.  This includes:
```

docs/content/topics/operators/deploy.md
```diff
@@ -11,4 +11,429 @@ aliases:
 
 # Deploying and managing Brigade
 
-TODO
\ No newline at end of file
+In this doc, we'll go over steps for setting up and managing a full,
```
krancour, line 14: Total nit: I'd strike "full" just because I don't think it adds value on top of the "production-grade".

docs/content/topics/operators/deploy.md
```diff
@@ -11,4 +11,429 @@ aliases:
...
+
+(Alternatively, if you're interested in using Brigade in a local development
+environment or trial capacity, take a look at the [QuickStart].)
+
+[QuickStart]: /intro/quickstart.md
+
+## Prerequisites
+
+* A [Kubernetes cluster], version 1.18.8 or higher.
+  Your cluster must be accessible to the outside world, e.g. to third-party
```
krancour, line 29: Third party authn doesn't need the API server to be reachable by anything other than your own web browser. This is a requirement for the gateways more than anything else.

docs/content/topics/operators/deploy.md
```diff
@@ -11,4 +11,429 @@ aliases:
...
+```
+
+We're now ready to view and edit configuration options:
+
+```console
+$ helm inspect values ~/charts/brigade > ~/brigade2-values.yaml
+```
+
+In the next steps, we'll go through the configuration options needed to
+acheive the objectives stated at the start of this document.
```
krancour, line 67:
```suggestion
achieve the objectives stated at the start of this document.
```

krancour, line 67: Also wonder if it's better to say "achieve a production-grade deployment". Referencing the start of the document potentially disrupts the flow by making readers scroll back to refresh their memory.

docs/content/topics/operators/deploy.md
```diff
@@ -11,4 +11,429 @@ aliases:
...
+Before configuring Brigade to create an [Ingress resource] for its API server,
+we need to deploy an [Ingress Controller] as well as create a DNS hostname
+record which will map to the Ingress controller's external IP address.
+
+[Ingress resource]: https://kubernetes.io/docs/concepts/services-networking/ingress/
+[Ingress Controller]: https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/
+
+### Deploying the Ingress Controller
+
+There are myriad Ingress Controller options.  For this guide, we will use the
```
krancour, line 80:
```suggestion
There are many Ingress Controller options.  For this guide, we will use the
```

Love the word "myriad," but given the likely number of readers for whom English may be a second language, I think "many" will be more widely understood.

docs/content/topics/operators/deploy.md
```diff
@@ -11,4 +11,429 @@ aliases:
...
+We're now ready to view and edit configuration options:
+
+```console
+$ helm inspect values ~/charts/brigade > ~/brigade2-values.yaml
+```
+
+In the next steps, we'll go through the configuration options needed to
+acheive the objectives stated at the start of this document.
+
+## Managing inbound traffic via Ingress
```
krancour, line 69: idk that this is strictly required for a production-grade deployment.

Should we, perhaps, note that this is merely recommended?

krancour: Adding to this thought, perhaps installing an ingress controller is beyond the scope of this document and, in which case, perhaps we narrow the focus on enabling ingress-related options in `values.yaml`?

docs/content/topics/operators/deploy.md
```diff
@@ -11,4 +11,429 @@ aliases:
...
+    tls:
+      enabled: true
+      ## We set this to false as we'll use an external certificate provider
+      ## in the next step
+      generateSelfSignedCert: false
+```
+
+Save these changes to `brigade-values2.yaml` and proceed to the next section.
+
+## Certificates and TLS
```
krancour, line 175: I agree that TLS (with trusted certs) is a component of a production-grade deployment.

But are we going too far if we go into detail on deployment and use of a cert manager?

Should we consider narrowing the focus to cert-related option in the `values.yaml` (which includes options that don't involve a cert manager)?

docs/content/topics/operators/deploy.md
```diff
@@ -11,4 +11,429 @@ aliases:
...
+      cert-manager.io/cluster-issuer: "letsencrypt-prod"
+      cert-manager.io/acme-challenge-type: http01
+      nginx.ingress.kubernetes.io/secure-backends: "true"
+      nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
+```
+
+## Third-Party Authentication
+
+The default mode of authentication into Brigade, that of a root user and
+password, is highly insecure and does not scale for a setup involving
```
krancour, line 289: I'm wary of the phrase "highly insecure" because, for some readers, it might not reflect well on us. I wonder if we could improve here by making an analogy to an operating system and stating that the root user, if enabled, has a broad set of permissions that aren't suitable for sharing.

docs/content/topics/operators/deploy.md
```diff
@@ -11,4 +11,429 @@ aliases:
...
+multiple users.
+
+For production-grade user authentication, Brigade ships with support for
+OIDC-compatible providers (such as [Google Identity Platform] and
+[Azure Active Directory]) and [GitHub].  In this doc, we'll demonstrate
```
krancour, line 294: Nit: GitHub doesn't implement OIDC. With GitHub, it's plain OAuth2. That said, we don't have general OAuth2 support. So choices boil down to OIDC (including the examples you mentioned) _or_ GitHub.

docs/content/topics/operators/deploy.md
```diff
@@ -11,4 +11,429 @@ aliases:
...
+    ## be in charge of assigning authorization permissions to authenticated
+    ## users.
+    ## Otherwise, permissions will need to be manually configured for each
+    ## authenticated user.
+    admins:
+      - <enter your name here>
+```
+
+Don't forget to add your GitHub username under the `admins` list, so that you
+as the operator have full admin privileges once logged in to Brigade.
```
krancour, line 363: These are granted on _first_ login. It might be worth nothing this so that no one expects that they can add/revoke admin permissions for existing users by modifying that configuration and re-deploying.

krancour:
@vdice should we add any of the following?

* HA MongoDB
* HA Artemis
* Changing default passwords
    * MongoDB
    * Artemis
    * root

docs/content/topics/operators/deploy.md
vdice: I suppose I agree that use of cert manager is a specific scenario.  I'd love to preserve some version of this demonstration of using cert manager (and specific ingress controller, etc.) as I found a similar doc useful for setting up Brigade v1 in a secure manner... what are your thoughts here?  Maybe save for a blog post?  Or another 'example setup' document?

docs/content/topics/operators/deploy.md
vdice: True, it is fairly specific... I just recall that document showing all the steps towards a full deployment (even if specific technologies are chosen) was helpful in deploying Brigade in the past (v1)... see also my question here: https://github.com/brigadecore/brigade/pull/1518#discussion_r682846156

vdice:
> @vdice should we add any of the following?

Good call!  I can add text around changing the default passwords.  How about a follow-up to add HA configuration?

krancour:
> How about a follow-up to add HA configuration?

That sounds good.

One thing we could mention now, at least, is to increase the volume sizes for the database and queue. In our dogfood cluster, I bumped them each to `40Gi`. (Unfortunately, I don't have much data yet on what happens if/when those fill up, so erring on the side of larger is probably well recommended for now.)

docs/content/topics/operators/deploy.md
krancour: Both with and without ingress, we have provided four options for TLS:

1. Disable TLS
1. Let the chart generate a self-signed cert
1. Paste the cert and key straight into values.yaml
1. Do none of the above, and just make sure the cert secret exists through other means. This could be creating it yourself or allowing cert-manager or similar to create it.

imho, we should cover these options, but should stop short of becoming a cert-manager how to. imho, those who are inclined to use cert-manager probably don't need our hand holding.

Also, this is NOT a hill I am willing to die on. So if you disagree with me, that's fine. Please don't consider this a blocker.

> Maybe save for a blog post? Or another 'example setup' document?

I do want to start working soon on a series of short videos that cover specific topics. I think one on production-grade deployment would be perfect and we could quite reasonably cover this (to a certain extent) in that video. We can start with the prep work already done-- "We've deployed Nginx Ingress and Cert Manager to this cluster already and we're going to be integrating our production grade deployment with those services."

wdyt?

vdice: Ok, agreed we should focus on discussion all configuration options in this doc rather than sticking to any one specific approach.  Will remove extended ingress controller and cert manager setup and replace with high-level listing of config opts...

vdice:
Got into a weird rebase/merge state so consolidated commits off of v2.

Next: updating relevant sections per https://github.com/brigadecore/brigade/pull/1518#discussion_r683764525

docs/content/topics/operators/deploy.md
vdice: @krancour PTAL at https://github.com/brigadecore/brigade/pull/1518/commits/f7f7e67d6a51625c33e608075c53548b39702aea when convenient.

docs/content/intro/install.md
```diff
@@ -101,6 +101,8 @@ $env:PATH+=";$env:USERPROFILE\bin"
 
 Now that you have the Brigade server installed, the next step is to install a Brigade [Gateway].
 
+## Install Brigade Gateways
+
```
krancour, line 105: Can add BitBucket and CloudEvents gateways to this list.

docs/content/topics/operators/deploy.md
```diff
@@ -11,4 +11,348 @@ aliases:
...
+[Ingress resource]: https://kubernetes.io/docs/concepts/services-networking/ingress/
+[Ingress Controller]: https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/
+
+## Third-Party Authentication
+
+The default mode of authentication into Brigade, that of a root user and
+password, is not suitable for a setup involving multiple users.
+
+For production-grade user authentication, Brigade ships with support for
+OIDC-compatible providers (such as [Google Identity Platform] and
```
krancour, line 166: I think the acronym OIDC is only well-known to people with prior OpenID Connect experience. It might be better to shun that acronym in favor of explicitly saying "OpenID Connect."

docs/content/topics/operators/deploy.md
```diff
@@ -11,4 +11,348 @@ aliases:
...
+[Google Identity Platform]: https://cloud.google.com/identity-platform
+[Azure Active Directory]: https://azure.microsoft.com/en-us/services/active-directory/
+[GitHub]: https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps
+
+### Configuring GitHub Authentication
+
+To set up GitHub authentication with Brigade, you'll create a [GitHub OAuth
+App] with the authorization callback URL set to point to the
+`/v2/session/auth` endpoint on Brigade's API server address, e.g. the hostname
+reserved when setting up ingress.  You'll then use the GitHub OAuth Apps's
```
krancour, line 179: ... or public IP if not using ingress.

docs/content/topics/operators/deploy.md
```diff
@@ -11,4 +11,348 @@ aliases:
...
+    ## authenticated user.
+    admins:
+      - <enter your name here>
+```
+
+Don't forget to add your GitHub username under the `admins` list, so that you
+as the operator have full admin privileges upon first login to Brigade.  Note:
+since these privileges are granted on first login, adding/revoking these
+permissions must be done via the brig CLI directly rather than re-deploying
+with differing configuration.
```
krancour, line 239: This section could probably be more clear.

Suggestion:

```
Note:
since these privileges are granted on first login, adding/revoking these
permissions for users who have already logged in for the first time must be done via the brig CLI directly rather than re-deploying
with differing configuration.
```

docs/content/topics/operators/deploy.md
krancour: 👍

krancour:
Added a few more comments-- all pretty minor.

This is close.

vdice:
Thanks @krancour.  Feedback should be addressed now.  PTAL when convenient.

krancour:
LGTM

krancour (approved):

krancour:
Merging before today's bug bash because:

1. I think everyone benefits from the updates docs in this PR
2. Maintainers previously agreed amongst themselves that not all checks need to run for doc-only PRs

Issue #1520, opened 2021-08-03
Brigade Developers Guide: A guide for people who want to modify Brigade's code.


vdice:

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1521

PR #1521, opened 2021-08-03
docs(developers.md): v2 updates

vdice:
* Updates the developers guide per v2

Closes https://github.com/brigadecore/brigade/issues/1520

### DIFF
--- a/docs/content/topics/developers.md
+++ b/docs/content/topics/developers.md
@@ -101,11 +101,8 @@ same results in CI.
 
 All development-related tasks should "just work" on Linux and Mac OS systems.
 When developing on Windows, the maintainers strongly recommend utilizing the
-Windows Subsystem for Linux.
-
-[This blog post](https://nickjanetakis.com/blog/setting-up-docker-for-windows-and-wsl-to-work-flawlessly)
-provides excellent guidance on making the Windows Subsystem for Linux work
-seamlessly with Docker Desktop (Docker for Windows).
+Windows Subsystem for Linux 2.  See more details
+[here](https://docs.docker.com/docker-for-windows/install/).
 
 ## Working with Go Code
 
--- a/docs/content/topics/developers.md
+++ b/docs/content/topics/developers.md
@@ -8,47 +8,49 @@ aliases:
   - /topics/developers.md
 ---
 
-TODO: update per v2
-
 # Developer Guide
 
 This document explains how to get started developing Brigade.
 
-Brigade is composed of numerous parts.  The following represent the core components:
+Brigade is composed of numerous parts.  The following represent the core
+components:
 
-- brigade-controller: The Kubernetes controller for delegating Brigade events
-- brigade-worker: The JavaScript runtime for executing `brigade.js` files. The
-  controller spawns these, though you can run one directly as well.
-- brigade-api: The REST API server for user interfaces
-- brigade-project: The Helm [chart][brigade-project-chart] for installing Brigade projects
-- brigade-vacuum: The stale build cleaner-upper (optional; enabled by default)
+- apiserver: Brigade's API server, where the majority of core Brigade
+  logic lives
+- scheduler: Schedules event workers and jobs on the underlying substrate
+- observer: Observes (and records) state changes in workers and jobs and
+  ultimately cleans them up from the underlying substrate
+- worker: The default runtime for executing `brigade.js/.ts` files.
+- logger: Event log aggregator, with Linux and Windows variants
 - brig: The Brigade CLI
-- git-sidecar: The code that runs as a sidecar in cluster to fetch Git repositories (optional; enabled by default)
+- git-initializer: The code that runs as a sidecar to fetch Git repositories
+  for vcs-enabled projects
 
-Additionally, there are several opt-in gateways that can be enabled via Helm chart values.  These are:
+Additionally, there are several opt-in gateways that can be deployed alongside
+Brigade.  These are:
 
- - Brigade GitHub App Gateway: The implementation of the GitHub App web hooks. It requires
-  the controller.
- - Generic Gateway: A generic gateway offering flexibility to create Brigade events from webhooks
-  originating from an arbitrary service/platform.
-- Container Registry Gateway: A gateway supporting container registry webhooks such as the ones emitted by
-DockerHub and ACR.
+ - Brigade GitHub Gateway: The implementation of GitHub App-based web hooks.
 
-Read up on all the gateways above, as well as others, in the [Gateways doc](./gateways.md).
+For more information around available gateways and developing your own,
+see the [Gateways doc](./gateways.md).
 
-This document covers environment setup, how to run functional tests and development of
-`brigade-controller` and `brigade-worker` components.
+This document covers environment setup, how to run tests and development of
+core components.
 
 ## Prerequisites
 
-- Minikube or [kind](https://github.com/kubernetes-sigs/kind) (running k8s versions 1.16.0+)
-- Docker
+- A local Kubernetes cluster, 1.16.0+.  We recommend [kind] or [minikube].
+- [Docker]
 - make
+
+[kind]: https://github.com/kubernetes-sigs/kind
+[minikube]: https://github.com/kubernetes/minikube
+[Docker]: https://www.docker.com/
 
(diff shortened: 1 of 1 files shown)


docs/content/topics/developers.md
```diff
@@ -8,47 +8,49 @@ aliases:
...
+- worker: The default runtime for executing `brigade.js/.ts` files.
+- logger: Event log aggregator, with Linux and Windows variants
 - brig: The Brigade CLI
-- git-sidecar: The code that runs as a sidecar in cluster to fetch Git repositories (optional; enabled by default)
+- git-initializer: The code that runs as a sidecar to fetch Git repositories
+  for vcs-enabled projects
 
-Additionally, there are several opt-in gateways that can be enabled via Helm chart values.  These are:
+Additionally, there are several opt-in gateways that can be deployed alongside
+Brigade.  These are:
```
krancour, line 30: We no longer include any gateways in the base repo / base chart.

docs/content/topics/developers.md
```diff
@@ -8,47 +8,49 @@ aliases:
...
-- Docker
+- A local Kubernetes cluster, 1.16.0+.  We recommend [kind] or [minikube].
+- [Docker]
 - make
 
+[kind]: https://github.com/kubernetes-sigs/kind
+[minikube]: https://github.com/kubernetes/minikube
+[Docker]: https://www.docker.com/
+
 ## Clone the Repository In GOPATH
```
krancour, line 50: I'm unsure of the extent to which this is still necessary if using a recent version of Go.

docs/content/topics/developers.md
```diff
@@ -61,12 +63,16 @@ $ git clone https://github.com/brigadecore/brigade $GOPATH/src/github.com/brigad
...
 
-**Note**: this leaves you at the tip of **master** in the repository where active development
-is happening. You might prefer to checkout the most recent stable tag:
+**Note**: this leaves you at the tip of **master** in the repository where
+active development is happening. You might prefer to checkout the most recent
+stable tag:
 
-- `$ git checkout v1.2.1`
+```console
+$ git checkout v1.2.1
```
krancour, line 71: Especially given that these are v2 docs, should we perhaps be directing would-be contributors to checkout the `v2` branch instead?

krancour:
> [This blog post](https://nickjanetakis.com/blog/setting-up-docker-for-windows-and-wsl-to-work-flawlessly)
provides excellent guidance on making the Windows Subsystem for Linux work
seamlessly with Docker Desktop (Docker for Windows).

This wasn't near enough a change that I was able to comment on it in-line, but I am questioning whether this is still relevant.

I don't have a Windows machine at my disposal. I do know @willie-yao has been doing development in WSL. Maybe we can check with him to see if this is still relevant advice.

docs/content/topics/developers.md
```diff
@@ -108,297 +114,232 @@ seamlessly with Docker Desktop (Docker for Windows).
...
-$ make test-js
+$ make test-unit-js
 ```
 
-To re-run JS dependency resolution:
+To clear the JS dependency cache:
 
 ```console
-$ make yarn-install
+$ make clean-js
```
krancour, line 142: I don't think we have such a target anymore.

docs/content/topics/developers.md
```diff
@@ -108,297 +114,232 @@ seamlessly with Docker Desktop (Docker for Windows).
...
 ```
 
-(See `Running the Brigade-Worker Locally` below for live testing against a running instance.)
-
 ## Building Source
```
krancour, line 145: I think this section probably needs to note that the builds for all Docker images occur using Kaniko in a container (because this is a convenient approach in CI where DinD is hard) and because these targets don't produce an image that's immediately available/usable locally, that the `hack-build-x` variants of those targets bypass the containerized Kaniko and build on the local system using plain Docker.

willie-yao:
> Maybe we can check with him to see if this is still relevant advice.

This blog post is out of date, as it is regarding WSL 1 and not WSL 2, which is what most people use now. I think it would be more helpful to link to this [tutorial](https://docs.docker.com/docker-for-windows/install/) from the official Docker docs.

docs/content/topics/developers.md
```diff
@@ -108,297 +114,232 @@ seamlessly with Docker Desktop (Docker for Windows).
...
-your `$KUBECONFIG` is pointing to that cluster, you can run `brigade-controller`
-locally.
+As the Brigade maintainers use kind heavily, there currently exists a helper
+script that will create a new kind cluster with a local private registry
+enabled, as well as setting up nfs as the local storage provisioner. To use
+this script, run:
 
 ```console
-$ ./bin/brigade-controller --kubeconfig $KUBECONFIG
+$ ./hack/kind/new-cluster.sh
```
krancour, line 281: Or `make hack-new-kind-cluster`.

docs/content/topics/developers.md
```diff
@@ -108,297 +114,232 @@ seamlessly with Docker Desktop (Docker for Windows).
...
 
-For the remainder of this document, we will assume that your local `$KUBECONFIG`
-is pointing to the correct cluster.
+```console
+$ export DOCKER_REGISTRY=localhost:5000
+$ make hack
+```
 
-### Running the Functional Tests
+To expose the apiserver port, run the following command:
```
krancour, line 292: If you use our hack script/target for creating a new kind cluster, we automatically map `localhost:31600` to the API server. Is it worth mentioning `make hack-expose-apiserver` as being a fallback if you're not running with our KinD configuration?

docs/content/topics/developers.md
vdice: It appears we do... but were you thinking of mentioning a different one here?

docs/content/topics/developers.md
vdice: Ok, removed this section.

vdice:
> This blog post is out of date, as it is regarding WSL 1 and not WSL 2, which is what most people use now. I think it would be more helpful to link to this tutorial from the official Docker docs.

Thanks @willie-yao , will reference the provided link!

docs/content/topics/developers.md
krancour: Ok. Just checked some Go docs and tried this for myself.

As long as one is using Go 1.16+ and `GO111MODULE` is _not_ explicitly set to `off` then there's no need to clone this on the GOPATH.

imho, we should just say that we require Docker (recommended) _or_ Go 1.16+ and leave it at that.

docs/content/topics/developers.md
krancour: My bad. Yes. This still exists.

docs/content/topics/developers.md
vdice: Updated in https://github.com/brigadecore/brigade/pull/1521/commits/668c6d963f1dd3c0e28f7520b39f7101de0319d6; PTAL

docs/content/topics/developers.md
```diff
@@ -8,65 +8,67 @@ aliases:
...
-$ export PATH=$GOPATH/bin:$PATH
-$ mkdir -p $GOPATH/src/github.com/brigadecore
-$ git clone https://github.com/brigadecore/brigade $GOPATH/src/github.com/brigadecore/brigade
-$ cd $GOPATH/src/github.com/brigadecore/brigade
+$ git checkout v2
 ```
 
-**Note**: this leaves you at the tip of **master** in the repository where active development
-is happening. You might prefer to checkout the most recent stable tag:
+You might prefer to checkout the most recent stable tag:
```
krancour, line 64: I can see the usefulness of this suggestion for people who's intention may be to build the latest release from source, but I also worry that this could lead to people submitting PRs that have a lot of merge conflicts with the tip of `v2`.

Any thoughts about just dropping this part?

krancour: This above is my only remaining feedback. Premptively LGTM'ing. Dealer's choice on whether/how to address this comment.

krancour (approved):
LGTM

docs/content/topics/developers.md
vdice: Good point; will drop.  Perhaps we can entertain adding again when we have stable 2.0.x tags out there (sans `beta`, etc.)

krancour:
Still LGTM

PR #1523, opened 2021-08-03
brig term

### LINKED ISSUE #1476: port brig term for brigade v2
Brigade 1.x included a `term` command that brought up a text-based, interactive dashboard. Especially in light of not having a Brigade 2 compatible Kashti yet, this feature should be ported to v2.

krancour:
Fixes #1476 

This has been a long-running collaboration between @willie-yao and myself to port `brig term` from v1.x to v2.

This PR is admittedly imperfect and still contains a number of TODOs, but between @willie-yao and myself, we reached a point where we felt this was useful enough in its present form to PR it so others can both use it and improve upon it.

Generally, the way this all works is we have created a custom, centralized UI component called the page router. It exposes functions that can bring a projects, project, event, job, or logs "page" into focus and begins the business of refreshing that page, while terminating auto-refresh of any previously in-focus page.

Better error handling is the the major, outstanding thing that still needs to be solved, as it wasn't immediately clear how best to surface errors in the UI. Expect that to be addressed in a follow-up.


v2/cli/term/log_page.go
```diff
@@ -0,0 +1,161 @@
...
+	apiClient core.APIClient,
+	app *tview.Application,
+	router *pageRouter,
+) *logPage {
+	l := &logPage{
+		page:    newPage(apiClient, app, router),
+		logText: tview.NewTextView().SetDynamicColors(true),
+	}
+
+	l.maxBytes = 65535
```
vdice, line 34: For this and the logic around this max (truncating text, etc.), did we just want to align with how we present log text in the GitHub gateway (GH limits check text bodies to this byte maximum)?  Otherwise, we should be fine to present a larger amount of logs sans truncation -- or introduce pagination, etc.

v2/cli/term/project_page.go
```diff
@@ -0,0 +1,328 @@
...
+		eventsContinueValues: []string{""}, // "" == continue value for first page
+		eventsTable:          tview.NewTable().SetSelectable(true, false),
+		usage:                tview.NewTextView().SetDynamicColors(true),
+	}
+	p.projectInfo.SetBorder(true).SetBorderColor(tcell.ColorWhite)
+	p.eventsTable.SetBorder(true).SetTitle(" Events ")
+	// Create the layout
+	p.page.Flex = tview.NewFlex().
+		SetDirection(tview.FlexRow).
+		AddItem(p.projectInfo, 0, 2, false).
```
vdice, line 47: It may be helpful to add a comment explaining arguments 2-4 for the `AddItem` function somewhere...

v2/cli/term/project_page.go
```diff
@@ -0,0 +1,328 @@
...
+	})
+
+}
+
+func (p *projectPage) fillProjectInfo(project core.Project) {
+	p.projectInfo.Clear()
+	p.projectInfo.SetTitle(fmt.Sprintf(" %s ", project.ID))
+	infoText := fmt.Sprintf("[grey]Description: [white]%s", project.Description)
+	if project.Spec.WorkerTemplate.Git != nil {
+		infoText = fmt.Sprintf("%s\n[grey]Git:", infoText)
```
vdice, line 131: Do we want to show other fields if non-empty?  (commit, ref, initSubmodules)?

v2/cli/term/event_page.go
```diff
@@ -0,0 +1,344 @@
...
+		).SetCell(
+			row,
+			ageCol,
+			&tview.TableCell{
+				Text:  created,
+				Align: tview.AlignLeft,
+				Color: color,
+			},
+		)
+		// TODO: Add age-- needs Job to track create time
```
vdice, line 299: Is this still a TODO?  I see the created time added to the ageCol above.

vdice (approved):
This looks great!  Really fun to use.  Will be great to broadcast this feature once we're ready to do so.

Only had comments/questions...

v2/cli/term/log_page.go
willie-yao: That is one reason. However, the main reason for implementing the circular buffer is that when there are a large amount of logs, we don't want to store a massive string in memory, which may cause memory issues.

vdice: Ah yes, thank you for the reminder.  So perhaps a follow-up could be to utilize the same circular buffer but instead of truncation, add pagination to the logs window (if needed).

krancour: The API only streams logs-- it doesn't have any paging capability. If we needed to add that, that's a relatively complex addition to the API. This isn't to say it cannot be done, but only that we should consider it only after the need is very clear.

v2/cli/term/project_page.go
krancour: Probably. We tried to limit ourselves to the most useful information since every UI puts a premium on screen real estate, but it's admittedly odd to show _some_ git details and not others.

v2/cli/term/event_page.go
krancour: Nice catch. This can be addressed now. Will do so.

vdice (approved):

PR #1530, opened 2021-08-09
update examples so all subscribe to events from cli

krancour:
Follow up to #1445 

Since #1445, several examples couldn't be triggered by the CLI because they weren't subscribed to any events from the CLI.

This PR fixes that.

### DIFF
--- a/examples/01-hello-world/project.yaml
+++ b/examples/01-hello-world/project.yaml
@@ -5,6 +5,10 @@ metadata:
   id: hello-world
 description: The simplest possible example
 spec:
+  eventSubscriptions:
+  - source: brigade.sh/cli
+    types:
+    - exec
   workerTemplate:
     logLevel: DEBUG
     defaultConfigFiles:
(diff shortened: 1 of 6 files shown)


examples/11-kitchen-sink/.brigade/brigade.ts
```diff
@@ -17,8 +17,9 @@ events.on("brigade.sh/cli", "exec", async event => {
 
   const job2Name = newJobName()
   let job2 = new Job(job2Name, "alpine:latest", event)
-  job2.primaryContainer.command = ["echo"]
-  job2.primaryContainer.arguments = [`Hello, ${job2Name}!`]
+  job2.primaryContainer.command = ["sleep"]
```
vdice, line 20: Were the changes in this file intended for this PR?

krancour: Brilliant catch. No. Not intended. This snuck in somehow because I was working on timeout fix stuff in parallel.

Will fix.

krancour: Fixed

vdice (approved):

PR #1550, opened 2021-08-09
build cli for arm64

### LINKED ISSUE #1546: brig CLI for arm64 architectures
Right now there are only binaries for x86 Macs and people with newer macs will be forced to use rosetta for things. It would be nice if there were builds for ARM Macs. No need to build the server for it I think, but the CLI would be useful

krancour:
Fixes #1546

### DIFF
--- a/.brigade/brigade.ts
+++ b/.brigade/brigade.ts
@@ -2,7 +2,7 @@ import { events, Event, Job, ConcurrentGroup, SerialGroup } from "@brigadecore/b
 
 const releaseTagRegex = /^refs\/tags\/(v[0-9]+(?:\.[0-9]+)*(?:\-.+)?)$/
 
-const goImg = "brigadecore/go-tools:v0.1.0"
+const goImg = "brigadecore/go-tools:v0.2.0"
 const jsImg = "node:12.3.1-stretch"
 const kanikoImg = "brigadecore/kaniko:v0.2.0"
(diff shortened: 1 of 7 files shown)


vdice (approved):

radu-matei (approved):

krancour:
So it looks like Go 1.16 cannot cross-compile win arm64.

Go 1.17 beta can.

I guess are choices are to use Go 1.17 beta or forget about arm64 for win.

Thoughts?

krancour:
Ok... despite claims that 1.17 can build win arm64, I can't actually get that to work.

I'm going to skip win arm64. I don't _think_ that's currently a widely used combo among developers.

krancour:
As it turns out, we were still using v0.1.0 of our go toolkit, which only includes go 1.15.6. That version cannot even cross-compile darwin arm64 as it turns out.

I cut a release of the go toolkit (v0.2.0) because the main branch there was already upgraded to go 1.16.5. I've amended this PR to use that newer toolkit for all go builds, tests, etc.

After that, I was able to further amend the PR to build:

* linux amd64
* darwin amd64
* windows amd64
* linux arm64
* darwin arm64

I still cannot get windows/arm64 to build, even using go 1.17. So be it.

@radu-matei @vdice asking for a re-review because of the additional changes to this PR.

vdice (approved):

PR #1553, opened 2021-08-09
fix(kind/new-cluster): prevent error when kind cluster already exists

### LINKED ISSUE #1549: docs & hack: make hack-new-kind-cluster
We should _name_ the kind cluster so that it won't conflict with an existing cluster having the default name "cluster."

adamreese:
Fixes #1549

### DIFF
--- a/hack/kind/new-cluster.sh
+++ b/hack/kind/new-cluster.sh
@@ -13,7 +13,7 @@ if [ "${running}" != 'true' ]; then
 fi
 
 # Create a kind cluster with the local Docker image registry enabled
-cat <<EOF | kind create cluster --config=-
+kind get clusters | grep brigade || cat <<EOF | kind create cluster --name brigade --config=-
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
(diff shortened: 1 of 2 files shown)


vdice (approved):
LGTM!  Just needs a [signed commit](https://github.com/brigadecore/community/blob/master/contributing.md#signed-commits) when convenient.

vdice (commented):
One thing to double-check: Can you run `DOCKER_REGISTRY=localhost:5000 make hack` to deploy to the newly-created kind cluster?  For some reason, although the images push to the localhost:5000 registry for me, the core component pods on the kind node hit ErrImagePull.  Not sure why that would be yet... (same flow appears to work on the v2 branch.)

hack/kind/new-cluster.sh
```diff
@@ -30,14 +30,16 @@ EOF
 
 # Make sure the local Docker image registry is connected to the same network
 # as the kind cluster
-docker network connect "kind" "${reg_name}" || true
+if ! docker network inspect kind >/dev/null 2>&1; then
```
vdice, line 33: I think this may be the added piece that may be tripping me up... for instance if the `kind` docker network persists but I have created a new kind cluster and registry... with this logic, the necessary `docker network connect` command would be skipped.

krancour: Ya... if it's a new registry and existing network, this leads to a problem.

krancour: I fixed this, signed the commit, and force pushed.

vdice (approved):

PR #1554, opened 2021-08-09
docs(install.md): add note around gateways

vdice:
This adds some clarifying text around gateways in the install doc.  @thomastaylor312 suggested the need for this.  @radu-matei also mentioned ngrok... maybe we'll get a full doc on this approach?!

We could also remove mention of gateways from the install doc entirely and only have users see/explore these options in the Topics section(s).  Thoughts?


thomastaylor312 (approved):
This is great! Much clearer expectations

radu-matei (approved):

PR #1555, opened 2021-08-09
Added brig init instructions to quickstart

### LINKED ISSUE #1547: Docs: Replace Add brig init to Quickstart
Replace the "Create A Project" section of the Quickstart guide with brig init, which is a more elegant way to create project files rather than downloading an example project

willie-yao:
Fixes #1547 

Adds instructions to quickstart for using `brig init` instead of downloading a sample project file.

### DIFF
--- a/docs/content/intro/quickstart.md
+++ b/docs/content/intro/quickstart.md
@@ -212,33 +212,30 @@ A Brigade [project] defines event handlers, such as the definition of a CI pipel
     apiVersion: brigade.sh/v2-beta
     kind: Project
     metadata:
-    id: first-project
+      id: first-project
     description: My new Brigade project
     spec:
-    ## Subscribe below to any events that should trigger your script.
-    ## The commented example depicts a subscription to "exec"
-    ## events originating from the Brigade CLI. 
-    eventSubscriptions:
-    - source: brigade.sh/cli
-        types:
-        - exec
+      eventSubscriptions:
+        - source: brigade.sh/cli
+          types:
+            - exec
     workerTemplate:
-        logLevel: DEBUG
-        defaultConfigFiles:
-        brigade.ts: |
-            import { events, Job } from "@brigadecore/brigadier"
-            
-            // Use events.on() to define how your script responds to different events. 
-            // The commented example below depicts handling of "exec" events originating 
-            // from the Brigade CLI.
-            
-            events.on("brigade.sh/cli", "exec", async event => {
-                let job = new Job("hello", "debian:latest", event)
-                job.primaryContainer.command = ["echo"]
-                job.primaryContainer.arguments = ["Hello, World!"]
-                await job.run()
-            })
+      logLevel: DEBUG
+      defaultConfigFiles:
+      brigade.ts: |
+        import { events, Job } from "@brigadecore/brigadier"
         
+        // Use events.on() to define how your script responds to different events. 
+        // The example below depicts handling of "exec" events originating from
+        // the Brigade CLI.
+        
+        events.on("brigade.sh/cli", "exec", async event => {
+            let job = new Job("hello", "debian:latest", event)
+            job.primaryContainer.command = ["echo"]
+            job.primaryContainer.arguments = ["Hello, World!"]
+            await job.run()
+        })
+
         events.process()
     ```
 
--- a/docs/content/intro/quickstart.md
+++ b/docs/content/intro/quickstart.md
@@ -200,19 +200,47 @@ brig login --insecure --server https://IP_ADDRESS --root
 ## Create a Project
 
 A Brigade [project] defines event handlers, such as the definition of a CI pipeline.
-In this example project, the handler prints a message using a string passed in the event payload.
 
 1. Initialize a new Brigade project with the `brig init` CLI command.
 
     ```
     brig init --id first-project
     ```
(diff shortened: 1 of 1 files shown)


docs/content/intro/quickstart.md
```diff
@@ -202,27 +202,22 @@ brig login --insecure --server https://IP_ADDRESS --root
 A Brigade [project] defines event handlers, such as the definition of a CI pipeline.
 In this example project, the handler prints a message using a string passed in the event payload.
```
krancour, line 203: I don't believe this is an accurate description of the project that `brig init` generates.

I would probably just cut this line entirely instead of rewording it.

docs/content/intro/quickstart.md
```diff
@@ -199,20 +199,48 @@ brig login --insecure --server https://IP_ADDRESS --root
 
 ## Create a Project
 
-A Brigade [project] defines event handlers, such as the definition of a CI pipeline.
-In this example project, the handler prints a message using a string passed in the event payload.
```
krancour, line 203: You removed too much here. Line 202 was good.

docs/content/intro/quickstart.md
```diff
@@ -199,20 +199,48 @@ brig login --insecure --server https://IP_ADDRESS --root
...
 
 1. Initialize a new Brigade project with the `brig init` CLI command.
 
     ```
     brig init --id first-project
     ```
+    This will create a project definition like the following in `.brigade/project.yaml`, which defines a handler for the "exec" event, that reads the event payload string and prints it out with "Hello, World!".
```
krancour, line 208: This description isn't accurate. The script that's generated doesn't do anything with the payload.

krancour: Which means some other commands farther on down might have to be edited as well, as they depict a payload being set, but it's not needed.

willie-yao: Oh right. All it does is it defines a handler for the exec event, and echoes hello world right?

krancour: Yes

docs/content/intro/quickstart.md
```diff
@@ -200,37 +200,60 @@ brig login --insecure --server https://IP_ADDRESS --root
...
 A Brigade [project] defines event handlers, such as the definition of a CI pipeline.
-In this example project, the handler prints a message using a string passed in the event payload.
 
-1. Download the example project to the current directory.
+1. Initialize a new Brigade project with the `brig init` CLI command.
 
-    **posix**
-    ```bash
-    curl -o project.yaml https://raw.githubusercontent.com/brigadecore/brigade/v2/examples/12-first-payload/project.yaml
     ```
```
krancour, line 206: Add the word "console" after the triple backticks.

docs/content/intro/quickstart.md
```diff
@@ -200,37 +200,60 @@ brig login --insecure --server https://IP_ADDRESS --root
...
-1. Open project.yaml
-
-    <script src="https://gist-it.appspot.com/https://raw.githubusercontent.com/brigadecore/brigade/v2/examples/12-first-payload/project.yaml"></script>
-
-    The project defines a handler for the "exec" event, that reads the event payload string and prints it out with "Hello, PAYLOAD!".
 
 1. Create the project in Brigade with the following command.
 
     ```
```
krancour, line 247: Add the word "console" after the triple backticks.

docs/content/intro/quickstart.md
```diff
@@ -200,37 +200,60 @@ brig login --insecure --server https://IP_ADDRESS --root
...
-    brig project create --file project.yaml
+    brig project create --file .brigade/project.yaml
     ```
 
 1. List the defined projects with `brig project list` and verify that you see your new project:
 
     ```console
     $ brig project list
```
krancour, line 254: We probably need to rip through _all_ the docs and settle on a single convention-- `$` before commands or _no_ `$` before commands, but for now, let's just be consistent within this one page. None of the other commands have the leading `$` on this page, so let's remove it here also.

krancour (approved):
LGTM

willie-yao:
@krancour There seems to be something wrong with the deployed version of the docs. It looks fine on GitHub so is there another way we should be formatting code blocks?

image

krancour:
@willie-yao interesting. Rescinded my LGTM until that's worked out.

vdice:
> @krancour There seems to be something wrong with the deployed version of the docs. It looks fine on GitHub so is there another way we should be formatting code blocks?

@willie-yao try indenting the contents of the yaml block.  Not sure why this is needed/different from the other block types (console, etc.), but it seemed to render appropriately when I tried it.

vdice:
@willie-yao , alternatively, I think the comment sections of the yaml (`# ....` lines) were messing up the yaml block formatting.  Once those were removed, it seemed to display appropriately even without the indenting.

willie-yao:
> I think the comment sections of the yaml (`# ....` lines) were messing up the yaml block formatting.

That would make sense. I think it will be fine to remove the comments since the quickstart explains what's going on in the file. Good catch!

willie-yao:
@vdice After removing the comments, the yaml will still not display correctly. The indenting works but makes the code block unaligned with the rest of the code blocks. I'll look more into this issue
image

vdice (approved):

PR #1560, opened 2021-08-11
add startup message to git initializer

### LINKED ISSUE #1529: git initializer should have similar startup message as other components
Knowing the precise semver and sha of a running component is a useful datapoint for any troubleshooting exercise.

Most v2 components print these details at startup, but the git initializer does not.

We should fix this.

krancour:
Fixes #1529

### DIFF
--- a/v2/git-initializer/main.go
+++ b/v2/git-initializer/main.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
 	"time"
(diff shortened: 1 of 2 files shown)


v2/git-initializer/main.go
```diff
@@ -30,6 +32,12 @@ const (
 )
 
 func main() {
+	log.Printf(
+		"Starting Brigade API Server -- version %s -- commit %s",
```
vdice, line 36:
```suggestion
		"Starting Brigade Git Initializer -- version %s -- commit %s",
```

krancour: D'oh! Copy pasta.

krancour: Fixed

vdice (approved):

PR #1564, opened 2021-08-12
explicitly disable user management functions when 3rd party authn is disabled

krancour:


v2/apiserver/internal/api/users.go
```diff
@@ -65,6 +65,14 @@ func (u UserList) MarshalJSON() ([]byte, error) {
 	)
 }
 
+// UsersServiceConfig encapsulates several configuration options for the USers
```
vdice, line 68:
```suggestion
// UsersServiceConfig encapsulates several configuration options for the Users
```

vdice (approved):

PR #1567, opened 2021-08-12
Fix wrong anchor

khaosdoctor:

In the docs, there's a link to a scripting tutorial, but it points back to the intro where the user just came from. This PR points the link to the right scripting guide.

**If applicable**:

### DIFF
--- a/docs/content/intro/quickstart.md
+++ b/docs/content/intro/quickstart.md
@@ -104,7 +104,7 @@ events.on("exec", () => {
 
 When the `exec` event is triggered, Brigade will create a Build based on this brigade.js file. This Build will create a single job that will start an image based on `alpine:3.8` which will simply do a couple of echoes. The Kubernetes Pod that will be created to run this Build job will have a name starting with 'do-nothing'. As you can probably guess, a new Kubernetes Secret will also be created to store information about this Build.
 
-Moreover, you can check out [this tutorial](https://docs.brigade.sh/topics/scripting/) for more on creating scripts.
+Moreover, you can check out [our scripting guide](https://docs.brigade.sh/topics/scripting/) for more on creating scripts.
 
 ### Running a Build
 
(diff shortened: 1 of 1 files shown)


khaosdoctor:
Fixed DCO

krancour:
Thanks @khaosdoctor !

krancour:
The test that's failing is entirely unrelated to this change.

@khaosdoctor in looking at this PR closer, I don't believe the original link was an error.

The intent here, as I understand it, was not to link to the comprehensive scripting guide; rather to link to a tutorial that, among other things, gives the user their first introduction to scripting in Brigade.

cc @vdice wdyt?

vdice:
> cc @vdice wdyt?

I'm not sure what the original intention was, but I think I agree with @khaosdoctor that it might make more sense to link to the scripting guide proper instead of back to the intro here.

krancour:
@vdice @khaosdoctor 

I'd be fine with linking to the scripting guide we think that makes more sense, but if we do that, imho, the wording should change in addition to the link:

> Moreover, you can check out [this tutorial](https://docs.brigade.sh/topics/scripting/) for more on creating scripts.

☝️  The scripting guide isn't organized like a _tutorial._

Perhaps something like the following works?

> Moreover, you can check out [our scripting guide](https://docs.brigade.sh/topics/scripting/) for more on creating scripts.

khaosdoctor:
> @vdice @khaosdoctor
> 
> I'd be fine with linking to the scripting guide we think that makes more sense, but if we do that, imho, the wording should change in addition to the link:
> 
> > Moreover, you can check out [this tutorial](https://docs.brigade.sh/topics/scripting/) for more on creating scripts.
> 
> ☝️ The scripting guide isn't organized like a _tutorial._
> 
> Perhaps something like the following works?
> 
> > Moreover, you can check out [our scripting guide](https://docs.brigade.sh/topics/scripting/) for more on creating scripts.

I like this idea too, let me create this change

krancour (approved):
Thanks @khaosdoctor!

LGTM

krancour:
Merging without CI as maintainers have agreed that doc changes don't require the full test suite to run.

Issue #1561, opened 2021-08-11
Gateways: Using and developing Brigade gateways


vdice:

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1569

PR #1569, opened 2021-08-13
docs(gateways.md): v2 updates

vdice:
* Updates the Gateways doc

Deploy preview here: https://deploy-preview-1569--brigade-docs.netlify.app/topics/operators/gateways/

Closes https://github.com/brigadecore/brigade/issues/1561


docs/content/intro/install.md
```diff
@@ -99,26 +99,26 @@ $env:PATH+=";$env:USERPROFILE\bin"
...
-## Install Brigade Gateways
-
-A Brigade [Gateway] generates events in Brigade in response to external triggers, for example a git push to repository.
-Brigade does not install gateways by default, which limits inbound events to those generated by the brig CLI,
-such as `brigade.sh/cli:exec` (the event source and type generated from the `brig event create` command.)
-
-Currently, the list of official Brigade v2 gateways that process external events is as follows:
+A Brigade [Gateway] generates events in Brigade in response to external
+triggers, for example a git push to repository. Brigade does not install
+gateways by default, which limits inbound events to those generated by the brig
```
krancour, line 113: This wording could trip up people who were used to gateways being included in Brigade 1.x that just needed to be enabled.

Could we reword to something like "Brigade does not ship with any gateways, which limits......... but compatible gateways are easy to install as add-ons" ?

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,518 @@ aliases:
 
 # Brigade Gateways
 
-This guide explains how gateways work, and provides guidance for creating your own
-gateway.
+This guide explains how gateways work, and provides guidance for creating your
+own gateway.
```
krancour, line 15: Total nit: "gateways" (plural) might read better here. _Total_ nit. Dealer's choice.

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,518 @@ aliases:
...
-
-These include the [Container Registry Gateway](dockerhub.md), the [Github Gateway](./github.md)
-and the [Generic Gateway](./genericgateway.md).  They can all be enabled via top-level
-Helm [chart flags](https://github.com/brigadecore/charts/blob/master/charts/brigade/values.yaml).
+While there is no default gateway that ships with Brigade, installing one
+alongside Brigade is as simple as a [Helm] chart install, along with any
+additional setup particular to a given gateway. See
+[below](#available-gateways) for a current listing of v2-compatible gateways.
 
 All of these provide HTTP-based listeners that receive incoming requests
```
krancour, line 29: Suggestion: "HTTP(S)-based" since all our gateways support TLS and have it enabled by default.

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,518 @@ aliases:
...
+(from Github or other platforms and systems) and generate Brigade events as a
+result.
 
 However, Brigade's gateway system works with more than just webhooks.
 
-For example, the `brig` client also acts as a gateway. When you execute a `brig run`
-command, `brig` creates a Brigade event. By default, it emits an `exec` event. And
-Brigade itself processes this event no differently than it processes the GitHub
-or container registry events.
+For example, the `brig` client also acts as a gateway. When you execute a
```
krancour, line 35: Willie and I had some back and forth over this with noisy neighbor-- which I elected not to count as a "gateway."

Unless I'm mistaken, in general (not Brigade, specifically) "gateways" are generally considered to be components that bridge / join / mediate between two systems...

I'm not sure if the CLI qualifies.

If you agree, this could perhaps be rectified by saying "the `brig` client behaves similarly to a gateway."

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,518 @@ aliases:
...
+Currently, the list of official Brigade v2 gateways that process external
+events is as follows:
 
-The most important thing to understand about a Brigade event is that it is simply
-a [Kubernetes Secret](https://kubernetes.io/docs/concepts/configuration/secret/)
-with special labels and data.
+* [Github Gateway](https://github.com/brigadecore/brigade-github-gateway)
+* [BitBucket Gateway](https://github.com/brigadecore/brigade-bitbucket-gateway/tree/v2)
+* [CloudEvents Gateway](https://github.com/brigadecore/brigade-cloudevents-gateway)
+* Noisy Neighbor?
```
krancour, line 58: See https://github.com/brigadecore/brigade/pull/1569/files#r690545417.

I think I would remove this from the list.

I've been counting it in a separate category.

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,518 @@ aliases:
...
+    vi. The event `source`
+    vii. The event `type`
+  2. The `worker.spec` section, which contains worker configuration inherited
+    from the project definition associated with the event in combination with
+    system-level defaults.
+  3. The `worker.status` section, which contains the `started` and `ended`
+    timestamps and current `phase` of the worker handling the event. In the
+    example above, it has already reached the terminal phase of `SUCCEEDED`.
+
+To explore the current SDK definitions of an Event object, see the
```
krancour, line 118: Nit: I'd drop "current" because it subtly suggests that things are subject to change, but we're avoiding breaking changes at all costs now.

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,518 @@ aliases:
...
-    payload: $("${base64[@]}" <<<"${payload}")
-    script: $("${base64[@]}" <<<"${script}")
-EOF
-  sleep 60
-done
-```
+Since the Brigade API server is the point of contact for gateway
+authentication/authorization and event submission, gateway developers will need
+to pick an [SDK] to use. For this example, we'll be using the [Go SDK]. As of
+this writing, a [Javascript/Typescript SDK] and a [Rust SDK] also exist.
```
krancour, line 135: The Rust SDK is WIP and I don't think it's been touched in a long while. We should perhaps either note that it's WIP or else not mention it.

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,518 @@ aliases:
...
+    apiClientOpts,
+  )
+
+  // Create a root session, which returns an auth token
+  token, err := authClient.CreateRootSession(ctx, rootPassword)
+  if err != nil {
+    log.Fatal(err)
+  }
+  tokenStr := token.Value
+}
```
krancour, line 202: I don't think this is a pattern I'd want people to copy (and people will). It might be better to show commands to generate a service account and source the service account's token from an env var.

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,518 @@ aliases:
...
+  ctx := context.Background()
+
+  // Construct a Brigade Event
+  event := core.Event{
+    // This is the ID/name of the project that the event will be intended for
+    ProjectID: projectID,
+    // This is the source value for this event
+    Source:    "brigade.sh/example-gateway",
+    // This is the event's type
+    Type:      "create-event",
```
krancour, line 228: With the payload being "Dolly," you don't want to make this a "hello" event?? 😉

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,518 @@ aliases:
...
+```go
+// createEvent creates a Brigade Event for the provided project, using the
+// provided sdk.Client
+func createEvent(client sdk.APIClient, projectID string) error {
+  ctx := context.Background()
+
+  // Construct a Brigade Event
+  event := core.Event{
+    // This is the ID/name of the project that the event will be intended for
+    ProjectID: projectID,
```
krancour, line 224: Do we want to show this? Calling out a specific project by ID undermines the subscription model a bit. It's not that this isn't legit or doesn't work-- it certainly is and certainly does-- but is it the "textbook use" we want to show?

(This is an actual question. I don't know. Thinking out loud.)

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,518 @@ aliases:
...
+`Source` and `Type`. As a bonus, we're also adding a `Payload`. That's just the
+start of what a Brigade Event can contain, however. Other notable fields worth
+researching are:
 
-And we can pack that into a Docker image by running `docker build -t technosophos/example-cron:latest .`.
-You should replace `technosophos` with your Dockerhub username (or modify the
-above to store in your Docker registry of choice).
+- `Qualifiers`: A list of qualifier values. For a project to receive events
+  from a gateway, the qualifiers on an event must exactly match the qualifiers
+  set on the project's event subscription, when provided.
```
krancour, line 284: Nit and dealer's choice.

This explanation might be more clear if we flip it:

> For a project to receive an event, the qualifiers on an event's subscription must exactly match the qualifiers
  on the event (in addition to matching source and type).

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,518 @@ aliases:
...
 
-Then push the image to a repository that your Kubernetes cluster can reach:
+- `Labels`: A list of labels. Projects can choose to utilize these for
+  filtering purposes. In contrast to qualifiers, project event subscription
+  labels don't need to exactly match event labels in order to receive them, as
+  long as the event does have any/all labels that the project might use for
+  filtering.
```
krancour, line 290: Additional clarification might be in order? Something like:

> In contrast to qualifiers, a project's event subscriptions does not need to match an event's labels in order to receive it. Labels, however, can be used to _narrow_ an event subscription by optionally selecting only events that labeled a particular way.

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,518 @@ aliases:
...
 
+- `ShortTitle`: A short title for the event.
+
+- `LongTitle`: A longer, more descriptive title for the event.
+
+- `SourceState`: A key/value map representing event state that can be persisted
+  by the Brigade API server so that gateways can track event handling progress
+  and perform other actions, such as updating upstream services.
+
+### Linking it all together
```
krancour, line 300: I think if we source the API token from and env var as mentioned in another comment, and also perhaps omit projectID, this program gets even smaller...

I'm wondering... if we get it small enough, can we present it all in one shot instead of building part of it and then revisiting it?

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,518 @@ aliases:
...
+    types:
+      - create-event
+  workerTemplate:
+    logLevel: DEBUG
+    defaultConfigFiles:
+      brigade.ts: |
+        import { events } from "@brigadecore/brigadier"
+
+        events.on("brigade.sh/example-gateway", "create-event", async event => {
+          console.log("Hello, " + event.payload + "!")
```
krancour, line 454: There it is. "Hello, Dolly!" 😂

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,518 @@ aliases:
...
+  "log"
 
-payload=$(date)
-script=$(cat <<EOF
-const { events } = require("brigadier");
-events.on("my_event", (e) => {
-  console.log("The system time is " + e.payload);
-});
-EOF
+  brigadeOS "github.com/brigadecore/brigade-foundations/os"
```
krancour, line 163: imho, I wouldn't show use of brigade-foundations.

If all we're doing is getting some env vars, I'd worry this is more likely to distract from the important bits than it is to save anyone much work.

Use `os` from stdlib?

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,518 @@ aliases:
...
+
+Now that we have a project subscribing to events from this gateway, we're ready
+to build and run the example gateway!
+
+First, we'll need to take care of a few bootstrapping items needed for our Go
+program. Here we initialize this program's Go modules file needed for tracking
+dependencies. Then, we fetch the needed deps.
+
+```console
+$ go mod init
```
krancour, line 477: What would you think about moving this command and `go get github.com/brigadecore/brigade/sdk/v2` to the start of the example before we start coding?

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,518 @@ aliases:
...
+go: creating new go.mod: module github.com/brigadecore/brigade/examples/gateways/example-gateway
+go: to add module requirements and sums:
+	go mod tidy
+
+$ go mod tidy
```
krancour, line 482: Can this command potentially be omitted?

iirc, most go commands, including `go run` will automatically perform dependency resolution.

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,518 @@ aliases:
...
-Whenever you are done with this example, you can delete it with
-`kubectl delete cronjob example-cron-gateway`.
+Hopefully this brief guide showing a sample gateway written using Brigade's Go
+SDK was helpful. All of the sample code can be found in the
+[examples/gateways/example-gateway] directory.
 
-That's it! We have create both a local shell script gateway and an in-cluster
-cron gateway.
+We look forward to seeing the Brigade Gateway ecosystem expand with
+contributions from readers like you!
```
krancour, line 526: ❤️

docs/content/topics/operators/gateways.md
vdice: `go run` didn't do it for me (go 1.16.6) but I've replaced `go mod tidy` w/ the `go get github.com/brigadecore/brigade/sdk/v2` line since that's now the only dep for this program -- as it should be!

docs/content/topics/operators/gateways.md
vdice: Good call; removed.

docs/content/topics/operators/gateways.md
vdice: Yuuup, consolidated into one section now.  Thanks for the simplifications.

docs/content/topics/operators/gateways.md
vdice: Removed here and added a mention in the section where we go over other fields not used.

docs/content/topics/operators/gateways.md
vdice: Yes!  Good call.

docs/content/topics/operators/gateways.md
vdice: Good call.  Updated!

docs/content/topics/operators/gateways.md
vdice: Thanks, just left it at `... for creating your own.`

krancour: I like that ever better. Nice.

docs/content/topics/operators/gateways.md
krancour: Admittedly, I am usually doing this after the fact as well...

Wondering now though if we can get away with omitting go mod stuff entirely. In the absence of go.mod, `go get` should yank packages down to the gopath where they're easily found/used.

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,421 @@ aliases:
...
-scripts run as a response to one or more events. In Brigade, a _gateway_ is an
-entity that generates events. Often times, it translates some external trigger
-into a Brigade event.
+The [Brigade architecture](/topics/design) is oriented around the concept that
+Brigade scripts run as a response to one or more events. In Brigade, a gateway
+is an entity that generates events. Often times, it translates some external
+trigger into a Brigade event.
 
-Brigade ships with the ability to enable various gateways that are ready to go.
+While there is no default gateway that ships with Brigade, installing one
```
krancour, line 24: I'm slightly worried that "default gateway" (singular) and "installing one" leave this sentence open to misinterpretation-- that gateways exist in a 1:1 with a Brigade instance.

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,421 @@ aliases:
...
 ## What Is A Brigade Gateway?
 
-The [Brigade architecture](design.md) is oriented around the concept that Brigade
-scripts run as a response to one or more events. In Brigade, a _gateway_ is an
-entity that generates events. Often times, it translates some external trigger
-into a Brigade event.
+The [Brigade architecture](/topics/design) is oriented around the concept that
+Brigade scripts run as a response to one or more events. In Brigade, a gateway
+is an entity that generates events. Often times, it translates some external
+trigger into a Brigade event.
```
krancour, line 22: Any thoughts about backing off from the imprecise "often?" With a slightly narrower definition of gateway, this is _always_ true.

Suggestion:

> The [Brigade architecture](/topics/design) is oriented around the concept that Brigade scripts run as a response to one or more events. In Brigade, a gateway is a component that receives triggers from some external system and translates them into Brigade events.

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,421 @@ aliases:
...
-Helm [chart flags](https://github.com/brigadecore/charts/blob/master/charts/brigade/values.yaml).
-
-All of these provide HTTP-based listeners that receive incoming requests
-(from a container registry, Github or other platforms and systems) and generate
-Brigade events as a result.
+All of these provide HTTP(S)-based listeners that receive incoming requests
+(from Github or other platforms and systems) and generate Brigade events as a
+result.
 
 However, Brigade's gateway system works with more than just webhooks.
```
krancour, line 33: This could trip people up a little because the word webhook is used for the first time here without any proper introduction. I'm not sure we can take it for granted that all readers will correctly interpret "webhook" has a trigger that arrived over HTTP(S).

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,421 @@ aliases:
...
 
-For example, the `brig` client also acts as a gateway. When you execute a `brig run`
-command, `brig` creates a Brigade event. By default, it emits an `exec` event. And
-Brigade itself processes this event no differently than it processes the GitHub
-or container registry events.
+For example, the `brig` client behaves similarly to a gateway. When you execute
+a `brig event create` command, `brig` creates a Brigade event of source
+`brigade.sh/cli` and of type `exec`. Brigade itself processes this event no
+differently than it processes events from gateways tracking external activity.
```
krancour, line 38: I like the changes to this section, but now, considering the section as a whole in a broader context, I'm wondering if it maybe doesn't belong.

This paragraph seems to shift the conversation from "what is a gateway?" to "what is an event?"

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,421 @@ aliases:
...
 
 There are no rules about what can be used as a trigger for an event. One could
-write a gateway that listens on a message queue, or runs as a chat bot, or watches
-files on a filesystem... any of these could be used to trigger a new Brigade event.
+write a gateway that listens on a message queue, or runs as a chat bot, or
+watches files on a filesystem... any of these could be used to trigger a new
+Brigade event.
```
krancour, line 43: I really like this section. I'm wondering if it should maybe move up higher (because I think it's super important) and potentially replace the bit about our (current) gateways being HTTP(S) based...

I think the fact that gateways could hypothetically receive triggers through a variety of means trumps the fact that the ones we happen to have implemented so far use HTTP(S)...

Or maybe we could word something like this?

> Currently, all of our officially supported gateways are triggered by webhooks delivered from external systems over HTTP(S), but there are no rules about what can be used as a trigger for an event. A gateway could hypothetically listen on a message queue, run as a chat bot, or watch files on a file system.

krancour: This wording could also solve the problem of the term "webhooks" not being well-introduced, as mentioned in another comment.

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,421 @@ aliases:
...
 
-### The Anatomy of a Brigade Event Secret
+A Brigade Event is defined primarily by its source and type values, worker
+configuration and worker status.
 
-Here is the structure of a Brigade event secret. It is annotated to explain what
-data belongs in what fields.
+Here is a YAML representation of an event created via the `brig event create`
+command for the [01-hello-world sample project].
```
krancour, line 74: I wonder if we should explicitly state that many of the fields shown are system-generated.

Or maybe we can present an abridged YAML representation (and explicitly acknowledge that it is abridged) that focused only on the fields that can be set by a client (gateway in this case).

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,421 @@ aliases:
...
-const { events } = require("brigadier");
-events.on("my_event", (e) => {
-  console.log("The system time is " + e.payload);
-});
-EOF
-)
+Given the above description of how gateways work, we can walk through how a
+minimal example gateway can be built. We'll focus on the event creation side
+of a Brigade gateway, rather than going over other common attributes, such as
+an http(s) server that awaits external webhook events.
```
krancour, line 131: Nit: Elsewhere in this doc, http(s) is capitalized. I don't have a preference for upper/lowercase, but we should be internally consistent.

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,421 @@ aliases:
...
-#!/usr/bin/env bash
-set -euo pipefail
+#### Service Account creation
 
-# The Kubernetes namespace in which Brigade is running.
-namespace=${NAMESPACE:-default}
+All Brigade gateways require a service account token for authenticating with
+Brigade when submitting an event into the system. As preparation then, we'll
+create a service account for this gateway and save the generated token for use
+in our program.
```
krancour, line 159: ❤️  this section.

Nit: Suggest striking "then" from "As preparation then"

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,421 @@ aliases:
...
-  build_id: $(base64 -w 0 <<<"${uuid}")
-  payload: $(base64 -w 0 <<<"${payload}")
-  script: $(base64 -w 0 <<<"${script}")
-EOF
+Authorize this service account to create new events:
+
+```console
+$ brig role grant EVENT_CREATOR \
+    --service-account example-gateway \
+    --source brigade.sh/example-gateway
```
krancour, line 175: The rule of thumb to avoid source clashes is to use a URI you control. This means leading with one's own domain or the URL for something else one owns, like the URL for a github repo, for instance.

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,421 @@ aliases:
...
+	// If the returned events list has no items, no event was created
+	if len(events.Items) != 1 {
+		fmt.Println("No event was created.")
+		return nil
+	}
+
+	// The Brigade event was successfully created!
+	fmt.Printf("Event created with ID %s\n", events.Items[0].ID)
+	return nil
+}
```
krancour, line 289: imho, the example is easier to follow if all of this is inline in the `main` function.

docs/content/topics/operators/gateways.md
```diff
@@ -11,340 +11,421 @@ aliases:
...
-    payload: $("${base64[@]}" <<<"${payload}")
-    script: $("${base64[@]}" <<<"${script}")
-EOF
-  sleep 60
-done
-```
+Since the Brigade API server is the point of contact for gateway
+authentication/authorization and event submission, gateway developers will need
+to pick an [SDK] to use. For this example, we'll be using the [Go SDK]. As of
+this writing, a [Javascript/Typescript SDK] and a [Rust SDK] (WIP) also exist.
```
krancour, line 136: Tiniest of nits: WIP might not be as broadly understood as we assume it is. It could be better to say "work in progress."

Again... this is the tiniest of nits. Dealer's choice entirely.

docs/content/topics/operators/gateways.md
krancour: > go get should yank packages down to the gopath where they're easily found/used

Well... I was wrong about that.

But here's what I did figure out.

We can condense the instructions to just this:

```bash
$ mkdir example-gateway
$ cd example-gateway
$ go mod init example-gateway
$ go get github.com/brigadecore/brigade/sdk/v2
# code, code, code
$ go run main.go
```

docs/content/topics/operators/gateways.md
vdice: 👍 So I take it we'd like to add a mention here around the naming convention of the source value?

krancour: We probably should, but at the very least, we probably shouldn't be instructing people to use _our_ domain name as a component of their source string.

vdice: Oh gotcha - yup.

vdice: Note added (and example uri updated).

docs/content/topics/operators/gateways.md
vdice: Added additional text around this... WDYT?

docs/content/topics/operators/gateways.md
vdice: Updated and re-org'd... PTAL when convenient.

docs/content/topics/operators/gateways.md
vdice: Good point.  Removed this section.

docs/content/topics/operators/gateways.md
vdice: Removed this line.

vdice:
Thanks @krancour -- I _think_ I've addressed all feedback in the latest commit.

docs/content/topics/operators/gateways.md
vdice: Nice, thanks!  Incorporated.

examples/gateways/example-gateway/main.go
```diff
@@ -0,0 +1,66 @@
...
+
+func main() {
+	ctx := context.Background()
+
+	// Get the Brigade API server address and gateway token from the environment
+	apiServerAddress := os.Getenv("APISERVER_ADDRESS")
+	if apiServerAddress == "" {
+		log.Fatalf("Required environment variable APISERVER_ADDRESS not found.\n")
+	}
+	gatewayToken := os.Getenv("GATEWAY_TOKEN")
```
krancour, line 22: imho, `GATEWAY_TOKEN` has the potential to be misleading because it could be interpreted as a token for authenticating _to_ the gateway rather than a token used _by_ the gateway to authenticate to the Brigade API server.

Any thoughts on changing to `API_TOKEN` or something of the sort?

krancour: Apart from this, LGTM.

Dealer's choice on whether to act on this or merge.

krancour (approved):

examples/gateways/example-gateway/main.go
vdice: Agreed; updated.  Thanks!

krancour:
@vdice really happy with how all this turned out. This will be _really_ useful for anyone looking to build integrations.

Issue #1570, opened 2021-08-16
Authentication: Authentication strategies in Brigade


vdice:

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1571

PR #1571, opened 2021-08-16
docs(authentication.md): add authentication doc

vdice:
* Adds authentication doc

Preview: https://deploy-preview-1571--brigade-docs.netlify.app/topics/administrators/authentication/

Closes https://github.com/brigadecore/brigade/issues/1570

### DIFF
--- a/docs/content/topics/administrators/authentication.md
+++ b/docs/content/topics/administrators/authentication.md
@@ -22,9 +22,9 @@ The motivation behind this design stems from one of the core guiding principles
 in Brigade 2: Provide an abstraction between users and the underlying substrate
 (Kubernetes), such that users of Brigade do not need to interact with (or have
 any presence on) the substrate itself. Therefore, the need to source user
-identity from another authority became apparent. Rather than implementing such
-a system in Brigade itself, we've decided to utilize well-known external
-services to fulfill this need.
+identity from another authority became apparent. Rather than implement such a
+system in Brigade itself, we decided to utilize well-known external services
+to fulfill this need.
 
 ## Third-Party Authentication Options
 
@@ -103,13 +103,14 @@ App] with the authorization callback URL set to point to the
 `/v2/session/auth` endpoint on Brigade's API server address, e.g. the hostname
 reserved when setting up ingress or the external IP address if no ingress is
 used.  You'll then use the GitHub OAuth Apps's client ID and a generated client
-secret in the `brigade2-values.yaml` file.
+secret in the [values.yaml] file for the Brigade Helm chart.
 
   1. Follow the [GitHub OAuth App] creation instructions, supplying your choice
     of values for `Application Name`, `Homepage URL` and `Application description`.
   1. For the `Authorization callback URL`, you'll supply the a value based
     on the DNS hostname and path mentioned above.  For example, if the DNS
-    hostname is `mybrigade.example.com`, the value would be:
+    hostname for the Brigade API server is `mybrigade.example.com`, the value
+    would be:
       ```
       https://mybrigade.example.com/v2/session/auth
       ```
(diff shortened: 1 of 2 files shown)


docs/content/topics/administrators/authentication.md
```diff
@@ -11,4 +11,77 @@ aliases:
 
 # Brigade Authentication
 
-TODO
+Brigade utilizes third-party providers for its user authentication strategy.
+We'll go over the available options and their configuration in this document.
+
+Note that Brigade's default mode of root user access precludes functionality
+related to user management or authorization. Therefore, selection of a viable
+third-party auth strategy (and thus disabling root user mode) is a
```
vdice, line 19: Actually, looking for thoughts from @krancour here.  I had always assumed a binary root user mode OR third-party auth, but there isn't anything in the Brigade chart preventing deployment of both modes enabled.  If that's how we'd like it to continue to be, I can remove the `(and thus disabling root user mode)` as that wouldn't strictly be necessary to use third-party auth.

docs/content/topics/administrators/authentication.md
```diff
@@ -11,4 +11,77 @@ aliases:
...
+Note: since these privileges are granted on first login, adding/revoking these
+permissions for users who have already logged in for the first time must be
+done via the brig CLI directly rather than re-deploying with differing
+configuration.
+
+[values.yaml]: https://github.com/brigadecore/brigade/blob/v2/charts/brigade/values.yaml
+
+### GitHub OAuth Provider
+
+To use [GitHub's Oauth Provider], please see the
```
vdice, line 83: A few other options here; I could go either way.  We could duplicate the GitHub OAuth setup instructions here (but then have two areas to maintain/update) or we could put the full instructions here and move the link to the Deploy doc.  Or, of course, leave as is.

docs/content/topics/administrators/authentication.md
krancour: You can do both. It's like Linux. root is a thing, but you shouldn't share it-- which is where the 3rd party auth comes in.

docs/content/topics/administrators/authentication.md
```diff
@@ -11,4 +11,77 @@ aliases:
 
 # Brigade Authentication
 
-TODO
+Brigade utilizes third-party providers for its user authentication strategy.
```
krancour, line 14: Do you think it's worthwhile to explain the rationale for this? I could go either way.

The rationale is that with Brigade 2 providing an abstraction between users and the underlying substrate (Kubernetes), it cannot be taken for granted that every user of Brigade is also a Kubernetes user. That introduces a need for sourcing user identity from somewhere else.

docs/content/topics/administrators/authentication.md
krancour: imho, the best option is to move those instructions to here and link to it from the deploy doc.

krancour:
Overall, looks good.

docs/content/topics/administrators/authentication.md
vdice: Good call, I'd been taking that point for granted, definitely worth a mention.  Will add.

krancour: My wording was quick and dirty tho. I'm sure you can improve on it.

vdice: Ok, updated.  PTAL when convenient.  Thanks!

docs/content/topics/administrators/authentication.md
```diff
@@ -11,4 +11,154 @@ aliases:
...
+enabling the suite of user authorization commands.
+## Background
+
+The motivation behind this design stems from one of the core guiding principles
+in Brigade 2: Provide an abstraction between users and the underlying substrate
+(Kubernetes), such that users of Brigade do not need to interact with (or have
+any presence on) the substrate itself. Therefore, the need to source user
+identity from another authority became apparent. Rather than implementing such
+a system in Brigade itself, we've decided to utilize well-known external
+services to fulfill this need.
```
krancour, line 27: ❤️   Great explanation.

krancour (approved):
LGTM!

vdice:
Apologies, @krancour , I noticed an opportunity for a few more tweaks after your LGTM.  Please see https://github.com/brigadecore/brigade/pull/1571/commits/0cd3a29b1172ac94a34a53a4a04c08a9b8bb6c3a

krancour:
@vdice. No apologies necessary. Still LGTM.

PR #1578, opened 2021-08-23
don't assume a repo's default branch is named master

### LINKED ISSUE #1576: Bug: Don't assume a project's default branch is master
Do not confuse this issue with #1542. They're different.

Currently, when new events are created, logic in the Brigade API server sets the git ref to `refs/heads/master` if it was unspecified (empty string) by both the event and the project's worker spec from which the event's worker spec is inherited.

It's no longer a safe assumption that projects' default branches are master. (Was that ever really a safe assumption?)

Instead of setting the git ref to `refs/heads/master` when it was originally unspecified, we should _leave it_ unspecified and modify the git initializer component to try to make a real-time determination as to what the project's default branch name is.

krancour:
Fixes #1576

Formerly, if we had neither a commit sha or a reference, we defaulted to "refs/heads/master".

That's no longer a good assumption (really it never was).

Now we don't do that. We allow both sha _and_ commit to be an empty string and the git initializer is updated to know how to deal with that scenario. It handles that scenario _separately_ without change to existing logic-- although some existing logic does move to helper functions for improved clarity.

__EDIT:__ Worth noting: I did explore an alternate approach where instead of following a "simple" vs "complex " checkout path, the logic for listing references and building refspecs was different when both commit and ref were empty strings. (In that case, when you have nothing else to go on, you have to look for the HEAD symbolic ref, specifically, instead of ignoring it.) imho, this turned out to be much more convoluted because there were a lot more conditionals involved compared to the one conditional required with the approach I opted for.


sdk/v2/core/events.go
```diff
@@ -197,10 +197,14 @@ type GitDetails struct {
...
+	// Commit specifies a revision (by SHA) to be checked out. If non-empty, this
+	// field takes precedence over any value in the Ref field.
 	Commit string `json:"commit,omitempty"`
-	// Ref specifies a tag or branch to be checked out. If left blank, this will
-	// default to "master" at runtime.
+	// Ref is a symbolic reference to a revision to be checked out. If non-empty,
+	// the value of the Commit field supercedes any value in this field. Example
+	// uses of this field include referencing a branch (refs/heads/<branch name>)
+	// or a tag (refs/tags/<tag name>). If left blank, this field is interpreted
+	// as a reference to the repository's default branch.
```
krancour, line 207: Saw an opportunity here to improve these comments further while I was already touching them. Did the same in a few other places as well.

v2/apiserver/internal/api/events.go
```diff
@@ -453,14 +457,6 @@ func (e *eventsService) createSingleEvent(
...
 	}
 
-	// If no commit (sha) or ref (branch or tag) is specified, default to the
-	// master branch
-	if workerSpec.Git != nil {
-		if workerSpec.Git.Commit == "" && workerSpec.Git.Ref == "" {
-			workerSpec.Git.Ref = "refs/heads/master"
-		}
-	}
-
```
krancour, line 463: Very important-- this means it's now possible for commit _and_ ref to both be empty strings.

The initializer logic is updated accordingly to account for that.

v2/git-initializer/main.go
```diff
@@ -102,10 +102,62 @@ func gitCheckout() error {
 		auth = publicKeys
 	}
 
+	var worktree *git.Worktree
+	if event.Worker.Git.Commit == "" && event.Worker.Git.Ref == "" {
+		worktree, err = simpleCheckout(gitConfig.CloneURL, auth)
```
krancour, line 107: If we had no commit and no ref-- absolutely nothing to go on-- then a simple clone gets us what we want-- the default branch, regardless of what's named.

v2/git-initializer/main.go
```diff
@@ -102,10 +102,62 @@ func gitCheckout() error {
...
+	} else {
+		worktree, err = complexCheckout(
+			gitConfig.CloneURL,
+			event.Worker.Git.Commit,
+			event.Worker.Git.Ref,
+			auth,
+		)
```
krancour, line 114: The more complex scenario where we need to do some legwork to figure out if a ref was a fulls ref like `/refs/heads/foo` or a more "casual" ref like `foo` is accounted for by `complexCheckout()`. The logic that used to be right here in this function just moved down to there.

v2/git-initializer/main.go
```diff
@@ -102,10 +102,62 @@ func gitCheckout() error {
...
+	}
+	if err != nil {
+		return err
+	}
+
+	// Initialize submodules if configured to do so
+	if event.Worker.Git.InitSubmodules {
+		err = initSubmodules(worktree)
+	}
```
krancour, line 123: It doesn't matter whether the simple or complex scenario played out up to this point. Again, some logic that used to be here, just moved.

v2/git-initializer/main.go
```diff
@@ -102,25 +103,77 @@ func gitCheckout() error {
...
+	return err
+}
+
+func simpleCheckout(
+	cloneURL string,
+	auth transport.AuthMethod,
+) (*git.Worktree, error) {
+	// This will reliably get the whichever branch is designated as the default.
```
vdice, line 132:
```suggestion
	// This will reliably get whichever branch is designated as the default.
```

vdice (approved):
LGTM!

Issue #1579, opened 2021-08-24
Storage: Storage options and configuration for Brigade


vdice:

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1586

PR #1586, opened 2021-08-26
docs(storage.md): v2 updates

vdice:
* Updates the storage doc per v2

Preview: https://deploy-preview-1586--brigade-docs.netlify.app/topics/operators/storage/

Closes https://github.com/brigadecore/brigade/issues/1579

### DIFF
--- a/docs/content/topics/operators/storage.md
+++ b/docs/content/topics/operators/storage.md
@@ -1,6 +1,6 @@
 ---
 title: Storage
-description: 'How Brigade uses Kubernetes Persistent Storage.'
+description: Storage and configuration in Brigade
 section: operators
 weight: 3
 aliases:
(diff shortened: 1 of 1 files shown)


docs/content/topics/operators/storage.md
```diff
@@ -9,189 +9,169 @@ aliases:
   - /topics/storage.md
 ---
 
-TODO: update per v2 (no longer any job cache)
+# Storage in Brigade
 
-# How Brigade uses Kubernetes Persistent Storage
+Brigade utilizes storage for its in the following ways:
```
krancour, line 14: ~~for its~~

docs/content/topics/operators/storage.md
```diff
@@ -9,189 +9,169 @@ aliases:
...
 
-Brigade allows script authors to declare two kinds of storage:
+  * [Shared Worker storage](#shared-worker-storage) wherein a Brigade Worker's
+    workspace may be shared with its Job(s). Optional; not enabled by default.
```
krancour, line 17: Nit: The way this is worded, it isn't clear that jobs can use this to share information with one another.

docs/content/topics/operators/storage.md
```diff
@@ -9,189 +9,169 @@ aliases:
...
-- per-job caches, which persist across builds
-- per-build shared storage, which exists as long as the build is running
+## Shared Worker storage
 
-Usage of these is described within the [JavaScript docs](javascript.md) and the
-[scripting guide](scripting.md).
+The workspace for a Brigade Worker can be shared with all Worker Jobs. This is
+an opt-in feature and isn't enabled by default. When enabled, Brigade will
+create a [PersistentVolume] on the underlying Kubernetes substrate and
+automatically add the corresponding volume mount to each Worker Job created.
```
krancour, line 26: Jobs also have to opt-in by specifying a mount path. If they don't, they don't get access to the shared storage.

docs/content/topics/operators/storage.md
```diff
@@ -9,189 +9,169 @@ aliases:
...
+[storage class] configured on your Kubernetes substrate. See the [Access Modes]
+matrix for compatibility. Brigade is well-tested using [NFS] and [Azure File]
+on [AKS]. ([Azure Disk] does *not* support this required access mode.)
 
-## Brigade and PersistentVolumeClaims
+To configure the storage class that should be used for shared Worker storage,
+set the `worker.workspaceStorageClass` field in Brigade's [Helm chart values]
+file to the name of the storage class. For example, if NFS is set up on the
+Kubernetes cluster and its storage class is named `nfs`, this would be the
+appropriate configuration:
```
krancour, line 39: Worth mentioning that nfs is also the default in the chart?

docs/content/topics/operators/storage.md
krancour: Nevermind... you cover this pretty well further on down. I don't think we need to say it here also.

docs/content/topics/operators/storage.md
```diff
@@ -9,189 +9,169 @@ aliases:
...
-$ helm install --name nfs stable/nfs-server-provisioner
+Each Worker Job requiring access to this workspace must then be configured with
+a filepath value designating where the workspace should be mounted within the
+Job's container. (Note that this may be the Job's `primaryContainer` and/or one
+or more of a Job's `sidecarContainer`(s), if applicable). This filepath value
+is assigned to the `workspaceMountPath` field on each applicable Job container.
+
+In the example `brigade.js` script below, the Job named "first-job" is
+configured with the `workspaceMountPath` value set to `/share` on both its
+primary container and its sidecar container, "sidecar". The sidecar container
```
krancour, line 76: Using a sidecar in this example might unnecessarily distract from what we're trying to get across.

Showing two jobs that execute in serial, with the first writing a message to a file and the second cat'ing the file might be more straightforward for readers.

docs/content/topics/operators/storage.md
```diff
@@ -9,189 +9,169 @@ aliases:
...
+## Artemis storage
 
-In either case, there are two storage class settings:
+Brigade uses [ActiveMQ Artemis] as its messaging queue component. For more
+details on its function in Brigade, see the [Design] doc.
 
-- `cacheStorageClass`: This is used for the Job cache.
-- `buildStorageClass`: This is used for the shared per-build storage.
+Messages (i.e. work to be scheduled on the substrate) should be persisted even
+when/if Artemis itself goes down or is restarted. Therefore, by default,
```
krancour, line 113: Super-nit: Is "if/when" more common than "when/if." I the voice in my head stumbled over this.

docs/content/topics/operators/storage.md
```diff
@@ -9,189 +9,169 @@ aliases:
...
 
-values.yaml
-```yaml
-worker:
-  defaultBuildStorageClass: nfs
-  defaultCacheStorageClass: nfs
-```
+As the access mode is `ReadWriteOnce`, nearly all storage class types should
+function correctly for this PersistentVolume. By default, no storage class is
+specified in the chart, which means the default storage class on the substrate
```
krancour, line 125: A thought: Talk of "substrates" might make sense for end-users who are well-abstracted from Kubernetes. For an operator, it might be distracting. Since k8s is a given for operators, throughout this particular bit of doc, we could probably get away with directly saying "Kubernetes."

docs/content/topics/operators/storage.md
```diff
@@ -9,189 +9,169 @@ aliases:
...
-  managed-premium
-❯ nfs
-  Leave undefined
-```
+## Examples
 
-If you would prefer to use the NFS provisioner as a cluster-wide default volume provider
-(and have Brigade automatically use it), you can do so by making it the default
-storage class:
+### Using an NFS Server
```
krancour, line 155: The placement of this section threw me off. Would it make more sense if it were moved up to where shared workspace storage is discussed?

krancour:
@vdice I added a few nits and other thoughts, but overall this is great. Very concise.

docs/content/topics/operators/storage.md
```diff
@@ -9,189 +9,169 @@ aliases:
...
+when/if Artemis itself goes down or is restarted. Therefore, by default,
+persistence is enabled via the appropriate configuration on the
+[Brigade Helm Chart][Helm chart values]. The default access mode for the
+backing PersistentVolume is `ReadWriteOnce`. This mode, although configurable,
+should not need changing.
 
-To set these at the Brigade server level, set the values below accordingly in the Brigade
-chart's `values.yaml` file:
+There are, however, other persistence options that may be useful to customize,
+such as volume size and storage class type. All configuration options can be
```
krancour, line 120: Should we emphasize that the default storage size is very small and not adequate for production deployments?

Same Q goes for mongodb.

docs/content/topics/operators/storage.md
vdice: Good point... what about `may be shared among its Job(s)`?

docs/content/topics/operators/storage.md
vdice: Oh, yes, I forgot it was the default.  I'll add a mention.

docs/content/topics/operators/storage.md
krancour: How about "with _and_ among?"

docs/content/topics/operators/storage.md
vdice: Perhaps true.  I just noticed that the example shared workspace project does what you've mentioned (two jobs in serial) and there didn't seem to be any examples showing configuring a Job's sidecar container as well, so I wanted to demonstrate that knowledge here.  However, no super strong opinion.  If we can get an example up (or maybe update the aforementioned example showing sidecar config), then I'd be cool with changing this up per your recommendation.  WDYT?

docs/content/topics/operators/storage.md
vdice: Good call, I moved it up and reworded things... lmk what you think: https://github.com/brigadecore/brigade/pull/1586/commits/42a85411f712ebf74b8a76d27e25be9f6c45391a

docs/content/topics/operators/storage.md
krancour: Yes. Let's add a separate sidecar example and maybe we mention sidecars in the scripting guide.

krancour: fyi, this is the only unresolved issue I can spot.

vdice: Oh thanks, just noticed this.  Will update next.

vdice: Updated; created separate issue tracking sidecar container docs/example.

krancour (approved):
LGTM

PR #1588, opened 2021-08-27
worker: git sidecar gets pull policy from job config

### LINKED ISSUE #1584: ImagePullPolicy For Init Container should not be hardcoded
https://github.com/brigadecore/brigade/blob/e4c8dc44ec2aa9a1ecd8ecc881b4b4938bbdcde3/brigade-worker/src/k8s.ts#L874

As seen here, the value is hardcoded to "IfNotPresent", this should be using the value supplied by job.imageForcePull

krancour:
Fixes #1584


krancour:
This was closed accidentally.

PR #1595, opened 2021-08-31
add worker support for event summary field, pt2

### LINKED ISSUE #1590: add event summary field
Events have a payload that bears _input_ from the source (upstream system, via gateway) to the worker/script that handles the event.

There's currently no complementary mechanism that allows a handler to "return" output.

This issue proposes adding a "summary" field to events.

Brigadier currently allows events to be handled like so:

```TypeScript
  // Handle the event directly or delegate to jobs...
});
```

As you can see, the handler function receives an event as a param and returns nothing. i.e. Return type is `void`.

I'd like to propose that the return type be changed to `void` _or_ `string`. (I'm pretty sure TS/JS can do that.) Since void would still be supported, this would not be a breaking change.

If a `string` is returned by the handler, the worker can treat it as a summary and automatically use the API to persist this.

Gateways that monitor the status of events they have created for purposes of reporting status upstream may be able to make good use of information found in the summary.

My WIP in Slack gateway is a good example of a gateway that would get a lot of value out of this enhancement. This gateway translates Slack slash commands into events. Events are immediately ack'ed and then monitored for completion. Currently, upon completion, the gateway messages the channel from which the slash command originated to provide a status update. Knowing an event succeeded is often not sufficiently contextually useful for the human user who reads the status update message in slack. Users would likely derive more benefit if the status update messages included a use case specific summary created by the worker/script that handled the event.

cc @itowlson because this idea came via a chat I had with you. Thanks!

krancour:
Follows up on #1593 

Fixes #1590

### DIFF
--- a/examples/02-first-event/project.yaml
+++ b/examples/02-first-event/project.yaml
@@ -17,6 +17,10 @@ spec:
 
         events.on("brigade.sh/cli", "exec", async event => {
           console.log("Hello, World!");
+
+          // Optionally return a string and it will automatically be persisted
+          // in the event's summary field.
+          return "All done."
         });
 
         events.process();
--- a/v2/brigadier-polyfill/package.json
+++ b/v2/brigadier-polyfill/package.json
@@ -26,7 +26,7 @@
     "ts-node": "^9.1.1"
   },
   "dependencies": {
-    "@brigadecore/brigade-sdk": "^v2.0.0-beta.2",
+    "@brigadecore/brigade-sdk": "^v2.0.0-beta.4",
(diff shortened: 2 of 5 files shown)


vdice (approved):
Code LGTM.  Thoughts on adding an example of adding a summary to an event in one of the example project scripts?  Could be a follow-up, of course.

krancour:
@vdice I'll add an example to this PR.

krancour:
Needs bug fix from https://github.com/brigadecore/brigade-sdk-for-js/pull/40 before this can be merged.

Issue #1587, opened 2021-08-26
Events: Understanding and handling Brigade Events


vdice:

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1596

PR #1596, opened 2021-08-31
docs(events.md): add Events doc

vdice:
* Adds Events doc

Preview: https://deploy-preview-1596--brigade-docs.netlify.app/topics/project-developers/events/

Closes https://github.com/brigadecore/brigade/issues/1587

### DIFF
--- a/docs/content/topics/project-developers/events.md
+++ b/docs/content/topics/project-developers/events.md
@@ -66,10 +66,10 @@ on a particular GitHub repo.
 ### Payload
 
 An event Payload can be used to send input that may be utilized when handling
-the event, but otherwise opaque to Brigade itself. One example would be sending
-the original GitHub push event payload on the corresponding GitHub gateway
-event, so that these additional details may be parsed in a project's Brigade
-script.
+the event, but is otherwise opaque to Brigade itself. One example would be
+sending the original GitHub push event payload on the corresponding GitHub
+gateway event, so that these additional details may be parsed in a project's
+Brigade script.
 
 ### Project ID
 
@@ -85,9 +85,10 @@ Projects can choose to utilize Labels for filtering purposes. The labels on a
 project's event subscription do not need to exactly match the labels on an
 Event in order for the project to receive it. Labels can optionally be used to
 narrow an event subscription's scope by selecting only events that are labeled
-in a particular way. For example, a label of `branch: main` might be used on an
-event subscription to limit inbound GitHub events to only those that occur on
-the `main` branch of a given repo.
+in a particular way. For example, the Slack gateway labels all events with the
+channel of origin. A project's subscription can optionally ignore the label and
+get all events for the workspace or specify the label and get events only from
+a specific channel.
 
 ### Qualifiers
 
@@ -101,8 +102,9 @@ their event subscription in order to receive an event.
 
 ### Short Title
 
-A short title may be provided for an event. This can then be utilized for
-logging, categorization or visual representation purposes, etc.
+_Reserved for future use._ A short title may be provided for an event. This can
+then be utilized for logging, categorization or visual representation purposes,
+etc.
 
 ### Long Title
 
--- a/docs/content/topics/project-developers/events.md
+++ b/docs/content/topics/project-developers/events.md
@@ -11,7 +11,7 @@ aliases:
 
 # Brigade Events
 
-Events are the lingua franca in Brigade: [Gateways] emit events and [Project]
+Events are the _lingua franca_ in Brigade: [Gateways] emit events and [Project]
 developers write logic to handle them.
 
 In this document, we'll look at:
@@ -39,8 +39,8 @@ The full list of fields on an Event is:
   * [Type](#type)
   * [Payload](#payload)
   * [ProjectID](#project-id)
-  * [Qualifiers](#qualifiers)
   * [Labels](#labels)
+  * [Qualifiers](#qualifiers)
   * [Short Title](#short-title)
   * [Long Title](#long-title)
   * [Source State](#source-state)
@@ -50,11 +50,11 @@ Each field is reviewed in depth below.
 
 ### Source
 
-An event's Source can be thought of as its domain. In fact, the rule of thumb
-to avoid clashes is to use a URI you control. This means leading with one's own
-domain or the URL for something else one owns, like the URL for a GitHub repo.
-As an example, events emitted from Brigade's [GitHub Gateway] use a source
-value of `brigade.sh/github`.
+An event's Source can be thought of as its domain. For developers writing their
+own gateway, the rule of thumb to avoid clashes is to use a URI one controls.
+This means leading with one's own domain or the URL for something else one
+owns, like the URL for a GitHub repo. As an example, events emitted from
+Brigade's [GitHub Gateway] use a source value of `brigade.sh/github`.
 
 ### Type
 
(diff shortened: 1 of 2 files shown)


docs/content/topics/project-developers/events.md
```diff
@@ -9,4 +9,229 @@ aliases:
   - /topics/project-developers/events.md
 ---
 
-TODO
\ No newline at end of file
+# Brigade Events
+
+Events are the lingua franca in Brigade: [Gateways] emit events and [Project]
```
krancour, line 14: Total nit, but "lingua franca" should probably be italicized, as that's the common convention for foreign words.

This is so much of a nit that I don't care at all if you ignore it.

docs/content/topics/project-developers/events.md
```diff
@@ -9,4 +9,229 @@ aliases:
...
+
+Each field is reviewed in depth below.
+
+### Source
+
+An event's Source can be thought of as its domain. In fact, the rule of thumb
+to avoid clashes is to use a URI you control. This means leading with one's own
+domain or the URL for something else one owns, like the URL for a GitHub repo.
+As an example, events emitted from Brigade's [GitHub Gateway] use a source
+value of `brigade.sh/github`.
```
krancour, line 57: I might be a useful clarification that this rule of thumb is applicable when developing one's own gateway. That context seems to be missing, and without it, this could be confusing.

docs/content/topics/project-developers/events.md
```diff
@@ -9,4 +9,229 @@ aliases:
...
+### Type
+
+The Type of an event is used to specify increased granularity of the event's
+role/purpose as it relates to the source. For example, an event emitted by
+the GitHub Gateway with Type `push` signifies that a push event has occurred
+on a particular GitHub repo.
+
+### Payload
+
+An event Payload can be used to send free-form input that may be utilized when
```
krancour, line 68: I might suggest describing the contents of the field as "opaque" (as far as Brigade itself is concerned), with the value being mutually understood by the originating system and the event-handling logic, but meaningless to anything in between.

krancour: I think there were a few other spots where "opaque" could be applied to describe this field (and others, like summary) more accurately.

docs/content/topics/project-developers/events.md
```diff
@@ -9,4 +9,229 @@ aliases:
...
+handling the event. One example would be sending the original GitHub push event
+payload on the corresponding GitHub gateway event, so that these additional
+details may be parsed in the Brigade script.
+
+### Project ID
+
+Although not normally used by general-purpose gateways, a Project ID value may
+be set on an event. In such cases, the event will _only_ be eligible for
+receipt by the project indicated by this fields' value.
```
krancour, line 77: I _really_ like that way you've described this because it doesn't present it as undermining the subscription-based model, rather as an extension of it.

I do wonder if it's worth mentioning that this is how `brig event create --project <id>` works. That could prove instructive.

docs/content/topics/project-developers/events.md
```diff
@@ -9,4 +9,229 @@ aliases:
...
+
+### Qualifiers
+
+For a project to receive a particular Eeent, the qualifiers on the project's
```
krancour, line 81: s/Eeent/Event/

docs/content/topics/project-developers/events.md
```diff
@@ -9,4 +9,229 @@ aliases:
...
+Eeent subscription must exactly match the qualifiers on the event itself (in
```
krancour, line 82: s/Eeent/Event/

docs/content/topics/project-developers/events.md
```diff
@@ -9,4 +9,229 @@ aliases:
...
+addition to matching Source and Type). For example, a project might supplement
+its `brigade.sh/github` event subscription with a qualifier of
+`repo: brigadecore/brigade`. Only events exactly matching this qualifier would
+be received by the project, i.e., only GitHub events on the brigadecore/brigade
+repository.
+
```
krancour, line 88: I think you accurately stated how qualifiers work, but I'm worried the example might resemble regular label usage.

Do you think it would be useful to introduce labels _first_ as something that _optionally_ narrows a subscription and _then_ describe qualifiers as "like labels" except _required_ in order to match an event?

krancour: Do you suppose that for qualifiers and labels (and I guess source and type for that matter, as well) we should explicitly state that gateways are expected to document what kind of metadata they attach to their events?

docs/content/topics/project-developers/events.md
```diff
@@ -9,4 +9,229 @@ aliases:
...
+### Short Title
+
+A short title may be provided for an event. This can then be utilized for
+logging, categorization or visual representation purposes, etc.
+
+### Long Title
+
+A longer, more descriptive title may be provided for an event. This may be
+helpful for providing additional context for users consuming event details.
```
krancour, line 107: In v1, the only place we used them was Kashti. They got preserved in v2, but we haven't used them anywhere yet.

We don't have to get into the history, but do you think it's worth noting that we don't currently do anything with these fields and they're essentially reserved for future use?

docs/content/topics/project-developers/events.md
```diff
@@ -9,4 +9,229 @@ aliases:
...
+the `qualifiers` and `labels` fields respectively, to further refine inbound
+events.
+
+Here's a look at an event subscription configuration using all of the above:
+
+```yaml
+eventSubscriptions:
+- source: brigade.sh/github
+  qualifiers:
+    repo: brigadecore/brigade
```
krancour, line 171: I vaguely recall that we had a discussion about not using our own repos as examples because it could introduce confusion.

I can't recall if we took any other action after that convo or if it was just something to keep an eye out for going forward.

At any rate, I think seeing the word "brigade" in reference to userland things might improve the clarity of the examples.

docs/content/topics/project-developers/events.md
```diff
@@ -9,4 +9,229 @@ aliases:
...
+```
+
+In fact, all event handlers follow the same pattern, that of:
+
+```javascript
+events.on("event source", "event type", async event => {
+  // handle event
+});
+```
+
```
krancour, line 228: Is it worth mentioning that scripts don't _need_ to be embedded in project definitions and that this is merely one option that happens to be suitable for a simple demo?

I know some people could be instantly turned off at the thought of having to embed JS in YAML.

krancour:
All of this looks like it's on the right track. Made a few suggestions, but nothing major.

docs/content/topics/project-developers/events.md
```diff
@@ -9,4 +9,236 @@ aliases:
...
+The Type of an event is used to specify increased granularity of the event's
+role/purpose as it relates to the source. For example, an event emitted by
+the GitHub Gateway with Type `push` signifies that a push event has occurred
+on a particular GitHub repo.
+
+### Payload
+
+An event Payload can be used to send input that may be utilized when handling
+the event, but otherwise opaque to Brigade itself. One example would be sending
```
krancour, line 69: s/otherwise/is otherwise/

docs/content/topics/project-developers/events.md
```diff
@@ -9,4 +9,236 @@ aliases:
...
+
+### Labels
+
+Projects can choose to utilize Labels for filtering purposes. The labels on a
+project's event subscription do not need to exactly match the labels on an
+Event in order for the project to receive it. Labels can optionally be used to
+narrow an event subscription's scope by selecting only events that are labeled
+in a particular way. For example, a label of `branch: main` might be used on an
+event subscription to limit inbound GitHub events to only those that occur on
+the `main` branch of a given repo.
```
krancour, line 90: This particular example might be dicey because the github gateway doesn't actually use such a label and that could confuse people.

I have looked into adding that label, but there are a lot of events the gateway deals with that are only specific to the repo and _not_ specific to any particular commit, branch, tag, etc...

I haven't ruled out adding this tag in cases where it is applicable, but I worry that if some events have this tag and others don't, it may really complicate things on the subscription end.

krancour: So an example that's more hypothetical might cause less confusion.

Or... a real example that isn't "public" yet would involve my prototype Slack gateway which labels all events with channel of origin. So you can optionally ignore the label and get all events for the workspace or specify the label and get events only from a specific channel.

docs/content/topics/project-developers/events.md
```diff
@@ -9,4 +9,236 @@ aliases:
...
+
+### Qualifiers
+
+Qualifiers are like labels; albeit _required_ rather than optional. When an
+event contains qualifiers, a project's event subscription must declare the same
+qualifiers in order for the project to receive it. For example, no one wants to
+subscribe to events from _all_ GitHub repositories (or even just all the
+repositories in one's org), so the GitHub gateway _qualifies_ events,
+specifying the repository of origin. Projects must match this qualifier in
+their event subscription in order to receive an event.
```
krancour, line 100: Love the way this came out.

docs/content/topics/project-developers/events.md
```diff
@@ -9,4 +9,236 @@ aliases:
...
+
+### Short Title
+
+A short title may be provided for an event. This can then be utilized for
+logging, categorization or visual representation purposes, etc.
+
+### Long Title
+
+_Reserved for future use._ A longer, more descriptive title may be provided
```
krancour, line 109: Same for short title.

docs/content/topics/project-developers/events.md
krancour: > Do you suppose that for qualifiers and labels (and I guess source and type for that matter, as well) we should explicitly state that gateways are expected to document what kind of metadata they attach to their events?

I think this might still be outstanding.

docs/content/topics/project-developers/events.md
vdice: Ok, usin' the Slack gateway example -- best wrap that gateway up! 😜

krancour: I'm on it!

krancour (approved):
LGMT!

Issue #1598, opened 2021-09-01
Using Secrets: Using secrets in Brigade Projects


vdice:

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1599

PR #1599, opened 2021-09-01
docs(secrets.md): updates per v2

vdice:
* Updates the secrets doc per v2

Preview: https://deploy-preview-1599--brigade-docs.netlify.app/topics/project-developers/secrets/

Closes https://github.com/brigadecore/brigade/issues/1598

### DIFF
--- a/docs/content/topics/project-developers/secrets.md
+++ b/docs/content/topics/project-developers/secrets.md
@@ -9,239 +9,157 @@ aliases:
   - /topics/project-developers/secrets.md
 ---
 
-TODO: update per v2
-
 # Secret Management
 
-Brigade provides tools for storing sensitive data outside of your `brigade.js` scripts,
-and then passing that information into the jobs that need them. Brigade accomplishes
-this by making use of Kubernetes secrets.
+Brigade provides tools for storing sensitive data outside of your Brigade
+scripts, and then passing that information into the jobs that need them.
+
+This is accomplished by associating secrets with a given project, which the
+project's script can then access as needed. Secrets are persisted on the
+substrate in the project namespace.
 
 ## Adding a Secret to Your Project
 
-Imagine a case where we need to pass a sensitive piece of information to one of the jobs
-in our `brigade.js`. For example, we might need to pass an authentication token to a job that
-must authenticate to a remote service.
+Imagine a case where we need to pass a sensitive piece of information to one of
+the jobs in a project's Brigade script. For example, we might need to pass an
+authentication token to a job that must authenticate to a remote service.
 
-We do this by storing the secret inside of the project definition.
+We first need to set the secret key and value on the project via `brig`. There
+are two methods to do this, both provided on the `brig project secret set`
+command:
 
-Project definitions are typically managed by `brig`. As you may recall from the [installation
-manual](../intro/install.md), a new project is created like via `brig project create`.
+ * Set a secret in-line:
 
-During the creation process, we are able to add secrets when prompted. If we'd forgotten to
-add a secret and/or additional secrets need to be added to this existing project, we can do so via
-`brig project create --replace -p <existing project name>`.
+  ```console
+  $ brig project secret set --project my-project --set foo=bar
+  ```
 
-Here we add a new secret with key `dbPassword` and value `supersecret` to our existing
-`brigadecore/empty-testbed` project:
+  * Setting a secret via a `secrets.yaml` file:
+
+  ```console
+  $ echo "foo: bar" >> ./secrets.yaml
+  $ brig project secret set --project my-project --file ./secrets.yaml
+  ```
+
+## Accessing a Secret within a Brigade script
+
+Within a Brigade script (`brigade.js` or `brigade.ts` file), we can access any
+of the secrets defined on our project.
+
+This first example demonstrates direct access of a secret defined on a project
+in the event handler code outside of any Job:
+
+```javascript
+// THIS IS NOT SAFE! IT LEAKS YOUR SECRET INTO THE LOGS.
+const { events } = require("@brigadecore/brigadier");
+
+events.on("brigade.sh/cli", "exec", async event => {
+  console.log("Project secret foo = " + event.project.secrets.foo);
+});
+
+events.process();
+```
+
+For passing project secrets into a Job, the best practice is to do so via the
+Job's environment. Brigade inherently treats all Job environment variables as
+sensitive by default, so any instance where their values would otherwise be
+logged will be redacted.
+
+Project secrets that are accessed directly in a Job, i.e. outside of its
+environment, *will* be leaked, both in the Job Pod resource on the substrate
+and via `brig` commands that consume this information, e.g.
+`brig event get --id <event id>`.
+
+Here is an example of passing a project secret into a Job:
+
+```javascript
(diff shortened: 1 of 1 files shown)


docs/content/topics/project-developers/secrets.md
```diff
@@ -9,239 +9,157 @@ aliases:
...
 # Secret Management
 
-Brigade provides tools for storing sensitive data outside of your `brigade.js` scripts,
-and then passing that information into the jobs that need them. Brigade accomplishes
-this by making use of Kubernetes secrets.
+Brigade provides tools for storing sensitive data outside of your Brigade
+scripts, and then passing that information into the jobs that need them.
+
+This is accomplished by associating secrets with a given project, which the
+project's script can then access as needed. Secrets are persisted on the
```
krancour, line 18: Can we emphasize that secrets live _only_ on the substrate and are never stored in the database?

docs/content/topics/project-developers/secrets.md
```diff
@@ -9,239 +9,157 @@ aliases:
...
-  console.log("My DB password is " + project.secrets.dbPassword)
-})
-```
-
-Secrets can be selectively passed to jobs by setting environment variables.
+This first example demonstrates direct access of a secret defined on a project
+in the event handler code outside of any Job:
 
 ```javascript
 // THIS IS NOT SAFE! IT LEAKS YOUR SECRET INTO THE LOGS.
```
krancour, line 53: Can we move this comment closer to the console.log command? I think that would be more clear about what, specifically, is unsafe.

docs/content/topics/project-developers/secrets.md
```diff
@@ -9,239 +9,157 @@ aliases:
...
-trusted key store such as Vault.  See the [example](#fetching-secrets-from-a-remote-store) below.
-
-Alternatively, you could use `secretKeyRef` to reference existing secrets already in your
-Kubernetes cluster.  See the [example](#using-kubernetes-secrets) below.
-
-**I don't want to use Helm to manage my project/secrets. Can I do it manually?**
+For passing project secrets into a Job, the best practice is to do so via the
+Job's environment. Brigade inherently treats all Job environment variables as
+sensitive by default, so any instance where their values would otherwise be
+logged will be redacted.
```
krancour, line 66: This statement was confusing. We don't _actively_ redact secrets whatever they might appear in the logs. We do store every job's environment in a k8s secret, but that is the extent of the measures we take. It's the user's responsibility to be careful about how they use their secrets.

docs/content/topics/project-developers/secrets.md
```diff
@@ -9,239 +9,157 @@ aliases:
...
-})
-```
+const { Job, events } = require("@brigadecore/brigadier");
 
-To run this example, we replace `BRIGADE_PROJECT_NAME` with our Brigade project name and run:
+events.on("brigade.sh/cli", "exec", async event => {
+  let job = new Job("second-job", "debian:latest", event);
+  job.primaryContainer.environment = {"FOO_SECRET": event.project.secrets.foo};
+  job.primaryContainer.command = ["bash"];
+  job.primaryContainer.arguments = ["-c", "echo $FOO_SECRET | wc -c"];
```
krancour, line 82: This example has an interesting mix of good and bad practices. It's good that the secret isn't included directly in the command where it can be seen by anyone who can view the pod. The command gets the secret from an env var. This is a _great_ pattern, I think. But... as far as that the command _does_... it's putting the secret straight into the logs, which is definitely a problem.  Is there something interesting we could do with the secret without printing it?

docs/content/topics/project-developers/secrets.md
```diff
@@ -9,239 +9,157 @@ aliases:
...
+**Why don't all jobs automatically get access to all of the project's secrets?
+Why do I have to pass them to the `Job.environment`?**
 
-As a result, `mySecretReference` points directly to this secret value in this job's environment.
+Brigade is designed to use off-the-shelf Docker images. In the examples above,
+we used the `debian:latest` image straight from DockerHub. We wouldn't want to
+just automatically pass all of our information straight into that container.
+For starters, doing so might inadvertently override an existing environment
+variable of the same name. More importantly, the data might get misused or
+unintentionally exposed by the container.
```
krancour, line 145: This is a great explanation. 👍

docs/content/topics/project-developers/secrets.md
```diff
@@ -9,239 +9,157 @@ aliases:
...
-brig run BRIGADE_PROJECT_NAME -f brigade.js
-```
+**Can I encrypt my secrets?**
 
-We will then see `"hello world"` displayed in the logs.
+Brigade uses [Kubernetes Secrets] for holding sensitive data. Kubernetes Secret
+data is base64-encoded but not encrypted by default. Operators looking to
+encrypt these resources should reference the [Kubernetes documentation]. Note
+that support for this feature will depend on the distribution used (e.g. the
+cloud provider offering or other self-hosted option.)
```
krancour, line 155: imho, we need to state that bottom line is you can trust Brigade secrets as much as you trust your Kubernetes secrets. If you trust how k8s stores your secrets, you're all set. If you don't, then you need to solve that problem before you can trust Brigade's secret handling.

docs/content/topics/project-developers/secrets.md
```diff
@@ -9,239 +9,157 @@ aliases:
...
+
+Another option for Brigade developers to access secret values is via a trusted
+key store external to Kubernetes, such as HashiCorp's [Vault] or [Azure Key
+Vault]. The scripting logic can then use the appropriate CLI or API calls to
+fetch secrets in the event handler(s) that require them.
```
krancour, line 160: This is an interesting option, but... any secrets you use will _still_ end up on the substrate while the jobs run. And beyond that, you probably need to use Brigade secrets to store key vault creds...

So at the end of the day, even this is no more secure than whatever your k8s distro offers.

vdice: True.  Removed this section for now.

docs/content/topics/project-developers/secrets.md
vdice: Added, lemme know what you think.

docs/content/topics/project-developers/secrets.md
vdice: Good point.  Updated with a docker login example.  WDYT?

krancour:
This doc is looking good, but I feel a general concern that we're not being clear enough about why not to do certain things and _how_ those misteps translate into secrets being exposed.

I'm wondering if the examples of what NOT to do should perhaps be followed by an example of how the mistep led to secret exposure.

In the case of secrets ending up in the logs, perhaps a the corresponding command and sample output may be instructive. In the case of container commands, args, etc., perhaps showing a snippet of the corresponding pod manifest might be instructive.

I also think we shouldn't be shy about using bold text to visually emphasize caution.

vdice:
@krancour I pushed some updated text/examples in response to https://github.com/brigadecore/brigade/pull/1599#issuecomment-914682894.  See https://github.com/brigadecore/brigade/pull/1599/commits/aa5d083e870537f2a67525933eb8832586dd3c29.  WDYT?

krancour:
Great update. Nailed it. LGTM

krancour (approved):

PR #1601, opened 2021-09-02
use go 1.17 features to improve tests

krancour:


v2/apiserver/internal/lib/restmachinery/endpoints.go
```diff
@@ -158,6 +158,9 @@ func WriteAPIResponse(
 	statusCode int,
 	response interface{},
 ) {
+	if statusCode == 0 {
```
vdice, line 161: Curious on what provoked these changes?

krancour: This seems to have broken in Go 1.17. It seems that previously, no status code was treated as 200. Does not seem to be the case anymore.

vdice (approved):

PR #1602, opened 2021-09-02
build our own artemis image

krancour:
Up until now, we've used an artemis image from Dockerhub that was produced using this source:

https://github.com/vromero/activemq-artemis-docker

It has worked well, but the project has recently been archived.

With Artemis playing such a critical role in Brigade, I want to be super-certain that we're not at the mercy of an individual in terms of being dependent on an image or source that they solely control and could delete at any time.

This PR introduced our own homegrown Artemis image that takes inspiration both from the repo mentioned above as well as the Artemis project's own instructions for producing a Docker image:

https://github.com/apache/activemq-artemis/tree/main/artemis-docker


krancour:
Converting this to draft while I sort out some issues.

vdice (commented):
Everything looks good.  I think we need to add `load-artemis` to the `hack-load-images` target for the KinD-based integration tests.

krancour:
> I think we need to add load-artemis to the hack-load-images target for the KinD-based integration tests.

D'oh! _That's why the int tests keep failing!_

vdice (approved):

Issue #1608, opened 2021-09-03
Securing Brigade: Securing Brigade via TLS, Ingress and Third-Party Auth


vdice:

krancour:
Wondering if TLS and 3rd party auth should be completely separate sections of the documentation.

vdice:
> Wondering if TLS and 3rd party auth should be completely separate sections of the documentation.

Yea, this issue title stems from the corresponding entry in the meta docs issue, which was written a while back.  Currently, TLS setup is described in the [Deployment doc](https://v2--brigade-docs.netlify.app/topics/operators/deploy/) and 3rd-party auth in the [Authentication doc](https://v2--brigade-docs.netlify.app/topics/administrators/authentication/).

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1609

PR #1609, opened 2021-09-03
docs(security.md): v2 updates

vdice:
* Updates the Security doc per v2

I admittedly didn't come up with a bunch of new material for this one.  Maybe someone more enthusiastic about adding new or updated text can chime in?  (Note: totally fine if commits from maintainers are pushed to this branch.)  Otherwise, we'll at least have updated the previous points per v2.

Preview: https://deploy-preview-1609--brigade-docs.netlify.app/topics/operators/security/

Closes https://github.com/brigadecore/brigade/issues/1608

### DIFF
--- a/docs/content/topics/operators/security.md
+++ b/docs/content/topics/operators/security.md
@@ -19,28 +19,35 @@ For that reason, it is prudent to configure security.
   in a Helm install, do `helm install --namespace brigade ...`.
 - *RBAC Enabled*: When installing with Helm, role-based access control is enabled by default.
   To disable,`--set rbac.enabled=false` will turn off role-based access control.
-- *Do not run more than one brigade per namespace*: Running multiple installs of Brigade
+- *Do not run more than one Brigade per namespace*: Running multiple installs of Brigade
   in the same namespace can cause naming collisions which could result in
   unauthorized access to pods.
-- *Do not run multiple tenants against the same brigade*: Brigade does not implement
-  security controls that allow multiple tenants to share the same Brigade instance.
-  Brigade supports multiple projects per Brigade server instance, but those projects
-  should be owned by the same tenant.
 
 ## API Server Security
 
-TODO: update
+If Brigade's API server will be exposed to the internet, e.g. via an external
+IP (recall that Brigade's default service type in the [Brigade chart] is
+`LoadBalancer` and so a public-facing IP will be provisioned), care should be
+taken to secure inbound communication. Minimally, TLS should be enabled and 
+SSL certificates should not be self-signed. You may also consider using an
+[Ingress Controller] for an extra layer of traffic filtering and/or protection.
+
+For more details on deploying Brigade securely, see the [Deployment] doc.
+
+[Brigade chart]: https://github.com/brigadecore/brigade/tree/v2/charts/brigade
+[Ingress Controller]: https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/
+[Deployment]: /topics/operators/deploy
 
 ## How RBAC Is Configured
 
-The Helm [chart](https://github.com/brigadecore/brigade/tree/v2/charts/brigade) for Brigade
-includes an RBAC configuration that is designed to run in an isolated namespace.
+The Helm [chart][Brigade chart] for Brigade includes an RBAC configuration that
+is designed to run in an isolated namespace.
 
-The RBAC defines a Service Account for every Deployment (API, Observer, Scheduler, etc.),
-as well as a Service Account for the Brigade worker.
+The RBAC defines a Service Account for every Deployment (API, Observer,
+Scheduler, etc.), as well as a Service Account for the Brigade worker.
 
-Each deployment also has a Role, which describes the permissions that the particular
-service needs.
+Each deployment also has a Role, which describes the permissions that the
+particular service needs.
 
 The resulting Role Bindings, then, are a simple one-to-one match between the
 Service Account and the Role.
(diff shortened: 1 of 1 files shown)


docs/content/topics/operators/security.md
```diff
@@ -19,28 +19,35 @@ For that reason, it is prudent to configure security.
   in a Helm install, do `helm install --namespace brigade ...`.
 - *RBAC Enabled*: When installing with Helm, role-based access control is enabled by default.
   To disable,`--set rbac.enabled=false` will turn off role-based access control.
```
krancour, line 21: This seems a bit off. Even for v1.x, it seems a bit off.

Setting `rbac.enabled=false` won't really _disable_ RBAC. It will prevent roles and rolebindings from being generated and installed, and that. On an RBAC-enabled cluster, setting `rbac.enabled=false` would result in Brigade simply not working.

Probably the best thing to say here is just that if RBAC is enabled in the cluster, then it needs to be enabled here as well. If it's disabled in the cluster, it has to be disabled here as well, although in that case, I think we should _also_ say that we won't vouch for the security of anything Brigade-related in a non-RBAC-enabled cluster.

docs/content/topics/operators/security.md
```diff
@@ -19,28 +19,35 @@ For that reason, it is prudent to configure security.
...
-- *Do not run more than one brigade per namespace*: Running multiple installs of Brigade
+- *Do not run more than one Brigade per namespace*: Running multiple installs of Brigade
   in the same namespace can cause naming collisions which could result in
   unauthorized access to pods.
```
krancour, line 24: I don't think this is an issue anymore on account of every project getting its own namespace, although I cannot say I have tried it.

docs/content/topics/operators/security.md
```diff
@@ -19,28 +19,35 @@ For that reason, it is prudent to configure security.
...
-- *Do not run multiple tenants against the same brigade*: Brigade does not implement
-  security controls that allow multiple tenants to share the same Brigade instance.
-  Brigade supports multiple projects per Brigade server instance, but those projects
-  should be owned by the same tenant.
```
krancour, line 28: I might be slightly hesitant to remove this outright, although we could perhaps reword it.

While we have much better access controls now, there could still be reasons to run a Brigade per team, for instance.

I say this because the `READER` role confers _global_ read access (with the exception of logs). This was a deliberate choice to avoid the very difficult problem of filtering project list results on the basis of which projects the user has access to... so a comparison could be made to GitHub. Every project could be likened to a _public_ repo. Everyone can see. Not everyone can write...

If for whatever reason that's unacceptable, then the practical solution is to run multiple Brigades.

docs/content/topics/operators/security.md
```diff
@@ -19,28 +19,35 @@ For that reason, it is prudent to configure security.
...
 
 ## API Server Security
 
-TODO: update
+If Brigade's API server will be exposed to the internet, e.g. via an external
+IP (recall that Brigade's default service type in the [Brigade chart] is
+`LoadBalancer` and so a public-facing IP will be provisioned), care should be
+taken to secure inbound communication. Minimally, TLS should be enabled and 
+SSL certificates should not be self-signed. You may also consider using an
+[Ingress Controller] for an extra layer of traffic filtering and/or protection.
```
krancour, line 33: I'm not sure that an ingress controller confers any additional degree of protection.

Perhaps consider rewording what precedes this. Something like:

> If Brigade's API server will be exposed to the internet either via a service of type `LoadBalancer` having a public IP or via an ingress controller, care should...

docs/content/topics/operators/security.md
```diff
@@ -19,28 +19,35 @@ For that reason, it is prudent to configure security.
...
+[Brigade chart]: https://github.com/brigadecore/brigade/tree/v2/charts/brigade
+[Ingress Controller]: https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/
+[Deployment]: /topics/operators/deploy
 
 ## How RBAC Is Configured
 
-The Helm [chart](https://github.com/brigadecore/brigade/tree/v2/charts/brigade) for Brigade
-includes an RBAC configuration that is designed to run in an isolated namespace.
+The Helm [chart][Brigade chart] for Brigade includes an RBAC configuration that
+is designed to run in an isolated namespace.
```
krancour, line 44: I'm not entirely clear on what this means.

docs/content/topics/operators/security.md
```diff
@@ -19,28 +19,35 @@ For that reason, it is prudent to configure security.
...
 
-The RBAC defines a Service Account for every Deployment (API, Observer, Scheduler, etc.),
-as well as a Service Account for the Brigade worker.
+The RBAC defines a Service Account for every Deployment (API, Observer,
+Scheduler, etc.), as well as a Service Account for the Brigade worker.
```
krancour, line 47: Each project gets its own namespace. Within each of those, there's a service account for the workers and another for the jobs.

This allows competent k8s users with adequate permissions to modify the service account(s) for a given project to permit things that you can't do otherwise-- such as scripting k8s itself.

docs/content/topics/operators/security.md
```diff
@@ -51,30 +58,31 @@ secrets, and persistent volume claims.
...
+[Secrets] on that project. In turn, project secrets are stored in a Kubernetes
+secret, so care should be taken in preventing unauthorized access to these
+resources.
 
-- Out-of-the-box, the project is the location where credentials should be stored
+- Out-of-the-box, credentials should be stored as project secrets
 - A project's credentials are accessible to any script running in that project,
   regardless of event.
-- For SSH-based Git clones, the SSH key should be stored on the project.
+- For SSH-based Git clones, the SSH key should be stored as a project secret.
```
krancour, line 70: iirc, there's a well-known secret name used for this purpose. Should we document that here?

docs/content/topics/operators/security.md
```diff
@@ -51,30 +58,31 @@ secrets, and persistent volume claims.
...
-to secure Helm.
+[Secrets]: /topics/project-developers/secrets
 
 ## Script Security
 
-Brigade scripts can create pods, secrets, and persistent volume claims. Brigade does not
-evaluate the security of the containers that a pod runs. Consequently, it is best
-to avoid using untrusted containers in Brigade scripts. Likewise, it is not recommended
-to inject secrets into a container without first auditing the container.
+Brigade scripts can create pods, secrets, and persistent volume claims. Brigade
```
krancour, line 76: I might suggest adding the word "indirectly" here so it's clear that scripts don't directly create job pods and other resources. The API server does all that.

docs/content/topics/operators/security.md
```diff
@@ -87,3 +95,8 @@ suggest the following features of a gateway:
   - The exception is alternative VCS implementations, where the git-sidecar may
     be replaced by another sidecar.
 
+See more info in the [Gateways] doc. Their you'll find links to official
```
krancour, line 98: s/Their/There/

vdice:
Oops, thanks for the initial review @krancour -- somehow missed the feedback.  Looking at addressing now.  Will ping when ready for re-review.

docs/content/topics/operators/security.md
vdice: Ah, right, just checked assumptions here.  As it stands currently, it is only possible for one Brigade instance to deploy (and function correctly) per cluster.  I'll add a note around this.  We can always revise if/when we support multiple instances on a cluster.

docs/content/topics/operators/security.md
vdice: Good point; didn't look super closely when this was copied over... doesn't really make sense.  Will trim.

docs/content/topics/operators/security.md
vdice: Actually, I removed this note for now, since we now have https://github.com/brigadecore/brigade/issues/1630

krancour (approved):
LGTM

PR #1612, opened 2021-09-08
Docs/brigadier

vdice:
* This is the first doc in the Scripting topic to be updated, so I've added text to the Scripting index and shuffled the order of this section a bit

* Updates the Brigadier doc:
  * Updates first sections per v2 changes
  * Removes the duplicated API docs...
 
* Which will now be served after having been generated straight from the well-documented/commented ts/js source code
  * Currently, the infra to do so is to build/generate both the TS and the JS docs and serve them via GH Pages
  * Demo representations can be seen on my fork via: https://vdice.github.io/brigade/js and https://vdice.github.io/brigade/ts

TODO
  - [ ] add ghToken secret to this project (assuming we go the gh pages route)


vdice:
@krancour wanted to put this WIP up now to get your gut take on current approach for doc generation/hosting.  What do we think so far?

krancour:
> Just want to make sure we're good with moving forward and thus hosting js and ts docs via this gh-pages approach.

That seems reasonable to me.

vdice:
Also renamed the docs publish yarn script name to `publish-docs` so as not to confuse w/ `yarn publish`...

PR #1616, opened 2021-09-16
upgrade to helm 3.7.0 ga + pre-release version bumps

### LINKED ISSUE #1594: Revisit Helm chart install instructions, etc. with Helm 3.7.0
Helm [3.7.0](https://github.com/helm/helm/milestone/111) is anticipated to ship with [changes to the experimental OCI support](https://github.com/helm/helm/pull/9782) that we rely on for installing Brigade (and all (?) Brigadecore Gateways).  Let's see what needs updating in docs, etc., once 3.7.0 is out.

krancour:
Fixes #1594 

This also includes pre-release version bumps because beta.2 was published using Helm 3.6 and is _not_ compatible with these Helm 3.7+ instructions.

i.e. We need to release beta.3 after we merge this.

While already touching quickstart and install instructions, I also switched from using "brigade2" as the release and namespace name and started using just plain "brigade."

The instructions were originally written with users who are already running brigade 1.x in mind, but as v2 marches closer to GA, we can start treating v2 as our assumption when "brigade" is discussed.


vdice (approved):
LGTM, pending the beta.3 release.

Issue #1605, opened 2021-09-02
Scripting Guide: A guide to writing `brigade.js`/`brigade.ts` scripts in Brigade


vdice:

vdice:
@krancour any thoughts/ideas on overall layout of this scripting guide?  I was initially thinking maybe just a gradual progression from trivial to more complex examples, just like the examples/ dir is structured.  In fact, we could do exactly that: present each script from the examples dir with a link back to each example... one possibility anyways.  Or, we could put our minds together and get more creative...

PR #1629, opened 2021-09-23
docs(scripting guide): update per v2

vdice:
* Updates the Scripting Guide per v2
* Re-orgs example projects, as the updated guide moves through some of them in ascending order of complexity
* Comment's out the `logLevel: DEBUG` config on the example projects so that the output is cleaner when presented in the guide

Closes https://github.com/brigadecore/brigade/issues/1605


docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,864 @@ aliases:
...
 
-- A Brigade server running in your cluster
-- A [project](./projects.md) in which the script will run
+- A Brigade server [running in your Kubernetes cluster][quickstart]
+- A [project] in which the script will run
 - (Optionally) A source code repository (e.g. git) that contains data our
-- (Optionally) A [`brigade.json`](workers.md) file that contains additional dependencies that can be used from `brigade.js`
-- (Optionally) Other configuration data, like [secrets](secrets.md)
+  script may utilize
+- (Optionally) A [`brigade.json`] file that contains additional dependencies
```
vdice, line 34: Note, this may need to be updated per https://github.com/brigadecore/brigade/issues/1636

vdice: Updated.

krancour (approved):
LGTM. Totally appreciate keeping the examples in the same order as the docs.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
+(`brigade.ts`).
 
 # Brigade Scripts, Projects, and Repositories
 
-Brigade scripts are stored in a `brigade.js` file. They are designed to contain short scripts
-that coordinate running multiple related tasks. We like to think of these as
-cluster-oriented shell scripts: The script just ties together a bunch of other programs.
+Brigade scripts can be stored either in a project definition or in a project's
+git repository. They are designed to organize and coordinate running multiple
+related tasks, usually in response to events. We like to think of these as
```
krancour, line 22: Nit: What do you think about dropping "usually?" There really is no case where the script _isn't_ triggered by an event.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
-- A [project](./projects.md) in which the script will run
+- A Brigade server [running in your Kubernetes cluster][quickstart]
+- A [project] in which the script will run
 - (Optionally) A source code repository (e.g. git) that contains data our
-- (Optionally) A [`brigade.json`](workers.md) file that contains additional dependencies that can be used from `brigade.js`
-- (Optionally) Other configuration data, like [secrets](secrets.md)
+  script may utilize
+- (Optionally) A [`package.json`] file that contains additional dependencies
+  that can be used in the script
+- (Optionally) Other configuration data, like [secrets]
```
krancour, line 36: I'd consider simplifying this section. Some aspects seem inaccurate in nuanced ways. (For instance, "A project in which the script will run" almost implies a project is a runtime environment like a container.) Rather than spend energy rewording, I'm wondering if understanding of what a Brigade script is can be conveyed without this level of detail.

Would it be enough to state that the Brigade script is the JavaScript or TypeScript is defined by a Brigade project and that Brigade executes that script in the context of a worker to handle events to which the project is subscribed?

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
+Created event "261229dc-1140-4f6a-bf91-bd2a69f31721".
+
+Waiting for event's worker to be RUNNING...
+2021-09-20T22:15:12.047Z INFO: brigade-worker version: a398ba8-dirty
+Hello, World!
+```
+
+This example executes the log in the _global scope_. While it's fine to work
+in the global scope, most of the time what we want to do with Brigade is write
+_event handlers_.
```
krancour, line 95: "Global scope" could be confusing. What would you think about saying that this script _unconditionally_ logs "Hello, World!" in response to _any_ event to which the project is subscribed and that it's more common to in corporate event handlers to handle specific events in different ways?

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
 
 ## Brigade Events and Event Handlers
 
-Brigade listens for certain things to happen, such as a new push on a GitHub repository or an
-image update on DockerHub. (The events that it listens for are configured in your
-project).
+Brigade listens for certain things to happen, such as a new push on a GitHub
+repository or an image update on DockerHub. (The events that it listens for are
+configured in your project under the `eventSubscriptions` section).
```
krancour, line 101: "Brigade listens" could be misleading. I know we probably don't want to rehash all the details of the event/gateway model right here, but maybe we could reword this a little bit and link to other relevant docs?

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
-is an _event handler_ that matches the _event_.
+When Brigade observes such an event, it will load the project script and
+determine if an _event handler_ matching the _event_ exists.
 
 Using the `brig` tool introduced above, we can see how this works.
 
-The `brig` tool triggers an `exec` event (short for _execute_) on Brigade. So our
-`brigade.js` file can intercept this event and respond to it:
+The `brig` tool triggers an `exec` event (short for _execute_) on Brigade. So
+our script can intercept this event and respond to it:
```
krancour, line 109: I understand why this is in this section and not the previous one-- because in the previous example, we unconditionally responding to all subscribed events the same, so the notion of an "event" first becomes relevant in this example.

That said, I feel like it broke the flow of the doc for me seeing `brig event create` explained _after_ having already used it.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
-- It declares a single event handler that says "on an 'exec' event, run this function".
+- It imports the `events` object from `@brigadecore/brigadier`. Almost all
+  Brigade scripts do this.
+- It declares a single event handler that says "on an 'exec' event, run this
+  function".
+- We're not actually utilizing the `event` object in this script, but we will
+  in a later example
+- The event handler returns a string, which is the [event summary]. This is can
+  be utilized by gateways interested in the results or other context related
+  to the handling of the event.
```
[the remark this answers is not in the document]
krancour: Also, the comment in the code says it's optional, but can we emphasize that here as well?

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
+- All scripts must call `events.process()` in order to start up the node
+  process that the script will run in.
```
krancour, line 138: Could I suggest:

> Scripts such as this one must invoke `events.process()` _after_ registering all handlers in order to dispatch the event.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
+Hello, World!
 ```
-[brigade-04.js](../../examples/brigade-04.js)
 
-The `brig` client will trigger the `exec` event. But then when that event has been
-processed, Brigade will trigger an `after` event before returning:
+Since Brigade did not see a GitHub push event (in this case, an event from
+source `brigade.sh/github` and of type `push`), it did not execute the
+corresponding event handler. It only executed the handler for the event created
+by `brig`, i.e. an event from source `brigade.sh/cli` and of type `exec`.
```
krancour, line 185: "Did not see" is another case where I worry we could be misrepresenting the event/gateway model. Could we reword this section to something like the following:

> Since the event we created was from `brigade.sh/cli` and of type `exec`, only the corresponding handler was executed. The handler for events from `brigade.sh/github` and of type `push` was not.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
+2021-09-20T22:24:57.655Z INFO: brigade-worker version: a398ba8-dirty
+Hello, World!
 ```
 
-One of the strengths of Brigade is that we can handle different events in the same file.
-Brigade will only trigger the appropriate events. For example, we can expand the example
-above to also provide a handler for the GitHub `push` event:
+One of the strengths of Brigade is that we can handle different events in the
+same file. Brigade will only trigger the appropriate events. For example, we
+can expand the example above to also provide a handler for a GitHub push event.
```
krancour, line 155: Rather than assert this is a strength, can we maybe connect the dots more explicitly here. Hopefully the readers will conclude on their own that it is a strength.

Could we say that since projects may subscribe to multiple types of events (from multiple sources, too), defining multiple event handlers in your script permits different types of events to be _handled_ differently?

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
-- When Brigade triggers an event, it creates a new `build`, which you can think of as "a
-  specific run of your `brigade.js` file.
-- A build has several _jobs_, where each job starts a container
-- A job runs zero or more _tasks_ inside of a container
+- An _event handler_ declares how an event (`brigade.sh/github:push`,
+  `brigade.sh/cli:exec`) is processed
+- When Brigade triggers an event that a project subscribes to, Brigade creates
+  a _Worker_ per the project's configuration
+- A Worker may run several _Jobs_, where each job starts a container
+- A Job runs zero or more _commands_ inside of a container
```
krancour, line 201: Two things:

Can we simplify much of this by saying that Brigadier exposes an API for defining and executing jobs?

Can we explain the benefit of jobs somehow?

A digression we don't necessary need to include-- in v1, we seemed to discourage people from performing actual work in their workers and we aren't really discouraging that now, which I think is good, but... `</digression>`

The worker's image is derived from a Node.js image and containers based on that image can execute JavaScript and TypeScript only. In the frequent case that other runtimes, tools, etc. _not_ present on the worker image are required in order to respond to an event, spawning a job permits aspects of event-handling to be delegated to a more suitable container.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
 ```
-[brigade-05.js](../../examples/brigade-05.js)
+[03-first-job](https://github.com/brigadecore/brigade/tree/v2/examples/03-first-job)
 
-The first thing to note is that _we have changed the first line_. In addition to importing
-the `events` object, we are now also importing `Job`. `Job` is the prototype for creating
-all jobs in Brigade.
+The first thing to note is that _we have changed the first line_. In addition
+to importing the `events` object, we are now also importing `Job`. `Job` is the
+prototype for creating all jobs in Brigade.
```
krancour, line 224: Nit: As someone who confessed to knowing just enough JS to be dangerous, "prototype" is a JS term that immediately triggers something negative in me. If there's another way to word this, it might be more welcoming to people who aren't so keen on JS.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
+To run a job, we use the Job's `run()` method. Behind the scenes, Brigade will
+start a new `debian` pod (downloading it if necessary), execute it, and
+monitor it. When the container is done executing, the run is complete.
 
-It is worth noting that a `run()` method is _asynchronous_. That means that when we call
-`run()`, it will start processing in the background. Later on, we will see how we can take
-advantage of that to create pipelines.
+It is worth noting that a `run()` method is _asynchronous_. This means that
+when we call `run()`, it will start processing in the background. Later on, we
+will see how we can take advantage of that to create pipelines.
```
krancour, line 265: Is it worth mentioning that `run()` is actually a request to the Brigade API server to _schedule_ the job for execution and that execution is subject to scheduling constraints (max concurrent jobs)? Is it worth further noting that `job.Run()` returns immediately after scheduling and the `await` keyword can (and usually should) be used to, instead, _block_ further script execution until the job has actually been executed and reached a terminal state?

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
+  let job = new Job("do-nothing", "debian", event);
+  await job.run();
+});
 
-  job.run()
-})
+events.process();
 ```
-[brigade-05.js](../../examples/brigade-05.js)
+[03-first-job](https://github.com/brigadecore/brigade/tree/v2/examples/03-first-job)
```
krancour, line 220: The example below doesn't actually line up with what's in the first job example-- which you've also used twice. Do we need another example that introduces jobs _before_ adding commands to jobs?

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
+  job.primaryContainer.arguments = ["My first job!"];
+  await job.run();
+});
 
-  job.run()
-})
+events.process();
 ```
-[brigade-06.js](../../examples/brigade-06.js)
+[03-first-job](https://github.com/brigadecore/brigade/tree/v2/examples/03-first-job)
```
krancour, line 303: See [here](https://github.com/brigadecore/brigade/pull/1629/files#r718631954). I think we used this example twice.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
 
-Tasks can be an arbitrary shell command _that is supported by the image you use_.
-Tasks are concatenated together and executed as one shell script. (In Linux, they
-are executed as `/bin/sh` commands, with `set -eo pipefail`.)
+A command can be anything that is supported by the job's image. In the example
+above, our command invokes the `echo` binary with the arguments supplied. For
+multiple commands, a common approach is for the `command` array to consist of
+one element such as `["/bin/sh"]` or `["/bin/bash"]` and then the first element
+of the `arguments` array be `"-c"`, followed by entries comprising the shell
+commands needed.
```
krancour, line 310: This lines up with what Docker, Docker Compose, and k8s all do. And though I felt there was a benefit to emulating familiar things, I have always struggled to explain this well. Is it worth it for us to take a peek at how those other tools explain command v arguments and what _they_ consider to be the common conventions. Perhaps we can get some insight into how to explain this better?

krancour: I think it's also worth mentioning that beyond a certain subjective degree of complexity, this gets unwieldy and we'd recommend writing a shell script and having the job just run the script.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
+[04-simple-pipeline](https://github.com/brigadecore/brigade/tree/v2/examples/04-simple-pipeline)
 
-The important new part is at the end. We have replaced this:
-
-```javascript
-hello.run()
-goodbye.run()
-```
+In this example we create two jobs (`my-first-job` and `my-second-job`). Each
+starts a Debian container and prints a message, then exits.
```
krancour, line 357: I think we should call out that the `await` keyword means job2 won't start until job1 completes, so these jobs implicitly execute in serial.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
-When a repository is checked out, it is stored by default in `/src`. So it is as if we
-started every job by doing a `git clone https://github.com/brigadecore/empty-testbed.git /src`.
-That means we can start our job knowing that we already have access to everything in our
-Git project.
+- `Job.sequence()` takes an array of runnables (e.g. a job or a group of
+  jobs) and runs them _in sequence_. A new runnable is started only when the
+  previous one completes. The sequence completes when the last runnable has
+  completed (or when any runnable fails).
+- `Job.concurrent()` takes an array of runnables and runs them all
+  _in parallel_. When run, all runnables are started simultaneously (subject to
```
krancour, line 426: Nit: There's a nuanced difference between "in parallel" and "concurrently." I'll refrain from waxing philosophical, but "concurrently" is the more general of the two terms, so I think we should prefer to use that when we can.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
+  scheduling constraints). The concurrent group completes when all Runnables
+  have completed.
```
krancour, line 428: Is it worth explicitly stating that a concurrent group is itself considered a success or failure on the basis of all of its jobs completing successfully?

Do we need to be more explicit about what happens when a job in a concurrent group fails?

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
-project. And from there on, we are working with the code from the repository.
+As both of these methods return a runnable, they can be chained. In the example
+above, `job1` and `job2` run _in sequence_, as do `jobA` and `jobB`, but both
+sequences are run _in parallel_.
 
-Being able to associated a Git repository to a project is a convenient way to provide
-version-controlled data to our Brigade builds. And it makes Brigade a great tool for executing
-CI pipelines, deployments, packaging tasks, end-to-end tests, and other DevOps tasks.
+This is the way script writers can control the order in which groups of jobs
+are run.
```
krancour, line 435: Worth mentioning a realistic usage of this?

A good example might be that if using Brigade to implement CI, you might wish to divide your checks into those that are resource intensive and those that are less so. Both groups can run the jobs within concurrently, but the groups themselves might be run in sequence, such that no resource intensive checks are executed unless/until all of the less resource intensive checks have passed.

docs/content/topics/scripting/guide.md
krancour: We should probably mention that `job.Run()` will throw an exception if the job fails.

We probably also need to mention (somewhere) that job success/failure is determined by its primary container's exit code.

We probably need to mention that same WRT the worker/script itself.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
-- `payload` contains any information that the hook sent when triggering the event. For
-  example, a GitHub push request generates [a rather large payload](https://developer.github.com/v3/activity/events/types/#pushevent). Payload is an unparsed string.
-
-When one event triggers another event, the triggered event also gets information about the
-thing that caused it. This is available on the `event.cause` field. The `after` and
-`error` events, for example, will be able to access the `cause` field and see the event
-that triggered them.
+From the event, we can find out what triggered the event, what data was sent
+with the event, and (if there is a repository) what repository commit we should
+be using.
```
krancour, line 533: Is the question of which commit relevant to someone writing a script?

By the time the script executes, the git initializer for the worker has already worked out what commit to use. Each job's git initializer (if applicable) will work that out as well-- all without the script author needing to be concerned with this.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
-But Brigade comes with a few built-in options that are useful in writing basic Brigade builds,
-and which don't require modifying your containers. In this part of the guide, we cover two
-built-in shared directories.
-
-### Build Storage (Shared Space)
 
-Each build gets its own shared storage. This is a file path that can be accessed by every
-job during the build, but which does not survive after the build has completed.
+Note that some event and project data should be treated with care. Things like
+`event.project.secrets` or event `event.worker.git.cloneURL` might not be the
```
krancour, line 626: s/or event/or even/

krancour: I'm not 100% clear on what the exposure is if a clone URL were leaked.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
+as on each job in the script that requires access to the workspace.
 
-  Group.runEach([one, two, three])
-})
-```
-[brigade-15.js](../../examples/brigade-15.js)
+As background, the decision to make this functionality opt-in as opposed to
+opt-out was based on the desire to have job results be cache-able and jobs
+themselves resumable/retryable, which can currently only be accomplished when
+they do not share a workspace.
```
krancour, line 642: I'm iffy on whether we should include this. You and I spent so much time on this that we understand this well. I'm trying to look at this through the eyes of a new user and I'm not sure I understood that.

Maybe it's better to cut this detail than to expand further? Honestly, I really don't know. Just thinking out loud.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
 
-  one.docker.enabled = true
+## Sidecar containers
 
-  one.run()
-})
-```
+Jobs can optionally be configured with one or more sidecar containers, which
+run alongside the job's primary container. All sidecar containers will be
+terminated a short time after the job's primary container completes.
```
krancour, line 698: Should we mention somewhere that the success/failure of a job is contingent only on the exit code of the primary container and is not (directly) affected by the sidecar containers?

krancour: Some other things that we probably need to note in this section:

1. All the containers are networked together such than processes listening for network connections in any one of them can be addressed by processes running in the others using the local network interface.

1. Brigade does not understand the relationship(s) between your containers and if, for instance, a process in the primary container should be delayed until some supplementary process in some sidecar container is up and running and listening for connections, then the script author needs to account for that themsevles.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,867 @@ aliases:
...
-```
-
-It is a good idea to call `toString()` on the `Result` object, since there is otherwise
-no guarantee about the type of data returned.
 
-With carefully constructed containers, you can get sophisticated and send structured
-data like JSON from one job to another. But remember that what is captured is the
-standard output (STDOUT) of the job, which is often where log data will also be sent.
-Sometimes it makes more sense to write structured files to the shared storage instead.
+### Accessing the host Docker socket
```
krancour, line 792: Given the increasing prominence of containerd on k8s nodes, which unlike Docker cannot be used to build images anyway, do you have any thoughts on whether we should just _cut_ this feature from 2.0 entirely?

Given that we would always encourage DinD over DooD and given that we might even encourage Kaniko (or similar) over DinD (and DooD), it's not as if there's any shortage of safer alternatives.

lmk what you think.

docs/content/topics/scripting/guide.md
vdice: Oh good point.  I'll remove the link from this one, and keep the link on the subsequent example which does match `03-first-job`.

docs/content/topics/scripting/guide.md
vdice: > Do we need to be more explicit about what happens when a job in a concurrent group fails?

What might you suggest here?

docs/content/topics/scripting/guide.md
vdice: Maybe if it were a private repo not intended to be known to others?

docs/content/topics/scripting/guide.md
vdice: Will cut; can re-intro when/if it proves necessary.

docs/content/topics/scripting/guide.md
vdice: I'm up for at least creating a ticket to track and discuss the proposal to cut it out 👍

docs/content/topics/scripting/guide.md
vdice: >  Perhaps we can get some insight into how to explain this better?

I propose we wait to see if/when we need to defend this decision.  For now, I think it's sufficient to just show how commands/args are constructed on a job.

vdice:
Thanks for all of the great feedback @krancour.  I think I've addressed all of the first round of comments in https://github.com/brigadecore/brigade/pull/1629/commits/261fec5d37a29548cd0af9943c1d7894290730ab.  PTAL when convenient.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,906 @@ aliases:
...
-### Adding Tasks to Jobs
+[Brigadier]: /topics/scripting/brigadier
+
+### Adding Commands to Jobs
 
-To make our Job do more, we can add tasks to it. A task is an _individual step to be run
-inside of the Job's container_.
+To make our Job do more, we can add a _command_ to it. A command can then be
+paired with a list of _arguments_. The command and arguments array is added to
+the job's _primaryContainer_, which is the container running the image supplied
```
krancour, line 294: Little bit of rewording for subject/verb agreement, maybe:

> The command and argument arrays are added to the job's _primaryContainer_...

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,906 @@ aliases:
...
+events.process();
+```
 
-So far we have looked at the brigade.js file, the event registry, and jobs and groups. As we
-advance our way through Brigade, we will next take a look at how `brigade.js` scripts can work
-with the BrigadeEvent and Project objects that are sent to every event handler.
+Being able to associated a Git repository to a project is a convenient way to
+provide version-controlled data to our Brigade scripts. And it makes Brigade a
+great tool for executing CI pipelines, deployments, packaging tasks, end-to-end
+tests, and other DevOps tasks.
```
krancour, line 554: I think I might add that even for use cases not otherwise requiring git, storing scripts and other configuration in a git repo is a recommended alternative to embedding complex scripts and configuration in the project definition.

docs/content/topics/scripting/guide.md
krancour: Fair point. I think that's one of those edge cases where one might consider standing up their own brigade instance. No need to get into that here. Current level of details is good.

docs/content/topics/scripting/guide.md
krancour: > I'm up for at least creating a ticket to track and discuss the proposal to cut it out

That sgtm. I'll do that. If we do it, it's a rare post-beta breaking change, but I think pre-GA, security is certainly defensible grounds for such a change.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,906 @@ aliases:
...
-can listen for. So what is the origin of these events?
+There are a few things to note about this script:
 
-To answer this question, we need a tiny bit of background knowledge about Brigade. Brigade has
-several components in its runtime. The _worker_ runs the JavaScript that we are writing
-here. The _controller_ starts and manages the _worker_. But there are also Brigade
-_gateways_.
+- It imports the `events` object from `@brigadecore/brigadier`. Almost all
+  Brigade scripts do this.
+- It declares a single event handler that says "on an 'exec' event, run this
```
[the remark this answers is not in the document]
krancour: Might have popped up in a few other places also.

krancour:
This is looking great. I think this is really, quite comprehensive and reads well.

docs/content/topics/scripting/guide.md
```diff
@@ -9,1401 +9,906 @@ aliases:
...
+  above, we use the image named `debian`, which is fetched
+  [straight from DockerHub](https://hub.docker.com/_/debian/).
+- the _event_ itself that this handler is for
 
-The image is a crucial part of the Brigade ecosystem. A container is created from this
-image, and it's within this container that we do the "real work". At the beginning, we
-explained that we think of Brigade scripts as "shell scripts for your cluster." When you
-execute a shell script, it is typically some glue code that manages calling one or more
-external programs in a specific way.
+The image is a crucial part of the Brigade ecosystem. A container is created
```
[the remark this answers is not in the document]
vdice: Yea, I'm not sure either.  May be carryover from previous doc... updated with suggestion.

krancour:
LGTM

krancour (approved):

Issue #1631, opened 2021-09-24
Add Brigadier README 


vdice:
Add a README in the v2/brigadier directory explaining Brigadier.  With this in place, Brigadier's [npm package site](https://www.npmjs.com/package/@brigadecore/brigadier) will also display this doc so that visitors can get more information on this library.

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1633

PR #1633, opened 2021-09-24
docs(brigadier): add README.md

vdice:
* Adds a README for the brigadier library

Closes https://github.com/brigadecore/brigade/issues/1631

### DIFF
--- a/v2/brigadier/README.md
+++ b/v2/brigadier/README.md
@@ -12,15 +12,15 @@ Brigadier supports writing scripts in either JavaScript or TypeScript.
 
 [![NPM](https://nodei.co/npm/@brigadecore/brigadier.png)](https://www.npmjs.com/package/@brigadecore/brigadier)
 
-Normally, the brigadier dependency is declared at the top of a Brigade script
-and will be installed when the Brigade Worker starts up before script
-execution, like so:
+Normally, the brigadier dependency is declared at the top of a Brigade script.
+The library itself is pre-loaded in the Brigade Worker:
 
 ```javascript
 const { events, Job } = require("@brigadecore/brigadier");
 ```
 
-You may also install brigadier to your environment with Yarn, NPM, etc.:
+To facilitate script development, you may also install brigadier to your
+environment with Yarn, NPM, etc.:
 
 ```console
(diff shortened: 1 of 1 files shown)


v2/brigadier/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+Brigadier supports writing scripts in either JavaScript or TypeScript.
+
+## Installation
+
+[![NPM](https://nodei.co/npm/@brigadecore/brigadier.png)](https://www.npmjs.com/package/@brigadecore/brigadier)
+
+Normally, the brigadier dependency is declared at the top of a Brigade script
+and will be installed when the Brigade Worker starts up before script
+execution, like so:
```
krancour, line 17: Not entirely accurate. In v1.x, brigadier is pre-installed on the worker image. In v2, a brigadier "polyfill" is pre-installed. This might be too much detail... or maybe not. I cannot decide.

Maybe we should say something like "you can install this library locally to facilitate script development and at runtime, this library (or a compatible replacement) will be pre-loaded?"

Needs a little more word-smithing.

v2/brigadier/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Versioning
+
+The 0.x Brigadier npm releases are compatible with
+[Brigade v1.x and under](https://github.com/brigadecore/brigade/tree/master).
+
+The 2.x brigadier npm releases are compatible with
+[Brigade v2.x](https://github.com/brigadecore/brigade/tree/v2).
+
+## Usage
```
krancour, line 37: These seem to be v2 examples. After acknowledging in the previous section that different versions of Brigadier are compatible with different versions of Brigade, perhaps it's incumbent to show examples here for both?

krancour: Or if we _don't_ want to show for both, should we at least clarify that these are v2 examples?

vdice:
Thanks @krancour.  How do the updates look in https://github.com/brigadecore/brigade/pull/1633/commits/30ad7a59cd20e4dda17cd98faf5956184eb8e503 ?

krancour (approved):
LGTM!

PR #1637, opened 2021-09-27
Support more than one brigade instance per cluster

### LINKED ISSUE #1630: Should be able to install more than one brigade per cluster
Or event more than one brigade per namespace.

This _might_ be a duplicate, but if it is, I haven't found the other issue yet...

The reasons why someone _might_ want to do this would boil down to not liking something about either our subscription model or our authorization model.

1. Any project can subscribe to any event. This means events should never contain secrets / sensitive information. (They really shouldn't anyway, on account of events being unencrypted at rest.) This is generally not a problem, as it's usually possible to only include opaque references to sensitive things in an event. Projects that are authorized to access those sensitive things will have the secrets necessary to grant them access already defined and securely stored.

    A good example of the above involves the GitHub gateway. Any project can consume events that originated from a private repository, but cannot access the code from such a repository without having the correct credentials.

    If, for some reason, someone is trying to accomplish something that really _cannot_ be done without including sensitive information in an event, standing up a one-off brigade to support such use cases could be their "nuclear option." Until / unless we can support more than one brigade per cluster, their nuclear option requires them to provision a whole new cluster. 🤮 

2. Our authz model says, with a few exceptions, that anyone who can read can read _anything_. (This does not extend to secrets.) Again, if this model proves undesirable for specific cases, the option to provision another brigade _without_ provisioning an entire new cluster would be helpful.

I know of two particular things that need to be addressed in order to support this:

1. k8s resources need to be labeled with a unique-within-the-cluster "installation ID." Selectors that monitor k8s resources, delete them, etc. should incorporate this. This is pretty easy to do because we can seamlessly generate a unique-within-the-cluster installation ID at install time by combining the namespace and the release name.

1. Certain RBAC-related resources are installed with a cluster scope. This prevents the brigade chart from being installed twice in one cluster, but I am confident that we could work around this with a flag that allows the installation of those particular resources to be skipped.

krancour:
Fixes #1630

This PR injects a unique-to-the-cluster "Brigade ID" into the apiserver and observer. Both of those incorporate this new metadata into all labels and label selectors.

Fluentd configs also get updated so that only logs from a given Brigade instance are sent to that Brigade instance's database.

### DIFF
--- a/charts/brigade/templates/apiserver/cluster-role-binding.yaml
+++ b/charts/brigade/templates/apiserver/cluster-role-binding.yaml
@@ -1,4 +1,3 @@
-{{- if .Values.rbac.enabled }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
@@ -14,4 +13,3 @@ subjects:
 - kind: ServiceAccount
   namespace: {{ .Release.Namespace }}
   name: {{ include "brigade.apiserver.fullname" . }}
-{{- end }}
--- a/charts/brigade/templates/apiserver/cluster-role.yaml
+++ b/charts/brigade/templates/apiserver/cluster-role.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.rbac.enabled }}
+{{- if .Values.rbac.installGlobalResources }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
--- a/charts/brigade/templates/apiserver/deployment.yaml
+++ b/charts/brigade/templates/apiserver/deployment.yaml
@@ -30,6 +30,8 @@ spec:
         args:
         - --logtostderr=true
         env:
+        - name: BRIGADE_ID
+          value: {{ .Release.Namespace }}.{{ .Release.Name }}
         - name: API_ADDRESS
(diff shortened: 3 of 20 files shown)


vdice (commented):
All looks good!  Only documentation-related comments:

- Can we update/remove the RBAC mention in the [security doc](https://github.com/brigadecore/brigade/blob/v2/docs/content/topics/operators/security.md) in this PR as well?
- Should we add a mention around the scenario where two or more Brigade instances are deployed to a k8s cluster?  (Primarily showing which chart value to toggle for the additional instances..)  Maybe in the [Deploy doc](https://github.com/brigadecore/brigade/blob/v2/docs/content/topics/operators/deploy.md) and/or other spots?

krancour:
I didn't think to update docs as part of this. Thanks for keeping me honest. I'll amend tomorrow.

krancour:
@vdice I've updated `security.md`. PTAL when convenient.

I also tried to find any spot where we might have previously said we were constrained to one Brigade instance per cluster, but I couldn't find one. Am I just not looking hard enough?

vdice (approved):
LGTM!

vdice:
> I also tried to find any spot where we might have previously said we were constrained to one Brigade instance per cluster, but I couldn't find one.

I think you're right -- the only instance I can recall was in the security.md doc...

PR #1644, opened 2021-09-28
docs(dependencies): v2 updates

vdice:
* Updates the Dependencies doc
* Adds a dependencies example

Closes https://github.com/brigadecore/brigade/issues/1634


docs/content/topics/scripting/dependencies.md
```diff
@@ -9,102 +9,123 @@ aliases:
...
+
+This approach works great for using dependencies that are not intended to be
+external packages, and which are located in the project repository. 
 
-Let's consider the following scenario: we have a JavaScript file located in `/local-deps/circle.js`, where `local-deps` is a directory at the root of our git repository. In our `brigade.js` file, we can use any exported method or variable from that package by simply using a `require` statement, just like in any other JavaScript project.
+Let's consider the following scenario: we have a JavaScript file located in
+`/local-deps/circle.js`, where `local-deps` is a directory at the root of our
+git repository. In our Brigade script, we can use any exported method or
+variable from that package by simply using a `require` statement, just like in
+any other JavaScript project.
```
krancour, line 128: The example might be more straightforward if the filed were located in the `.brigade/` directory beside the `brigade.js/ts`.

Your example works, but I think that keeping Brigade scripts, config, etc. "separate" from the rest of the project (i.e. tucked away in their own directory) leads to a tidier project layout. I think we should encourage that, even if only implicitly via demonstartion.

docs/content/topics/scripting/dependencies.md
```diff
@@ -9,102 +9,123 @@ aliases:
...
 
 ## Using local dependencies from the project repository
 
-Local dependencies are resolved using standard Node [module resolution](https://nodejs.org/api/modules.html#modules_all_together),
-with one change: the worker's `node_modules` directory is added as a fallback, so `brigade.js`—and any local dependencies—can resolve modules installed via `brigade.json`.
-This approach works great for using dependencies that are not intended to be external packages, and which are located in the project repository. 
+Local dependencies are resolved using standard Node [module resolution], with
+one change: the worker's `node_modules` directory is added as a fallback, so
+`brigade.js` - and any local dependencies - can resolve modules installed via
+`package.json`.
```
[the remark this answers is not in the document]
vdice: Eh, this is incorrect in v2, isn't it?  I think it was saying that a project could fallback on use of a worker's deps/node_modules -- but in v2 the node process is separate between project script and worker itself.

docs/content/topics/scripting/dependencies.md
vdice: That's fair.  I can certainly revise per your suggestion.  However, I can imagine scenarios where the Brigade project uses a git repo with local deps/js assets that may also be used by other software/CI tools/etc. associated with the repo.  In that case, I'd assume they'd want the local deps to be standalone and not tightly coupled to Brigade (i.e. not in the .brigade dir).  WDYT?

Issue #1646, opened 2021-09-29
Workers: More information regarding the Brigade Worker.


vdice:
Ref https://github.com/brigadecore/brigade/issues/1508

PR #1651, opened 2021-10-04
docs(workers.md): v2 updates

vdice:
* Updates the Workers doc

Closes https://github.com/brigadecore/brigade/issues/1646


docs/content/topics/scripting/workers.md
```diff
@@ -9,216 +9,117 @@ aliases:
...
+[long title]: /topics/project-developers/events#long-title
+[payload]: /topics/project-developers/events#payload
 
 ## Job Pod Names and Labels
 
-To be visible to the rest of Brigade (and related projects, such as Kashti) and
-recognizable as pods that implement a build's job(s), certain naming and
-labeling conventions must be adhered to when a worker spawns such pods.
+To be visible to the rest of Brigade, certain naming and labeling conventions
+must be adhered to for job pods spawned by the worker.
```
krancour, line 96: I think this whole section can go. It was applicable to v1 where workers created job pods directly. Now workers do this via API calls-- a big improvement because it means workers can't accidentally produce job pods that the rest of Brigade won't recognize as jobs.

docs/content/topics/scripting/workers.md
```diff
@@ -9,216 +9,117 @@ aliases:
...
+| `project.secrets` | The key/value map of [secrets] for the project |
+| `source` | The event [source] |
+| `type` | The event [type] |
+| `qualifiers` | The event [qualifiers] |
+| `labels` | The event [labels] |
+| `shortTitle` | A [short title] for the event |
+| `longTitle` | A [long title] for the event |
+| `payload` | The event [payload] |
+| `worker.apiAddress` | The address for the Brigade API server |
+| `worker.apiToken` | The token used for communicating with the Brigade API server |
```
krancour, line 75: Worth noting that these details can be used with the SDK of your choice to obtain a Brigade API client?

docs/content/topics/scripting/workers.md
```diff
@@ -9,216 +9,117 @@ aliases:
...
+| `worker.logLevel` | The log level for worker logs |
+| `worker.configFilesDirectory` | The directory in the worker where the Brigade config files may be found |
```
krancour, line 77: Should probably note that this is applicable when there's a git repo associated with the project and the path is relative to where the source code from that repo has already been mounted in the worker's container (`/var/vcs`).

vdice:
Docs only; merging despite yarn-audit failure.  Fix will be in a different PR.

Issue #1652, opened 2021-10-04
Scripting Guide - Advanced: Advanced examples for Brigade scripts.


vdice:

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1653

PR #1653, opened 2021-10-04
docs(advanced.md): v2 updates

vdice:
* Updates per v2

Closes https://github.com/brigadecore/brigade/issues/1652


docs/content/topics/scripting/advanced.md
```diff
@@ -37,188 +47,207 @@ function exec(e, p) {
     .then(() => {
         console.log("done");
     });
```
krancour, line 49: With the async/await example, you showed how to catch an error. Should we do the same here as well?

docs/content/topics/scripting/advanced.md
```diff
@@ -37,188 +47,207 @@ function exec(e, p) {
...
-};
+}
+
+events.process();
 ```
-[advanced-01.js](../../examples/advanced-01.js)
 
-In the example above, we use implicit JavaScript `Promise` objects for chaining two jobs, then printing `done` after the two jobs are run. Each `Job.run()` call returns a `Promise`, and we call that `Promise`'s `then()` method.
+In the example above, we use implicit JavaScript `Promise` objects for chaining
```
krancour, line 55: I'm questioning the "implicit" part of this statement. Runnables do literally return `Promise<void>`.

docs/content/topics/scripting/advanced.md
```diff
@@ -11,24 +11,34 @@ aliases:
...
-[JavaScript API](javascript.md).
+This guide provides some tips and ideas for advanced scripting. It assumes
+familiarity with [the scripting guide] and the [Brigadier API].
 
-## Using `async` and `await` to run Jobs
+[the scripting guide]: /topics/scripting/guide
+[Brigadier API]: /topics/scripting/brigadier
 
-Recent versions of JavaScript added a new way of declaring asynchronous methods, and then calling them. This way is compatible with promises. Brigade supports the new `async` and `await` decorators.
+## Promises and the `async` and `await` decorators
```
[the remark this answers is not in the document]
vdice: Just checked in w/ the mozilla js docs... they use "keyword" as well.  Updated.

docs/content/topics/scripting/advanced.md
vdice: I added a note to clarify that the same method can be used with either approach...

docs/content/topics/scripting/advanced.md
```diff
@@ -37,188 +47,207 @@ function exec(e, p) {
...
+
+events.process();
 ```
-[advanced-03.js](../../examples/advanced-03.js)
 
-In the example above, the second job (`j2`) will execute `exit 1`, which will cause the container to exit with an error. When `await j2.run()` is executed, it will throw an exception because `j2` exited with an error. In our `catch` block, we print the error message that we receive.
+In the example above, the second job (`j2`) will execute `exit 1`, which will
+cause the container to exit with an error. When `await j2.run()` is executed,
+it will throw an exception because `j2` exited with an error. In our `catch`
+block, we print the error message that we receive.
```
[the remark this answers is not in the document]
vdice: Good point.  Not intentional, but now that you mention it, I like the idea of leaving as is and adding clarification that while the j2 job will fail, the worker will not -- but that the script writer has full control on whether it should, etc.

Issue #1617, opened 2021-09-17
bug: duplicate jobs on retry


krancour:
In our dogfood cluster, this event is a retry of another. Note there are two of each job. This seems to be a bug,

```
$ brig event get -i 233c56f3-a269-4b38-b61c-073aa2d28bb2

ID                                  	PROJECT	SOURCE           	TYPE                 	AGE	WORKER PHASE
233c56f3-a269-4b38-b61c-073aa2d28bb2	brigade	brigade.sh/github	check_suite:requested	5m 	RUNNING     

Event "233c56f3-a269-4b38-b61c-073aa2d28bb2" jobs:

NAME             	STARTED	ENDED	PHASE    
validate-schemas 	1h     	1h   	SUCCEEDED
test-unit-js     	1h     	1h   	SUCCEEDED
lint-chart       	1h     	1h   	SUCCEEDED
yarn-audit       	1h     	1h   	SUCCEEDED
lint-js          	1h     	1h   	SUCCEEDED
validate-examples	1h     	1h   	SUCCEEDED
validate-examples	       	     	PENDING  
lint-go          	5m     	2m   	SUCCEEDED
lint-chart       	       	     	PENDING  
test-unit-go     	5m     	     	RUNNING  
validate-schemas 	       	     	PENDING  
test-unit-js     	       	     	PENDING  
yarn-audit       	       	     	PENDING  
lint-js          	       	     	PENDING
```

vdice:
Fixed in https://github.com/brigadecore/brigade/pull/1656

PR #1656, opened 2021-10-05
fix(retry): use custom EqualTo methods for determining equivalence

vdice:
* Adds equivalence methods to applicable structs to take full control of comparing JobSpecs
* This comparison was returning the wrong value when using reflect.DeepEqual in some cases (namely, on a retried event when the original and retried job spec matched) -- when this occurred, Brigade incorrectly saved an additional duplicate job to the retried event, resulting in duplicate listings.  We believe the source of the incorrect comparison result stems from comparing one job against another, where the latter that has been saved to (and fetched from) the backing mongo store; for some fields, bson marshal/unmarshal appears to return nil or empty where the in-memory comparison has the reverse.

Fixes https://github.com/brigadecore/brigade/issues/1617


v2/apiserver/internal/api/containers.go
```diff
@@ -23,3 +23,39 @@ type ContainerSpec struct {
 	// to be set within the OCI container.
 	Environment map[string]string `json:"environment,omitempty" bson:"environment,omitempty"` // nolint: lll
 }
+
+func (cs ContainerSpec) EqualTo(cs2 ContainerSpec) bool {
+	// Compare Command slices
+	if len(cs.Command) != len(cs2.Command) {
+		return false
+	}
```
krancour, line 31: Clever way to do it, since `len()` of a nil slice it 0. I like.

v2/apiserver/internal/api/containers.go
```diff
@@ -23,3 +23,39 @@ type ContainerSpec struct {
...
+	for i, command := range cs.Command {
+		if command != cs2.Command[i] {
+			return false
+		}
+	}
```
krancour, line 36: Can these really not be compared without iterating? May be the case. I cannot recall.

v2/apiserver/internal/api/jobs.go
```diff
@@ -127,6 +126,30 @@ type JobSpec struct {
...
+		if js.Host == nil && js2.Host == nil {
+			return js.PrimaryContainer.EqualTo(js2.PrimaryContainer) &&
+				js.TimeoutDuration == js2.TimeoutDuration
+		}
+		return false
+	}
+	// Return remaining field equivalence
+	return js.PrimaryContainer.EqualTo(js2.PrimaryContainer) &&
+		js.TimeoutDuration == js2.TimeoutDuration &&
+		js.Host.EqualTo(js2.Host)
```
krancour, line 150: This is the part I worry about...

If we add new fields in the future, it will be easy to forget to amend this bit of logic and I don't think we can necessarily count on the tests to fail in that case. So we could end up regressing on the bug this is meant to fix.

Is there some possibility that instead of comparing specific fields here by name, that we could nil out the fields we already evaluated for equality and use reflection to test equality of what's left?

v2/apiserver/internal/api/containers.go
vdice: Yes, I believe these slices must be iterated for comparison.

v2/apiserver/internal/api/jobs.go
vdice: That's a good idea.  I updated these methods to do so.  Let me know how it looks...

krancour:
Love how this turned out. LGTM

krancour (approved):

PR #1662, opened 2021-10-06
[v1] add a banner to the docs to denote version

flynnduism:
This PR puts a banner across the top of the docs site to explain that this is the v1 site, and also to provide a link to the v2 docs.

Screen Shot 2021-10-06 at 3 11 53 PM


flynnduism:
To be GDPR compliant, yes we should retain this banner because the site is using Google Analytics to track # of visitors. This article has a good overview of how cookie [compliance](https://plausible.io/blog/google-analytics-cookies) should be communicated - I will adjust this PR to ensure the user provides cookie consent

flynnduism:
This is updated with a better cookie banner - it will block any cookies from loading until the user consents to allowing cookies for Google Analytics. 

The top _"sticky"_ banner explains the Brigade version, and the cookie info moves to the bottom of the viewport. [Preview here](https://deploy-preview-1662--brigade-docs.netlify.app/).

PR #1663, opened 2021-10-06
[v2] add a banner to the docs to denote version

flynnduism:
Similar to #1662, this PR  puts a banner across the top of the docs site to explain that this is the v12site, and also to provide a link to the v1 docs.

Screen Shot 2021-10-06 at 3 21 20 PM


flynnduism:
Disabling cookie storage seems ineffective, the deploy preview still ends up setting cookies:

Screen Shot 2021-10-07 at 4 11 45 PM

Reverting this and putting the cookie consent back in, but in a different location so we keep the top of the page for project/version status.

PR #1671, opened 2021-10-08
build arm64 images in addition to amd64 images

### LINKED ISSUE #773: Provide ARM images

krancour:
Fixes #773


krancour:
@squillace confirmed these images work.

vdice (approved):

PR #1674, opened 2021-10-12
ci(brigade.ts): publish brigadier docs on v2 merge events

vdice:
Brigadier API doc publishing was added in https://github.com/brigadecore/brigade/pull/1612.  It added publishing of the docs on tag pushes.  Here, I'm proposing we also publish any time v2 is pushed to.  WDYT?


krancour:
@vdice can you amend and force push, please. The DCO check is stuck somehow.

Issue #1635, opened 2021-09-24
Worker: npm notices


vdice:
Noticing npm notice output when running a project script that uses a package.json for deps.

Here's output from an event created for the [npm example project]():

```console
 $ b event create -p npm -f

Created event "e4f314a0-3bbc-42f5-b453-38bd0f61716a".

Waiting for event's worker to be RUNNING...
2021-09-24T22:18:10.181Z INFO: brigade-worker version: 7118fc5
2021-09-24T22:18:10.186Z DEBUG: using npm as the package manager
2021-09-24T22:18:10.187Z DEBUG: found a package.json at /var/vcs/examples/07-npm/.brigade/package.json
2021-09-24T22:18:10.187Z DEBUG: installing dependencies using npm

added 4 packages, and audited 5 packages in 3s

found 0 vulnerabilities
npm notice
npm notice New minor version of npm available! 7.21.0 -> 7.24.1
npm notice Changelog: <https://github.com/npm/cli/releases/tag/v7.24.1>
npm notice Run `npm install -g npm@7.24.1` to update!
npm notice
2021-09-24T22:18:13.226Z DEBUG: path /var/vcs/examples/07-npm/.brigade/node_modules/@brigadecore/brigadier exists; deleting it
2021-09-24T22:18:13.233Z DEBUG: polyfilling @brigadecore/brigadier with /var/brigade-worker/brigadier-polyfill
2021-09-24T22:18:13.234Z DEBUG: found nothing to compile
2021-09-24T22:18:13.234Z DEBUG: running node brigade.js
Hello, dull-frog!
```

Specifically, we should address:
```
npm notice
npm notice New minor version of npm available! 7.21.0 -> 7.24.1
npm notice Changelog: <https://github.com/npm/cli/releases/tag/v7.24.1>
npm notice Run `npm install -g npm@7.24.1` to update!
npm notice
```

I assume this would be addressed by updating the Worker Dockerfile?

Additionally, should we consider silencing npm notices (if possible)?  I can see them being distracting/confusing for project developers using the default Brigade worker...

krancour:
Yea...we should definitely upgrade it and we should see what we can do about suppressing that warning.

PR #1677, opened 2021-10-12
chore(*): bump node image for worker/js components

vdice:
* Bumps the node image used for the Worker and running containerized JS tasks
* Adds `npm config set update-notifier false` to the Worker's Dockerfile to disable update notices from appearing on the npm/yarn run commands 

This addresses the issue mentioned in https://github.com/brigadecore/brigade/issues/1635 -- the `npm notice` logs are no longer appearing for `npm`-based projects.

Closes https://github.com/brigadecore/brigade/issues/1635


vdice:
> --update-notifier=false seems to work for both npm and yarn.

@krancour nice find!  Will add a commit adding both.

Issue #1628, opened 2021-09-23
Critical warning about events should be in docs


krancour:
I don't think we have a warning about this anywhere yet, but it should appear somewhere prominent.

Because Brigade v2 routes events to interested parties (projects) based on a _subscription model_, events should _never_ contain secrets / sensitive information. Always assume that _anyone_ in your cluster could be subscribed to _any_ event that a gateway creates.

In practice, this shouldn't be a difficult thing to overcome. Events can contain non-secret references to things that parties (projects) having appropriate secrets can access. By way of example, _anyone_ can subscribe to events from the GitHub gateway originating from _any_ repo-- even private ones-- but only projects having the correct secrets will ever be able to pull source from such a repo.

One potential location for this warning is the page that outlines how to create new gateways.

vdice:
Should I/we also add a mention in the security doc PR undergoing updates? https://github.com/brigadecore/brigade/pull/1609

krancour:
Yes. I think it's worth mentioning in both places.

PR #1680, opened 2021-10-13
docs(gateways.md): add note around events and sensitive information

vdice:
* Adds warning around events and sensitive data to the Gateways doc (much of the wording courtesy @krancour)
* Adds the Slack gateway to the gateway listings

Closes https://github.com/brigadecore/brigade/issues/1628


krancour:
Do we want to add that in cases where the threat model and subscription model conflict, operators can always choose to install and operate a private Brigade instance?

Issue #1418, opened 2021-05-25
feature: limit workers and jobs to specific nodes


krancour:
We do need the ability for operators to specify some worker/job scheduling constraints at install time.

This would have to include:

1. Optional node selectors that the API server would seamlessly add to worker and job pods to limit them to certain nodes.

1. Optional tolerations that the API server would seamlessly add to worker and job pods so they can run on nodes that have been specifically configured to repel non-brigade work.

There's some prior, relevant discussion of this in #1095.

vdice:
> I like the simple approach.

You know me, I'm nearly always in favor of keeping it simple and then iterating as needed.  Ok, thanks for the breakdown.  Will get started.

vdice:
Closed by https://github.com/brigadecore/brigade/pull/1681

PR #1681, opened 2021-10-14
feat(*): add support for worker/job nodeSelector and toleration

vdice:
* Adds support for node selector and toleration for worker and job pods at install time

Closes https://github.com/brigadecore/brigade/issues/1418


krancour:
Really awesome work here @vdice.

My only feedback was about the example comments in the values.yaml and that's dealer's choice.

LGTM'ing now and you can decide if you want to address or merge as is.

vdice:
@krancour how do the updates in https://github.com/brigadecore/brigade/pull/1681/commits/96b0ad1887b2f0c62d6df5063f0c6b19bf9e4062 look?

PR #1703, opened 2021-10-27
remove 2.0 proposal from v2 branch

krancour:
We can keep the 2.0 proposal in the `master` branch (which will eventually be renamed `v1`) so there's a clear record of what got us from v1 to v2.

This `v2` branch, however, (which will eventually be renamed `main`) has no real need for this proposal to live on in perpetuity.


vdice (approved):

krancour:
We're having some issues with the Artemis image build, but this PR is for sure not responsible for those. Maintainers have previously agreed that the full test suite doesn't need to pass for doc changes, so I am merging this.

PR #1704, opened 2021-10-27
update logo in docs

krancour:


vdice (approved):

krancour:
We're having some issues with the artemis-build, but this PR is for sure not responsible for those. Maintainers have previously agreed that the full test suite doesn't need to pass for doc changes, so I am merging this.

PR #1706, opened 2021-10-27
add several more useful linters-- and remediate any issues they found

### LINKED ISSUE #1696: Enable more linters in golangci-lint
I was thinking it would be useful to enable the following linters to catch bugs and inconsistencies in the codebase:

- `gosec`
- `exportloopref`
- `bodyclose`
- `forcetypeassert`
- `unconvert`

krancour:
Fixes #1696 

This PR is broken up with one new linter + applicable changes per commit.


v2/tests/common.go
```diff
@@ -97,10 +96,9 @@ func assertWorkerPhase(
 	}
 }
 
-// nolint: golint
 func assertJobPhase(
 	t *testing.T,
-	ctx context.Context,
+	ctx context.Context, // nolint: revive
```
vdice, line 101: Curious as to the (ignored) linter issue here?

vdice (approved):
LGTM, just one area where I was curious to understand the linting issue...

v2/tests/common.go
krancour: Neither golint not revive like it when a function takes a `context.Context` as an arg and it's not the _first_ arg in the list. Generally, I think that's a good rule. For this case, I think `*testing.T` is the more appropriate thing to come first in the argument list.

Note that the `// nolint: golint` previously on line 100 was also meant to bypass this same rule. I've just moved it down a few lines to make this exception a little closer to the broken rule.

PR #1712, opened 2021-11-04
update logger Dockerfiles

krancour:
The main point of this PR is to reduce build time for the Windows-based logger image by starting from an image that already has most of what we need. We didn't do this initially because the base image this PR uses didn't exist until relatively recently.

At the same time, I've updated various gems/fluentd plugins on the Windows-based logger image to match the versions used on the Linux-based logger image. The latter were updated fairly recently and it was an oversight that the Windows-based image didn't receive the same changes.

Last, both the Windows-based and Linux-based images now use `fluent/fluentd` images as their base instead of `fluentd`. `fluentd` trails `fluent/fluentd` by a couple major versions and also does not offer Windows-based images.

### DIFF
--- a/v2/logger-linux/Dockerfile
+++ b/v2/logger-linux/Dockerfile
@@ -1,7 +1,7 @@
 # This Dockerfile builds a custom Fluentd image that layers in the Fluentd
 # plugins we rely upon on top of a standard Fluentd image.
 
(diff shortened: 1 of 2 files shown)


krancour:
This cut build time for the Windows-based logger from ~13 mins to ~5.

vdice (approved):
Nice refinements!

Issue #1713, opened 2021-11-04
need a windows build of the git-initializer


krancour:
Without this, there's a lot of functionality that will not work with Windows.

PR #1715, opened 2021-11-04
build windows git initializer

krancour:
First of at least two PRs collectively aimed at addressing #1713 

A follow-up will make chart and code changes so that the windows variant of this image is applied as necessary.

This PR is only focused on getting the image to build.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -12,19 +12,19 @@ and well-documented pattern colloquially called "Docker in Docker" or simply
 
 Building a Windows-based OCI image inside a Windows-based OCI container is
 _not_, as of this writing, a common, well-understood, or well-documented
-pattern. Building Brigade's one Windows-based OCI image, therefore, requires the
-use of a physical Windows machine or virtual machine having an appropriate
+pattern. Building Brigade's Windows-based OCI images, therefore, requires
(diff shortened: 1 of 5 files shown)


.github/workflows/README.md
```diff
@@ -12,19 +12,19 @@ and well-documented pattern colloquially called "Docker in Docker" or simply
 
 Building a Windows-based OCI image inside a Windows-based OCI container is
 _not_, as of this writing, a common, well-understood, or well-documented
-pattern. Building Brigade's one Windows-based OCI image, therefore, requires the
-use of a physical Windows machine or virtual machine having an appropriate
+pattern. Building Brigade's two Windows-based OCI images, therefore, requires
```
vdice, line 15: Wonder if we want to just simplify to 'Building Brigade's Windows-based OCI images' (here and elsewhere)... unless we're pretty certain it'll now be capped at the two?

vdice (approved):
LGTM

.github/workflows/README.md
krancour: Pretty sure it's limited to these two, but I like your suggestion regardless. Changing.

krancour: All set.

PR #1723, opened 2021-11-11
change default service type and update docs

### LINKED ISSUE #1539: docs: narrow down who the quickstart's audience is
The quickstart talk to two disparate groups of readers-- those kicking the tires on a local cluster and those (presumably deploying a more robust setup) on a remote cluster.

We should narrow down who the quickstart is for and simplify the docs to avoid confusion. At the same time, docs focused on other scenarios can be moved elsewhere in the doc tree.

In particular, docs for local tire kickers should revolve around a non-LoadBalancer service, while docs for people deploying to remote clusters should revolve around a LoadBalancer service without ingress, or a ClusterIP service _with_ ingress.

### LINKED ISSUE #1724: chart: default service type for api server should be ClusterIP
It's been `LoadBalancer` until now. That makes sense if the chart's default configuration were optimized for production, but all other aspects of the chart are optimized for getting started quickly in a local, development-grade cluster.

We should change it to `ClusterIP` and provide appropriate guidance about when and how to change this in our docs that cover production-grade deployments.

A significant benefit to this change will be that the QuickStart can start using `--wait` on the `helm install` command. That didn't generally work before because in a local, development-grade cluster, a service of type `LoadBalancer` is never "ready" for lack of ability to acquire a public IP.

Addressing this will streamline the entire getting started process.

krancour:
Fixes #1724 
Fixes #1539 

Initially, this PR set out only to resolve #1724, but complementary doc changes had me touching the QuickStart, installation guide, and production deployment guide. It seemed as good a time as any to resolve overlap between those three guides, expand them, and narrow the focus of each to a particular audience.


vdice (approved):
Only very minor suggestions added; otherwise, LGTM!

PR #1725, opened 2021-11-12
chart: upgrade mongodb

krancour:
This PR upgrades to the latest MongoDB subchart.

There are a lot of changes to that chart since we last upgraded it and some are "breaking-ish." (Lots of deprecated fields.)

It want to settle this now before we go RC/GA because if the subchart has changes of this magnitude in the future, we aren't going to want to take the upgrade because of how it will impact our own users who have MongoDB configuration embedded in their Brigade deployment's `values.yaml`.

Note to reviewer: Most of the changes are to `values.yaml`. The subchart's `values.yaml` was added wholesale to our own `values.yaml` and minimally modified for our purposes. (e.g. Specified creation of "`brigade` database and username/password.)

All the other changes to the chart, just adapt to those `values.yaml` changes. For instance, where we once accessed fields like `{{ Values.mongodb.auth.username }}`, we now use `{{ index .Values.mongodb.auth.userrnames 0 }}`

### DIFF
--- a/charts/brigade/Chart.lock
+++ b/charts/brigade/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: mongodb
   repository: https://charts.bitnami.com/bitnami
-  version: 10.3.7
-digest: sha256:0370ef4db426321d173bf80ae4e8a5e6b7152bdf38cc3bfbe6d31a3c78fdb3b6
-generated: "2021-02-10T10:05:38.297831-05:00"
+  version: 10.29.2
+digest: sha256:8767dceb9c2c808c8a089b700f3287727c8fb48c1be288c5486e08d228030839
+generated: "2021-11-12T09:50:47.56396-05:00"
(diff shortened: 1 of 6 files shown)


vdice (approved):
One question: Do we want to trim the amount of mongodb sub-chart config in the values.yaml to those that are reasonably common to change in context w/ Brigade?  I think we did the same w/ the Artemis sub-chart.  We'd have a comment/link to the full upstream chart's values.yaml for users who desire more customized configuration.  Using this opportunity to suggest; doesn't need to be in this PR if we think this is a good idea.

LGTM!

krancour:
> Do we want to trim the amount of mongodb sub-chart config in the values.yaml to those that are reasonably common to change in context w/ Brigade?

I vaguely recall us discussing this before and I don't remember where we left it.

In general, I'm in favor of that, and do think we could make sensible choices about what is relevant and what isn't, but it's not something I would want to rush through at the 11th hour. If we want to pair this down in the future, it should be doable without breaking changes.

> I think we did the same w/ the Artemis sub-chart

fwiw, there was no reasonable Artemis chart in existence. I wrote that from scratch (with some inspiration from someone else's prior attempt at this). If the number of options there are smaller, it's only because I only built what we needed.

PR #1742, opened 2021-11-15
switch our own builds to stable api

krancour:
The dogfood cluster is now running a custom build from the head of the `v2` branch -- which is close to what we will release as rc.1.

This PR updates the project definition file accordingly. Updating this file doesn't really have any bearing on any of our automation, but it does impact our ability to recreate the project from this definition in the future.

### DIFF
--- a/.brigade/project.yaml
+++ b/.brigade/project.yaml
@@ -1,5 +1,5 @@
-# yaml-language-server: $schema=https://raw.githubusercontent.com/brigadecore/brigade/v2/v2/apiserver/schemas/project.json
-apiVersion: brigade.sh/v2-beta
+# yaml-language-server: $schema=https://schemas.brigade.sh/schemas-v2/project.json
+apiVersion: brigade.sh/v2
 kind: Project
 metadata:
   id: brigade


krancour:
Welp... CI is busted because I forgot that the GitHub gateway doesn't yet speak the stable API.

Working on it.

.brigade/project.yaml
```diff
@@ -1,5 +1,5 @@
-# yaml-language-server: $schema=https://raw.githubusercontent.com/brigadecore/brigade/v2/v2/apiserver/schemas/project.json
-apiVersion: brigade.sh/v2-beta
+# yaml-language-server: $schema=https://schemas.brigade.sh/schemas-v2/project.json
```
vdice, line 1: Nice.  Is this URL served via a DNS entry pointing to the raw GH URL?  Or do we upload schemas somewhere and serve?

vdice (approved):

.brigade/project.yaml
krancour: DNS tricks would only get us so far-- they'd let us swap `raw.githubusercontent.com` for `schemas.brigade.sh`, but wouldn't let us improve the path.

I created an Azure storage account for schemas.

Refreshing the contents of that on release isn't automated yet, but that is for sure something we can do in the near future.

krancour: btw... I wanted `https://schemas.brigade.sh/v2/project.json` but annoyingly, v2 is not a valid bucket name because it's too few characters. 🤷‍♂️

radu-matei (approved):

PR #1745, opened 2021-11-17
chart: refactor for more stable upgrades

krancour:
See comments at the top of `charts/brigage/templates/deployments.yaml` for the full explanation of this PR.


vdice (commented):
Thanks for the extended comments/explanation around the changes -- and I like how they are paired with the source so that others can reference.

The biggest drawback that I can see (and that we've discussed offline) is that we break the pattern of a tidy "one resource to one yaml file" approach that we usually like to follow (and that we still of course follow elsewhere in the chart).  So, I've been brainstorming on ways we might be able to get back to that, but the best I can come up with is this:

Would we be able to break up the current `deployments.yaml` into yaml files that grouped related resources together, under their component folders?

For instance, take the apiserver.  Could the templating logic for the values related to the apiserver (root password, TLS material) be paired with the apiserver secrets and the apiserver deployment in one yaml file under `templates/apiserver`, while the templating logic for the values related to the observer component (api token) would be paired with the observer secret and deployment in one yaml file under `templates/observer` (and so on)?

I think I see the one hurdle preventing this, and that is the addition of checksum annotations for most component secrets on the main apiserver deployment.  But the apiserver deployment [didn't used to have these](https://github.com/brigadecore/brigade/blob/v2/charts/brigade/templates/apiserver/deployment.yaml#L19-L23) (only tracked apiserver secrets).  What is the need for the apiserver deployment tracking extraneous secrets (observer api token, artemis pw, etc.)?  Would we want the apiserver to bounce when, say, the artemis pw is updated?

krancour:
> For instance, take the apiserver. Could the templating logic for the values related to the apiserver (root password, TLS material) be paired with the apiserver secrets and the apiserver deployment in one yaml file under templates/apiserver, while the templating logic for the values related to the observer component (api token) would be paired with the observer secret and deployment in one yaml file under templates/observer (and so on)?

I would find that more palatable as well.

> I think I see the one hurdle preventing this, and that is the addition of checksum annotations for most component secrets on the main apiserver deployment.

Indeed. You identified that correctly. The API tokens used by the observer and scheduler for communication with the API server are, essentially, shared secrets, so there are multiple deployments dependent on the values of those secrets.

> But the apiserver deployment didn't used to have these (only tracked apiserver secrets).

An oversight. It _should_ have had them, _but..._

Since those tokens don't really change once initially generated, maybe we _could_ get away with returning to not using the checksum trick with those in the API server's Deployment. That would allow us to at least move all observer and scheduler related resources back to the appropriate folders. The issue with that, however, is that if and when we introduce a way of rolling those tokens (for instance, allowing them to be manually specified in `values.yaml`, for instance) we'd end up right back where we are now.

> Would we want the apiserver to bounce when, say, the artemis pw is updated?

Yes, otherwise it will lose connectivity.

vdice (approved):
Definitely glad this fixes the issue of bouncing pods unnecessarily in certain cases.  Thanks for linking to the Helm issue.  Let's get this in and iterate/refine as it becomes possible.  LGTM!

PR #1748, opened 2021-11-19
release automation: fix windows image publishing

krancour:
This PR remediates the only issue we encountered with release automation during the rc.1 release.

Correct tags were not applied to the images when they were built, so the pushes, which did use correct tags, failed.

A reminder to any reviewer... we are currently leaning on GitHub Actions only for building and publishing our very few Windows-based Docker images because it doesn't yet seem possible to build a Windows-based image inside a container (ala Docker-in-Docker) yet. So we need a Windows VM to accomplish this and that rules out doing it with Brigade itself. GitHub Actions has us covered.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -13,9 +13,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Build Windows logger agent
-        run: make build-logger-windows
+        run: $env:VERSION=$env:GITHUB_REF.substring(10, $env:GITHUB_REF.length-10); make build-logger-windows
       - name: Build Windows git initializer
-        run: make build-git-initializer-windows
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -13,9 +13,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Build Windows logger agent
-        run: make build-logger-windows
+        run: $env:VERSION=$env:GITHUB_REF.substring(10, $env:GITHUB_REF.length-10); make build-logger-windows
```
krancour, line 16: Even though this `$env:VERSION=$env:GITHUB_REF.substring(10, $env:GITHUB_REF.length-10)` clause repeats a total of four times in this file, this seems to be the most straightforward way of doing this.

I attempted [this approach](https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-an-environment-variable) and didn't have much luck.

vdice (approved):

PR #1753, opened 2021-11-30
s/v2/main/ and other similar changes where applicable

krancour:
This PR straightens out links that break as a result of v2 being pushed onto the main branch (and the existing v2 branch presumably not being long for this world now).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,42 +1,38 @@
-# Brigade 2: Event-based Scripting for Kubernetes
+# Brigade: Event-based Scripting for Kubernetes
 
-![build](https://badgr.brigade2.io/v1/github/checks/brigadecore/brigade/badge.svg?appID=99005&branch=v2)
+![build](https://badgr.brigade2.io/v1/github/checks/brigadecore/brigade/badge.svg?appID=99005&branch=main)
 [![slack](https://img.shields.io/badge/slack-brigade-brightgreen.svg?logo=slack)](https://kubernetes.slack.com/messages/C87MF1RFD)
 
 <img width="100" align="left" src="logo.png">
 
-Brigade 2 has been lovingly re-engineered from the ground up. We believe we've
-remained faithful to the original vision of Brigade 1.x, and as such, much
-general knowledge of Brigade 1.x can be carried over.
+Brigade is a full-featured, event-driven scripting platform built on top of
+Kubernetes. It integrates with many different event sources, more are always
+being added, and it's easy to create your own if you need something specific.
+And the best part -- Kubernetes is well-abstracted so even team members without
+extensive Kubernetes experience or without direct access to a cluster can be
+productive.
 
 <br clear="left"/>
 
-_But we've also learned a lot from Brigade 1.x._ Brigade 2 has been designed,
-_explicitly_ to reduce the degree of Kubernetes knowledge required for success.
-While Brigade 1.x was hailed as "event driven scripting for Kubernetes," Brigade
-2 is "event driven scripting (for Kubernetes)." Moreover, great care has been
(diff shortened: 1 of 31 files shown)


docs/themes/techdoc-brigade/layouts/partials/header.html
```diff
@@ -1,5 +1,5 @@
 <header>
-<section class="sticky-banner">You are viewing docs for Brigade v2. Click <a href="https://docs.brigade.sh/">here for v1 docs</a>.</section>
+<section class="sticky-banner">You are viewing docs for Brigade v2. Click <a href="https://v1--brigade-docs.netlify.app/">here for v1 docs</a>.</section>
```
vdice, line 2: A reminder to update Netlify to deploy v1 docs to this URL.  At time of writing, it doesn't appear to be up yet.

README.md
```diff
@@ -26,7 +26,7 @@ While comprehensive documentation remains a work in progress, our [quickstart]
 is well-polished and we refer users wishing to test drive Brigade 2 to
```
vdice, line 26: I wonder if we want to make a few additional updates to this README, including (potentially):

 * Revising/rewording `While comprehensive documentation remains a work in progress` (Naturally, more docs are always helpful, but I think we're at a state where most docs are in and updated.
 * Revising/rewording `our quickstart is well-polished and we refer users wishing to test drive Brigade 2 to that documentation.` -- Do we drop '2' now that it is the de facto default?  Defer to your judgement here... I suppose it could also just stay the same/similar...
 * A link/mention to one or both of the v1 GH tree and v1 docs site now that they aren't the defaults

vdice (commented):
Looking great!  Only a few thoughts/suggestions on some README updates...

docs/themes/techdoc-brigade/layouts/partials/header.html
krancour: Thanks for the reminder. I will re-enable and force push so previews of these changes get built.

README.md
krancour: For sure. I can include that in this PR. I think there will be some follow-ups to this PR as well if/when we identify more spots like this where we talk about v2 as something new/beta, under "heavy development," etc.

docs/themes/techdoc-brigade/layouts/partials/header.html
krancour: Re-enabled.

README.md
krancour: I think I've applied all this feedback now. wdyt?

docs/themes/techdoc-brigade/layouts/partials/header.html
vdice: Can you double-check this?  Not blocking this PR as it's a Netlify task, but I'm seeing 'Not Found - Request ID: <request id>' when I go to https://v1--brigade-docs.netlify.app/

README.md
```diff
@@ -1,42 +1,38 @@
...
+Ready to get started? Check out our
+[QuickStart](https://docs.brigade.sh/intro/quickstart/) for comprehensive
+instructions.
 
-## The Brigade 2 Ecosystem
+## The Brigade Ecosystem
 
-Brigade 2's all new API lowers the bar to creating all manner of peripherals--
-tooling, event gateways, and more.
+Brigade's API make it easy to create all manner of peripherals-- tooling, event
```
vdice, line 29: nit: should it be `makes it easy`?

vdice (approved):
Updates to the README look great.  Only non-blocking comments.  LGTM!

README.md
krancour: Yerp

PR #1754, opened 2021-11-30
s/master/main/ and s/master/v1/ where applicable

krancour:
This PR straightens out links that break as a result of v1 moving from the master branch to the v1 branch.


docs/content/topics/developers.md
```diff
@@ -60,7 +60,7 @@ $ git clone https://github.com/brigadecore/brigade $GOPATH/src/github.com/brigad
 $ cd $GOPATH/src/github.com/brigadecore/brigade
 ```
 
-**Note**: this leaves you at the tip of **master** in the repository where active development
+**Note**: this leaves you at the tip of **main** in the repository where active development
```
vdice, line 63: Do we want to get even more explicit here and mention `where active development on Brigade v2.x is happening`?  And/or offer the reminder that the tip of v1 development (even though about to be in maintenance mode) can be checked out via the `v1` branch.

docs/readme.md
```diff
@@ -1,6 +1,6 @@
 # Brigade Docs
 
-* Browse the [docs on GitHub](https://github.com/brigadecore/brigade/tree/master/docs/content)
+* Browse the [docs on GitHub](https://github.com/brigadecore/brigade/tree/v1/docs/content)
 * Browse the [docs website at docs.brigade.sh](https://docs.brigade.sh)
```
vdice, line 4: This URL should be updated to `https://v1--brigade-docs.netlify.app/`, right?

docs/content/topics/developers.md
krancour: I started out on that path, initially, and then second guessed it. My worry was that I'd rather not promote unsolicited investment in v1. Any thoughts?

vdice: sgtm 👍

docs/readme.md
krancour: Good catch.

krancour: Fixed. And fyi, longer term, I will also work out a nicer URL than `https://v1--brigade-docs.netlify.app/`.

vdice (approved):

PR #1764, opened 2021-12-10
port integration tests to brigade 2 + cleanup

krancour:


v2/tests/common.go
```diff
@@ -1,235 +0,0 @@
...
-	val := os.Getenv(name)
-	if val == "" {
-		t.Fatalf(
-			"value not found for required environment variable %s",
-			name,
-		)
-	}
-	return val
-}
-
```
krancour, line 57: This was available in brigade-foundations.

v2/tests/common.go
```diff
@@ -1,235 +0,0 @@
...
-				)
-				return
-			}
-		case err := <-errCh:
-			t.Fatalf("error encountered watching worker status: %s", err)
-		case <-timer.C:
-			t.Fatal("timeout waiting for worker to reach a terminal phase")
-		}
-	}
-}
```
krancour, line 97: I moved this closer to where it's used.

v2/tests/common.go
```diff
@@ -1,235 +0,0 @@
...
-				)
-				return
-			}
-		case err := <-errCh:
-			t.Fatalf("error encountered watching job status: %s", err)
-		case <-timer.C:
-			t.Fatal("timeout waiting for job to reach a terminal phase")
-		}
-	}
-}
```
krancour, line 136: I moved this closer to where it is used.

v2/tests/common.go
```diff
@@ -1,235 +0,0 @@
...
-) {
-	assertLogs(
-		t,
-		ctx,
-		client,
-		e,
-		&core.LogsSelector{Container: "vcs"},
-		wantLogs,
-	)
-}
```
krancour, line 153: This function was called in only one or two places and it's not significantly more verbose to call `assertLogs()` directly with the correct `LogSelector`, so I've deleted this function.

v2/tests/common.go
```diff
@@ -1,235 +0,0 @@
...
-) {
-	assertLogs(
-		t,
-		ctx,
-		client,
-		e,
-		&core.LogsSelector{},
-		wantLogs,
-	)
-}
```
krancour, line 170: This function was called several place, but it's not significantly more verbose to call `assertLogs()` directly with the correct `LogSelector` (which can even be `nil` for this case), so I've deleted this function.

v2/tests/common.go
```diff
@@ -1,235 +0,0 @@
...
-) {
-	assertLogs(
-		t,
-		ctx,
-		client,
-		e,
-		&core.LogsSelector{Job: job},
-		wantLogs,
-	)
-}
```
krancour, line 188: This function was called in several places, but it's not significantly more verbose to call `assertLogs()` directly with the correct `LogSelector`, so I've deleted this function.

v2/tests/common.go
```diff
@@ -1,235 +0,0 @@
...
-		case <-ctx.Done():
-			break
-		}
-
-		// log and err channels empty; we haven't found what we're looking for
-		if logEntryCh == nil && errCh == nil {
-			t.Fatalf("logs do not contain expected string %q", wantLogs)
-		}
-	}
-}
```
krancour, line 235: Moved this closer to where it's used.

.brigade/brigade.ts
```diff
@@ -285,6 +285,33 @@ const publishBrigadierDocsJob = (event: Event, version?: string) => {
...
+    `sleep 20 && make hack-new-kind-cluster hack test-integration`
+  ]
+  job.sidecarContainers.docker = new Container(dindImg)
+  job.sidecarContainers.docker.privileged = true
+  job.sidecarContainers.docker.environment.DOCKER_TLS_CERTDIR=""
+  job.timeoutSeconds = 30 * 60
+  return job
+}
+jobs[testIntegrationJobName] = testIntegrationJob
+
```
krancour, line 314: Overall strategy here is to do DinD just like we do for image builds. We launch the Docker daemon in a sidecar container. That means KinD runs in the sidecar.

I'm firing up an image registry (the same way we do with local/hack workflows) and making sure all KinD nodes use that. Using `kind load...`, for some reason, was not working. We kept getting `ImagePullBackoff`, but the other method worked with no problems.

.brigade/brigade.ts
```diff
@@ -310,7 +337,8 @@ async function runSuite(event: Event): Promise<void> {
       buildWorkerJob(event),
       buildBrigadierJob(event),  
       buildCLIJob(event)
-    )
+    ),
+    testIntegrationJob(event)
```
krancour, line 341: Only test this after all of the less resource-intensive tests have passed.

I might revisit this decision in the future. Since everything gets _re-built_ in this step, maybe we can skip some of the preceding build jobs, knowing that this job covers our bases.

.brigade/brigade.ts
vdice: Approach sounds good to me.  fyi, when using 'kind load', image pull policy needs to be 'IfNotPresent' -- otherwise, kind always tries to pull images remotely, even if images have been loaded in.

vdice (approved):
Nice cleanup!  Glad to see this ported to v2.

.brigade/brigade.ts
krancour: Oooooooh. That makes sense! I may revisit this in a follow-up then.

PR #1766, opened 2021-12-14
use brigade 2 for brigade 1.x ci/cd

krancour:
Brigade v1 has a six month maintenance window.

If we can get Brigade v1's CI/CD (as well as CI/CD for a small number of other projects) to run on Brigade 2, we can shut down our own Brigade v1 instance much sooner than that window closes.

This is mainly being done in the interest of resource/cost savings.


.brigade/brigade.ts
```diff
@@ -0,0 +1,224 @@
+import { events, Event, Job, ConcurrentGroup, SerialGroup, Container } from "@brigadecore/brigadier"
+
+const releaseStoragePath = "/release-assets"
+
+const goImg = "brigadecore/go-tools:v0.1.0"
```
vdice, line 5: Should we bump to the latest release tag (v0.5.0)?  Or perhaps this is the best fit with the go version for v1?

krancour: I'm unopposed to doing that in a separate PR. For now I was doing my best to replicate the the old process using Brigade 2 and without changing anything I didn't have to.

vdice (approved):

PR #1770, opened 2021-12-15
fix permissions requirments for creating events

### LINKED ISSUE #1769: bug: Overly restrictive permissions check when creating project-specific events
At present, an `EVENT_CREATOR` with appropriate scope (source) can create events that are not project-specific.

To create project-specific events requires the `PROJECT_USER` role for the project in question.

This is too restrictive for gateways that rightfully wish to emit and event that is only for consumption by a given project.

A very good example of this is a `check_run:rerequested` webhook from GitHub, which becomes a `check_run:rerequested` event sent to Brigade. Because the check in question would have been created _by_ a specific Brigade project, that same specific Brigade project should be the only recipient of that `check_run:rerequested` event.

krancour:
Fixes #1769

### DIFF
--- a/.brigade/brigade.ts
+++ b/.brigade/brigade.ts
@@ -291,18 +291,18 @@ const testIntegrationJob = (event: Event) => {
   job.primaryContainer.sourceMountPath = localPath
   job.primaryContainer.workingDirectory = localPath
   job.primaryContainer.environment = {
-    "DOCKER_REGISTRY": "localhost:5000",
     "SKIP_DOCKER": "true",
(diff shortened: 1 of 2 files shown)


vdice (approved):
Changes LGTM; pending CI passing.

krancour:
Worker build keeps timing out. I'm speculating that there may be something going on with npmjs.org.

/brig run

krancour:
> I'm speculating that there may be something going on with npmjs.org.

And no sooner did I write that than the status page finally reflected the issue.

PR #1772, opened 2021-12-16
brigadier: allow jobs to be marked as fallible

### LINKED ISSUE #1768: feature: mark job as "fallible"
We might be able to come up with better terminology, but it would be useful in Brigadier to be able to mark a job as able to fail _without_ that failure compelling the entire worker process to fail.

There are ways of addressing this, but they're inconvenient. It would be better if this were an easily exercised option.

krancour:
Fixes #1768

This sample project demonstrates a job that fails and a worker that succeeds regardless.

```
# yaml-language-server: $schema=https://schemas.brigade.sh/schemas-v2/project.json
apiVersion: brigade.sh/v2
kind: Project
metadata:
  id: experiment
description: My new Brigade project
spec:
  eventSubscriptions:
  - source: brigade.sh/cli
    types:
    - exec
  workerTemplate:
    logLevel: DEBUG
    defaultConfigFiles:
      brigade.ts: |
        import { events, Job } from "@brigadecore/brigadier"

          let job = new Job("hello", "debian:latest", event)
          job.primaryContainer.command = ["false"]
          job.fallible = true
          await job.run()
        })

        events.process()
```

### DIFF
--- a/v2/brigadier-polyfill/src/jobs.ts
+++ b/v2/brigadier-polyfill/src/jobs.ts
@@ -51,35 +51,49 @@ export class Job extends BrigadierJob {
       const statusStream = jobsClient.watchStatus(this.event.id, this.name)
       statusStream.onData((status: core.JobStatus) => {
         this.logger.debug(`Current job phase is ${status.phase}`)
-        switch (status.phase) {
-        case core.JobPhase.Aborted:
-          reject(new Error(`Job "${this.name}" was aborted`))
-          break
-        case core.JobPhase.Canceled:
-          reject(new Error(`Job "${this.name}" was canceled before starting`))
-          break
-        case core.JobPhase.Failed:
-          reject(new Error(`Job "${this.name}" failed`))
-          break
-        case core.JobPhase.SchedulingFailed:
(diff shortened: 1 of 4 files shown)


vdice (commented):
Should we add test coverage around this new logic?

krancour:
> Should we add test coverage around this new logic?

We should, but it will require mocking out one of the most complex parts of the SDK and my comfort level doing that in TS isn't what it ought to be.

I'm going to give it a shot. If it becomes too much of a time-sink, however, how would you feel about covering this case with an integration test instead?

vdice:
> I'm going to give it a shot. If it becomes too much of a time-sink, however, how would you feel about covering this case with an integration test instead?

Sounds good 👍

krancour:
@vdice I tried to write unit tests for this, but found it impossible without some significant refactoring that might also have undone a lot of the encapsulation that was actually deliberate.

So I fell back on adding an integration test case as we discussed. The new case if visible in commit aba399db93681bfe343f3d00d117a5c4d292c62e

vdice (approved):
LGTM

PR #1777, opened 2022-01-09
cli: test connection to api server before returning to caller

### LINKED ISSUE #1707: `brig term` freezes on refresh when not connected to a server
Steps to reproduce:
- run `brig term` without being connected to a server
- hit `F5` or `r`, and then try to exit with `q`

The terminal freezes until the client completes 5 attempts to connect to the server. I think this situation can be handled a little more elegantly, like not launching term in the first place if a connection to the server is not established.

Unrelated to this, I also think it would be nice to display a message while trying to connect to the server, just so the user knows what is going on

snprajwal:
Closes #1707


krancour:
@snprajwal thanks for the contribution! This looks great.

The only change I would ask for is that the `getClient()` function's signature be modified to take an arg like `testConnection bool`.

Most calls to `getClient()` should pass `false` because testing the connection every time will slow things down quite a bit, but it would be nice to have the _option_ to test the client in advance for cases like what you were running into -- `brig term`, specifically.

krancour:
@snprajwal this is great! Thank you so much!

krancour (approved):

PR #1783, opened 2022-01-18
Support option to create non-existing project on update

### LINKED ISSUE #1776: Possibility to create non-existing project while using the project update command
Hi everyone!

Must say, I just love what you are doing. Keep up with the good work! 🥇 

I have one thing that would be very nice to have, that is a bit annoying. The should be a --install/--create flag
to the update project command in brig. Which creates the project if it does not exist. Very similar to the "helm update --install"-command.

krancour:
Fixes #1776

### DIFF
--- a/v2/apiserver/internal/api/projects.go
+++ b/v2/apiserver/internal/api/projects.go
@@ -95,6 +95,15 @@ func (p ProjectList) Contains(project Project) bool {
 	return false
 }
 
+// ProjectUpdateOptions represents useful, optional settings for updating a
+// Project. It currently has no fields, but exists to preserve the possibility
+// of future expansion without having to change client function signatures.
+type ProjectUpdateOptions struct {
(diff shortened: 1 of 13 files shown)


v2/apiserver/internal/api/projects.go
```diff
@@ -95,6 +95,15 @@ func (p ProjectList) Contains(project Project) bool {
 	return false
 }
 
+// ProjectUpdateOptions represents useful, optional settings for updating a
+// Project. It currently has no fields, but exists to preserve the possibility
```
vdice, line 99: Note to remove "It currently has no fields..."

vdice (approved):

v2/apiserver/internal/api/projects.go
krancour: Fixed in f02fbd3faa48ff69152a01a464105a9745887d22

PR #1791, opened 2022-01-26
flatten the SDK

krancour:
"Vanity packages" are buckets for components that don't logically _belong_ in separate packages, but are only on account of a developer bias / comfort level with a certain arbitrary organization scheme. In Go especially, where package dependency cycles are forbidden, it's easy to accidentally paint oneself into a corner this way.

Way back in #1528, out of necessity, we flattened the API server's domain model into a single package. At that time, we were already in beta and opted _not_ to impose the same package flattening on the Go SDK since it would be a breaking change. The result has been that since #1528, the SDK and API package structures are quite different.

#1787 has since introduced breaking changes that will result in a major release of the Go SDK (3.0; which will still "speak" API v2). With the next release of the Go SDK being a major one, now is a sensible time to flatten the SDK packages to match the API server's package structure. The end result should be an SDK that's much easier to consume because important components aren't spread out over many packages with increased possibility of dependency cycles as the SDK continues to evolve.

While I'm sorry for the size of this PR, the choices are basically to do this now or hold it for the _next_ major release of the SDK. I'd rather get it over and done with.

This PR is divided into five commits. One flattens the SDK. The others each make corresponding changes to SDK consumers, namely the scheduler, observer, git-initializer, and CLI.

This PR _only_ moves and renames things. There are no functional changes in this PR.

__Edit:__ On account of this PR's size, there is not a huge rush.


sdk/v3/role_assignments.go
```diff
@@ -1,34 +1,66 @@
...
+		}{
+			TypeMeta: meta.TypeMeta{
+				APIVersion: meta.APIVersion,
+				Kind:       RoleAssignmentKind,
+			},
+			Alias: (Alias)(r),
+		},
+	)
+}
+
```
krancour, line 57: Despite appearances in the diff, this is not new code. When packages were combined, code from two different `roel_assignments.go` go files got merged into one file. This happens in other places as well.

There's nothing net new in this whole PR.

vdice (approved):

PR #1796, opened 2022-01-31
implement "whoami" endpoint

### LINKED ISSUE #1795: enhancement: add "me" or "whoami" endpoint
There's currently no way for a client to get information about the authenticated user since the authentication process returns an opaque token and nothing else.

There are cases, however, where knowing who the user is would be essential. In the prototype for the new Kashti, for instance, we could use existing APIs to query for user permissions... but only if we know who the user is.

The new endpoint could simply return an instance of the existing principal reference type.

krancour:
Fixes #1795.

This is needed to continue work on the prototype v2-compatible Kashti.


v2/apiserver/internal/api/rest/token_auth_filter.go
```diff
@@ -15,6 +15,15 @@ import (
...
 
+var (
+	// Root is a singleton that represents Brigade's "root" user.
+	root = &api.RootPrincipal{}
+	// Scheduler is a singleton that represents Brigade's scheduler component.
+	scheduler = &api.SchedulerPrincipal{}
+	// Observer is a singleton that represents Brigade's observer component.
+	observer = &api.ObserverPrincipal{}
+)
+
```
krancour, line 26: These are not new. They just moved here from `v2/apiserver/internal/api/rest/named_principals.go` when other elements of that file were relocated to the `api` package.

sdk/v3/authn_client.go
```diff
@@ -1,12 +1,19 @@
...
+	rm "github.com/brigadecore/brigade/sdk/v3/internal/restmachinery"
 	"github.com/brigadecore/brigade/sdk/v3/restmachinery"
 )
 
 // AuthnClient is the root of a tree of more specialized API clients for dealing
 // with identity and authentication.
 type AuthnClient interface {
+	// WhoAmI returns a PrincipalReference for the currently authenticated
+	// principal.
+	WhoAmI(context.Context) (PrincipalReference, error)
```
krancour, line 16: The authn client can return more specialized clients for users and service accounts, but I figured whoami was better to include right here. If a caller doesn't know who they are, they may also not know if they're a service account or a user.

vdice (approved):
Know thyself!  🍎

PR #1798, opened 2022-02-02
docker login for all image builds in ci

krancour:
This is a step toward addressing DockerHub rate limiting. Our most frequent pulls are due to the `FROM` directive in Dockerfiles that we build in our jobs, so logging in before image builds, the way we already do before image pushes is one of the best measures we can take against bumping into the rate limit.

We may still need to do more, but this should be a significant improvement.


vdice (approved):
LGTM here and same for all associated repos with same changeset.

PR #1800, opened 2022-02-02
upgrade to latest go-tools image

krancour:


vdice (approved):
LGTM here and same for all associated repos with similar changeset.

PR #1803, opened 2022-02-04
update go.mod file to use go 1.17 conventions

krancour:


v2/go.mod
```diff
@@ -1,34 +1,27 @@
 module github.com/brigadecore/brigade/v2
 
-go 1.15
+go 1.17
```
krancour, line 3: This is the only thing that was manually changed. Every other change in this PR is the result of running `go mod tidy`.

vdice (approved):
LGTM pending CI passing.

PR #1806, opened 2022-02-07
New project role assignments endpoint

### LINKED ISSUE #1804: proposal: add new/improved project role assignments endpoint
The existing endpoint for listing project-level role assignments is at path `v2/projects/{:project_id}/role-assignments`.

Requiring a project ID to be specified makes it impossible to query for all of a given principal's project role assignments across all projects, which is a useful thing to do if one is trying to audit or otherwise develop a comprehensive notion of _all_ of a given principal's permissions.

I propose adding a second endpoint at path `v2/project-role-assignments` that takes all optional selection criteria, including project ID as URL query parameters. For compatibility with older clients, the old endpoint would continue to exist and function just as it does today, but newer clients could opt to use the newer endpoint to support queries that were not previously possible.

krancour:
Fixes #1804 

Builds on #1805 and should be rebased after that is merged.

### DIFF
--- a/v2/apiserver/internal/api/rest/project_role_assignments_endpoints.go
+++ b/v2/apiserver/internal/api/rest/project_role_assignments_endpoints.go
@@ -74,11 +74,11 @@ func (p *ProjectRoleAssignmentsEndpoints) list(
 	principalType := req.URL.Query().Get("principalType")
 	principalID := req.URL.Query().Get("principalID")
 
-	// This will yield a project ID if we got her via the original
+	// This will yield a project ID if we got here via the original
 	// /v2/projects/{id}/role-assignments endpoint.
 	projectID := mux.Vars(req)["id"]
 	// If we cannot pick a project ID out of the path, try to get it from a query
-	// parameters.
+	// parameter.
 	if projectID == "" {
 		projectID = req.URL.Query().Get("project")
 	}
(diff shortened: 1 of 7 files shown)


v2/apiserver/internal/api/rest/project_role_assignments_endpoints.go
```diff
@@ -67,8 +73,20 @@ func (p *ProjectRoleAssignmentsEndpoints) list(
 ) {
 	principalType := req.URL.Query().Get("principalType")
 	principalID := req.URL.Query().Get("principalID")
+
+	// This will yield a project ID if we got her via the original
```
vdice, line 77:
```suggestion
	// This will yield a project ID if we got here via the original
```

v2/apiserver/internal/api/rest/project_role_assignments_endpoints.go
```diff
@@ -67,8 +73,20 @@ func (p *ProjectRoleAssignmentsEndpoints) list(
...
+	// /v2/projects/{id}/role-assignments endpoint.
+	projectID := mux.Vars(req)["id"]
+	// If we cannot pick a project ID out of the path, try to get it from a query
```
vdice, line 80:
```suggestion
	// If we cannot pick a project ID out of the path, try to get it from the query
```

vdice (approved):

PR #1808, opened 2022-02-09
upgrade to go-tools v0.8.0 -- includes go 1.18

krancour:
This PR upgrades us to Go 1.18 across the board. Even though the SDK is build with Go 1.18, we're not using any Go 1.18-specific features because we don't want to force SDK users to upgrade to Go 1.18 yet if they're not ready.

We'll start taking advantage of generics (server side) in a follow-up to avoid this simple PR becoming unnecessarily bloated.

### DIFF
--- a/.brigade/brigade.ts
+++ b/.brigade/brigade.ts
@@ -1,7 +1,7 @@
 import { events, Event, Job, ConcurrentGroup, SerialGroup, Container } from "@brigadecore/brigadier"
 
 const azImg = "mcr.microsoft.com/azure-cli"
-const goImg = "brigadecore/go-tools:v0.7.0"
+const goImg = "brigadecore/go-tools:v0.8.0"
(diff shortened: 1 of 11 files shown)


v2/apiserver/internal/api/mongodb/projects_store.go
```diff
@@ -205,16 +205,16 @@ function() {
...
+			"$elemMatch": preliminaryMatchCriteria,
 		},
-		findOptions,
-	)
+		"$where": where,
+	}
+	if event.ProjectID != "" {
+		query["id"] = event.ProjectID
+	}
+	cur, err := p.collection.Find(ctx, query, findOptions)
```
krancour, line 217: This improvement, _besides_ being an overall improvement, eliminates the need to implement a `Contains()` member function on the generic `meta.List` type.

vdice (approved):

PR #1810, opened 2022-02-10
improve speed and reliability of integration tests

### LINKED ISSUE #1809: Improve speed of integration tests
Our integration tests take a _long_ time to run during CI. This is largely (but not exclusively) on account of the fact that they rebuild all images, one at a time, and load them into a KinD cluster before actually running the tests.

If we push candidate images to a registry as they are built, often concurrently, by preceding steps in the CI process, those images will not need to be rebuilt before the integration tests.

We wouldn't want to use a public registry for this purpose because we'd clutter it with unstable images. We could, however, deploy a private registry within our cluster and use it for this purpose.

My best guess is this will cut the time to execute CI tests by at least 50%.

krancour:
Fixes #1809

The premise behind this improvement is simple -- that during CI, it's faster to push images we've already built to a registry and pull them later for the integration tests than it is to _rebuild_ them later for the integration tests. What I do want to avoid, however, is publishing potentially unstable images that have not yet passed the full battery of tests to the same registries/repositories where we push stable images.

This PR modifies the `brigade.ts` script to select an "unstable" repository when pushing candidate images that still need to be integration tested and a "stable" repository when pushing the images associated with an official release.

The speed increase here also makes integration tests far less likely to time out.

I've already stood up an ACR instance to use for unstable images and relevant secrets have already been added. We can blow away all repositories in that registry periodically. We may even be able to automate that in a follow-up PR.

Other brigadecore repos do _not_ need similar changes because this repo's complex integration tests are unique.


.brigade/brigade.ts
```diff
@@ -337,30 +338,11 @@ events.on("brigade.sh/github", "ci:pipeline_requested", async event => {
...
-      pushArtemisJob(event),
-      pushAPIServerJob(event),
-      pushGitInitializerJob(event),
-      pushLoggerLinuxJob(event),
-      pushObserverJob(event),
-      pushSchedulerJob(event),
-      pushWorkerJob(event),
-      publishBrigadierDocsJob(event),
-    ).run()
-  }
```
krancour, line 363: We used to publish "edge" images on merge to master. This has really been making a big mess of our public Docker Hub repos, so we stop doing it here.

jorgearteiro (approved):
LGTM

Issue #1821, opened 2022-02-16
improvement: smoother startup


krancour:
Brigade components do a lot of "flapping" after helm install/upgrade.

The API server depends on both MongoDB and Artemis being up and running. The observer and scheduler depend on the API server being up and running. Components whose network-bound dependencies aren't ready yet still try to start, they fail, and crash loop backoffs occur. If the backoffs progress to lengthy enough intervals between retries, they can _really_ slow down the total time that the install/upgrade takes.

I want to propose that a better approach to all this "free" eventual consistency (courtesy of k8s) may be to use the `retries` package from [brigadecore/brigade-foundations](https://github.com/brigadecore/brigade-foundations) to take control of connection retries ourselves _without_ crashing. The result could be components that are slower to start when their dependencies aren't satisfied yet, but _will_ eventually start without having to suffer crashes in the interim.

krancour:
More info:

The error message typically seen when the API server crashes on startup is:

```
error adding indexes to events collection: context deadline exceeded
```

This is because, per documentation, the `mongo.Connect(...)` call made at startup doesn't actually perform any I/O.

Attempts to ensure the indices on various collections are the _first_ attempts at any database I/O.

If we test the connection immediately after obtaining it and put that login inside a retry loop, we'll be in a much better position.

PR #1826, opened 2022-02-17
fix bug in message writing abstaction

krancour:
I've discovered the source of a known bug while investigating #1821.

This AMQP client connect/re-connect logic is managed by `retries.ManageRetries(...)`. The retry logic works well when refreshing a stale connection, however, if there is no connection and it attempts to dial the message server for the first time, and fails, the AMQP client is no longer nil, but _also_ is not fully initialized. On the _next_ retry, the call to `w.amqpClient.Close()` will trigger a nil pointer dereference. 😦 

We've _frequently_ seen panics related to this in API server startup logs.

The reason this hasn't been a bigger problem than it has been to date is that Artemis usually comes up faster than MongoDB. The API server ends up crashing and restarting itself several times before MongoDB is up and running, and by the time it is, Artemis is usually up and running and the very first attempt to dial the messaging server succeeds without a hitch.

Once #1821 is addressed, DB unavailability won't trigger immediate API server restarts, which means, when combined with the above, that we'll _never_ get a good connection to the message server.

This PR fixes the nil pointer dereference so the retry logic for dialing the message server works as intended.

### DIFF
--- a/v2/apiserver/internal/lib/queue/amqp/writer.go
+++ b/v2/apiserver/internal/lib/queue/amqp/writer.go
@@ -63,10 +63,11 @@ func (w *writerFactory) connect() error {
 			if w.amqpClient != nil {
 				w.amqpClient.Close()
 			}
-			var err error
-			if w.amqpClient, err = myamqp.Dial(w.address, w.dialOpts...); err != nil {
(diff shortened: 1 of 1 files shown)


vdice (approved):
Nice find!

PR #1827, opened 2022-02-17
Smoother startup

### LINKED ISSUE #1821: improvement: smoother startup

krancour:
Fixes #1821 

On a brand new KinD cluster (image cache not primed) on my MBP, `helm install` now succeeds in about 1 minute with no services restarting in the process.

__Edit:__ This is such a significant improvement that I'm going to plan on a v2.3.1 release to put this in people's hands as soon as possible.

### DIFF
--- a/v2/apiserver/main.go
+++ b/v2/apiserver/main.go
@@ -2,8 +2,11 @@ package main
 
 // nolint: lll
 import (
+	"context"
 	"log"
+	"time"
 
+	"github.com/brigadecore/brigade-foundations/retries"
 	"github.com/brigadecore/brigade-foundations/signals"
 	"github.com/brigadecore/brigade-foundations/version"
 	"github.com/brigadecore/brigade/v2/apiserver/internal/api"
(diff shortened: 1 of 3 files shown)


vdice (approved):
🎉 Glad to hear the install time has been reduced substantially!  LGTM.

krancour:
> Glad to hear the install time has been reduced substantially!

@vdice and that, in-turn further stabilizes the integration tests, too!

PR #1832, opened 2022-02-22
docs: portuguese language support

### LINKED ISSUE #1828: docs in other languages?
@jorgearteiro has done a great job in generating interest in the project recently, especially in Brazil. As a result, we have native Portuguese speakers who would appreciate Portuguese language docs _and_ we have native Portuguese speakers willing to translate the existing docs.

This raises some questions about where to host and how to manage internationalized documentation.

@flynnduism and @vdice looking for some insight from you if you have any.

One consideration that has crossed my mind is that it might be best to house docs for other languages in separate repos and publish them to separate subdomains. I _think_ this would make sense because the maintainers for said repos could be fluent speakers of the language in question and it would eliminate any expectations or assumptions about core maintainers updating all translations every time we touch the original English language docs (which we're not qualified to do!)

Thoughts?

krancour:
Fixes #1828 

This PR internationalizes the docs through modest changes to site configuration and layout and localizes for Portuguese by sourcing a translation from https://github.com/brigadecore/docs-portuguese as a submodule.


krancour:
https://deploy-preview-1832--brigade-docs.netlify.app/

Now there are _two_ homepages in the Portuguese docs, but I think the problem is on my end. Let me see if I can fix it.

PR #1835, opened 2022-02-22
docs: remove unused images

krancour:
These images do not seem to be used. The ones in `docs/static/img` are.


krancour:
Correction: Some of these are used and some are not. Fix incoming...

PR #1841, opened 2022-02-23
docs: menu fixes

### LINKED ISSUE #1385: On smaller viewports, sidebar overlaps with main content
On the https://docs.brigade.sh/ site, when the viewport is < ~1200px, the fixed sidebar overlaps with the main content view. Here is a screenshot when the viewport is 700px:
image

I'm able to reproduce this on Chrome (v90.x) and Safari.

One quick fix would be to wrap the `.sidebar.fixed nav` class with a media query 

```
@media only screen and (min-width: 1200px) {
  .sidebar.fixed nav {
      position: fixed;
      max-width: 300px;
      left: 0;
      top: 0;
      bottom: 0;
      z-index: 1500;
      overflow-y: scroll;
   }
}
```

I'm not sure if brigade docs use some template or theme so feel free to close if this is not the right place for this issue.

krancour:
Way back in #865, it seems the sidebar was customized to be driven by data in `config.toml` instead of by front matter from content. As I understand it, this was done, at least in part, in order to take control of the order of items in the menu, which until then were in lexical order by title. (Although I think there may have been a simpler resolution available than a custom menu. There _is_ a `ordersectionsby = "weight"` option that could have been specified in `config.toml`.)

Per the `docs/readme.md` our workaround left us in a less than optimal place:

> Currently, the weights configured in the `config.toml` file specify ordering of each section in the main sidebar menu.
>
> Weight values on individual documents are meant to specify ordering of pagination (previous and next navigation) within each doc's section.
>
> This is fairly unwieldy, so if any docs/Hugo pros have knowledge on how to simplify, we'd love the contribution!

In the course of investigating options for internationalizing our docs (see #1828), I've come to believe that a sidebar menu that is driven by data in `config.toml` instead of by front matter from content is going to be an impediment to internationalization, so I set out to correct this duplication of weights.

This led me to the following:

1. Undo menu customizations from #865
1. Remove index from `config.toml`
1. Ensure sidebar is sorted by weight using `ordersectionsby = "weight"` option in `config.toml`

This left me in a place where the sidebar looked exactly as it should _except_ that in some places the name of the link in the sidebar menu used to be _overridden_ by data that has since been removed from `config.toml`.

Some research revealed the existence of the [predefined `linkTitle` front matter variable](https://gohugo.io/content-management/front-matter/#predefined). So I:

1. Added `linkTitle` to front matter in our content, as needed.
1. Minimally customized the sidebar menu to use `{{ .LinkTitle }}` where it previously used `{{ .Title }}`.
    1. I also discovered that overrides / customizations to Hugo theme are meant to be applied by making modifications in one's _own_ `layout/` folder rather than directly overwriting files in the _theme's_ `layout/` folder, so I used this approach. See [here](https://tangenttechnologies.ca/blog/hugo-themes/) for reference.

Where this leaves us is:

* Menus now are identical to before this PR, but...
* We no longer need the duplicate index/weights in `config.toml`, clearing the path to internationalization.
* Our sidebar overrides are a bit cleaner.


vdice (approved):
Great simplification! LGTM.

PR #1843, opened 2022-02-23
docs: update to latest hugo-theme-techdoc

### LINKED ISSUE #1385: On smaller viewports, sidebar overlaps with main content

krancour:
Fixes #1385 

This rips out our Frankenstein customizations to the Hugo Techdoc theme and replaces it with the latest version using a submodule. This is apparently how themes are meant to be used.

I've updated our customizations to work with recent changes to the theme. #1385 is fixed in the process. Those customizations are all applied _outside_ the `themes/` folder as is recommended.


krancour:
Seems there's some missing CSS:

Screen Shot 2022-02-23 at 3 29 46 PM

krancour:
This issue of poor color contrast in code blocks only appears with newer versions of Hugo:

Screen Shot 2022-02-23 at 4 08 09 PM

docs/layouts/partials/head.html
```diff
@@ -0,0 +1,41 @@
...
+<script src="https://cdn.jsdelivr.net/npm/jquery.easing@1.4.1/jquery.easing.min.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/clipboard@2.0.6/dist/clipboard.min.js"></script>
+<script src="{{ "js/bundle.js" | absURL }}"></script>
+{{- partial "meta/google-analytics-async.html" . -}}
+{{- partial "meta/tag-manager.html" . -}}
+{{- partial "meta/google-site-verification.html" . -}}
+{{- partial "custom-css.html" . -}}
+{{- template "_internal/opengraph.html" . -}}
+{{- template "_internal/twitter_cards.html" . -}}
+{{- template "_internal/schema.html" . -}}
```
krancour, line 41: This file is _copied_ from the theme and minimally modified to include include our favicon and custom stylesheet. Hugo favors this file over the original in `themes/...`

docs/layouts/partials/menu/open-menu.html
```diff
@@ -1,30 +1,30 @@
 {{$currentNode := .}}
```
krancour, line 1: This file is _copied_ from the theme and minimally modified to favor using each page's `titleLink` over its `title` when generating the menu. Hugo favors this file over the original in themes/...

docs/layouts/partials/meta/google-analytics-async.html
```diff
@@ -94,3 +68,6 @@
     });
   };
 </script>
+<!-- END: Google Analytics -->
+{{- end -}}
+{{- end -}}
```
krancour, line 73: This file is _copied_ from the theme and minimally modified to require users consent to the use of cookies before Google Analytics is activated. Hugo favors this file over the original in `themes/...`

This is consistent with the behavior prior to this PR. The only difference is that the hack is now part of our site code rather than a modification to the theme.

docs/layouts/partials/prepend-body.html
```diff
@@ -0,0 +1 @@
+<!-- This is deliberately empty to override something we don't want from the theme. -->
```
krancour, line 1: This overrides the file of the same name found in `themes/...`

The original uses a `<noscript>` element to activate Google Analytics in an invisible <iframe> when JS is disabled in the user's browser.

We don't want that because we don't want to activate Google Analytics unless the user explicitly consents to the use of tracking cookies, and we cannot gather that consent without JS enabled, so we simply have to live without GA in this case.

vdice (approved):
Rather than going through all of the changed files in detail, I checked out the deploy preview and all looks great to me.

It sounds like this changeset gets us up-to-date and the internal doc workings vastly cleaned up, which all sounds excellent.

Noticed some notes around Google Analytics -- I'll defer to @krancour to make sure all is working as expected.

Otherwise, LGTM from me!

krancour:
> Noticed some notes around Google Analytics -- I'll defer to @krancour to make sure all is working as expected.

@vdice it is for sure working identically to before. Using the browser's developer tools, I was able to compare before/after and verify that Google Analytics is now incorporated into the site only after a user explicitly consents to cookies -- same as before. The only thing that's up in the air is that I'm missing access to view the collected data, but we'll work that out.

PR #1866, opened 2022-03-10
upgrade artemis

krancour:


krancour:
I'm combining this with #1868

PR #1867, opened 2022-03-10
switch worker to alpine

krancour:
Upgrades us to Node 1.16.14 everywhere we use it (for instance, running tests, linter, etc.) and specifically swaps the worker image to an alpine-based image. The alpine base image has significantly better results when scanned for known OS-level vulnerabilities.


krancour:
Closing in favor of #1869

PR #1869, opened 2022-03-10
switch to alpine-based images for artemis and worker

krancour:


Makefile
```diff
@@ -267,14 +267,13 @@ build-git-initializer-windows:
 
 .PHONY: build-%
 build-%:
-	docker login $(DOCKER_REGISTRY) -u $(DOCKER_USERNAME) -p $${DOCKER_PASSWORD}
```
krancour, line 270: This was removed accidentally and should be put back.

Makefile
```diff
@@ -277,6 +276,28 @@ build-%:
...
+.PHONY: build-worker
+build-worker:
+	docker buildx build \
+		-f v2/worker/Dockerfile \
+		-t $(DOCKER_IMAGE_PREFIX)worker:$(IMMUTABLE_DOCKER_TAG) \
+		-t $(DOCKER_IMAGE_PREFIX)worker:$(MUTABLE_DOCKER_TAG) \
+		--build-arg VERSION=$(VERSION) \
+		--build-arg COMMIT=$(GIT_VERSION) \
+		--platform linux/amd64,linux/arm64v8 \
+		.
```
krancour, line 299: These need different values for the `--platform` flag than other images do. The reason is that multiarch images don't exist for Alpine that can handle both `linux/amd64` and `linux/arm64`. The Dockerfiles in these cases select _different_ base images depending on the target architecture. They select either `amd64/___:___` or `arm64v8/___:___`. `arm64/___:___` images do not exist, so that's what forces us to use different values for the `--platform` flag.

`arm64` and `arm64v8` are the _same_ architecture afaik, and are in fact two different ways of saying what is _officially_ `aarch64`.

None of this is about an architectural mismatch. It's simply about strings. :sigh:

Makefile
```diff
@@ -343,6 +364,30 @@ push-%:
...
+	docker buildx build \
+		-f v2/worker/Dockerfile \
+		-t $(DOCKER_IMAGE_PREFIX)worker:$(IMMUTABLE_DOCKER_TAG) \
+		-t $(DOCKER_IMAGE_PREFIX)worker:$(MUTABLE_DOCKER_TAG) \
+		--build-arg VERSION=$(VERSION) \
+		--build-arg COMMIT=$(GIT_VERSION) \
+		--platform linux/amd64,linux/arm64v8 \
+		--push \
+		.
+
```
krancour, line 390: See [explanation](https://github.com/brigadecore/brigade/pull/1869/files#r824111189)

.brigade/brigade.ts
```diff
@@ -2,7 +2,7 @@ import { events, Event, Job, ConcurrentGroup, SerialGroup, Container } from "@br
 
 const azImg = "mcr.microsoft.com/azure-cli"
 const goImg = "brigadecore/go-tools:v0.6.0"
-const jsImg = "node:16.11.0-bullseye"
+const jsImg = "node:16.14.0-bullseye"
```
krancour, line 5: Bumping to 16.4.0, but continuing to use Debian Bullseye for things like running linters and unit tests. This PR is really primarily concerned with changing the base images of images we _distribute_ as part of each Brigade release.

Makefile
```diff
@@ -267,7 +267,6 @@ build-git-initializer-windows:
 
 .PHONY: build-%
 build-%:
-	docker login $(DOCKER_REGISTRY) -u $(DOCKER_USERNAME) -p $${DOCKER_PASSWORD}
```
krancour, line 270: This line didn't belong to begin with.

v2/artemis/Dockerfile
```diff
@@ -1,6 +1,6 @@
-FROM openjdk:8-jre-bullseye
+FROM $TARGETARCH/alpine:3.15.0
```
krancour, line 1: Picks `amd64/alpine:3.15.0` or `arm64v8/alpine:3.15.0`.

v2/artemis/Dockerfile
```diff
@@ -1,6 +1,6 @@
...
 
-ENV ARTEMIS_VERSION=2.19.0
+ARG ARTEMIS_VERSION=2.20.0
```
krancour, line 3: Bumping

v2/artemis/Dockerfile
```diff
@@ -10,20 +10,17 @@ WORKDIR /opt/artemis
...
-  apt-get install -y --no-install-recommends \
-    ca-certificates \
+RUN apk update && \
+  apk add \
     curl \
-    libaio1 \
-    xmlstarlet &&\ 
-  rm -rf /var/lib/apt/lists/* && \
+    openjdk11-jre \
+    xmlstarlet && \
```
krancour, line 17: Switching from Debian to Alpine means package manager changes from `apt` to `apk`.

v2/artemis/Dockerfile
```diff
@@ -10,20 +10,17 @@ WORKDIR /opt/artemis
...
   curl https://dlcdn.apache.org/activemq/activemq-artemis/${ARTEMIS_VERSION}/apache-artemis-${ARTEMIS_VERSION}-bin.tar.gz \
     -o apache-artemis-${ARTEMIS_VERSION}-bin.tar.gz && \
   tar xzf apache-artemis-${ARTEMIS_VERSION}-bin.tar.gz --strip 1 && \
-  rm apache-artemis-${ARTEMIS_VERSION}-bin.tar.gz && \
   mkdir /var/lib/artemis && \
-  groupadd -g 1000 -r artemis && \
-  useradd -r -u 1000 -g artemis artemis && \
+  addgroup -S -g 1000 artemis && \
+  adduser -S -u 1000 artemis -G artemis && \
```
krancour, line 23: Debian and Alpine use different implementations of these commands.

v2/worker/Dockerfile
```diff
@@ -1,33 +1,44 @@
-FROM node:16.11.0-bullseye-slim
-ARG VERSION
-ENV WORKER_VERSION=${VERSION}
+FROM --platform=$BUILDPLATFORM debian:bullseye-slim
```
krancour, line 1: We use Debian for the first few steps because the inline scripts we run only work in bash. Alpine doesn't have bash. I tried hard to make the scripts POSIX-compliant so they'd work in any shell, but I just couldn't nail it.

krancour: Also, this stage always runs on the native architecture of the build server.

v2/worker/Dockerfile
```diff
@@ -1,33 +1,44 @@
...
-RUN bash -c 'if [[ $WORKER_VERSION =~ ^v[0-9]+(\.[0-9]+)*(\-.+)?$ ]]; then \
-      sed -i s/0.0.1-placeholder/$(echo $WORKER_VERSION | cut -c 2-)/ package.json ; \
-    fi ; \
-    yarn install --prod && yarn build \
+RUN bash -c 'if [[ $VERSION =~ ^v[0-9]+(\.[0-9]+)*(\-.+)?$ ]]; then \
+      sed -i s/0.0.1-placeholder/$(echo $VERSION | cut -c 2-)/ package.json ; \
+    fi \
   '
 
+FROM $TARGETARCH/node:16.14.0-alpine3.15
```
krancour, line 26: Picks `amd64/node:16.14.0-alpine3.15` or `arm64v8/node:16.14.0-alpine3.15`.

krancour:
Closing in favor of #1871

PR #1878, opened 2022-03-19
Update quickstart.md with root login disclaimer

### LINKED ISSUE #1544: Docs: Link from Quickstart to the Authorization docs
Just as a matter of course, when I was reading through the quickstart and logged in using `--root`, I wondered how user accounts work. Adding a link to something like this from the [Log Into Brigade](https://v2--brigade-docs.netlify.app/intro/quickstart/#log-in-to-brigade) link would be good.
https://docs.brigade.sh/topics/administrators/authorization/

AnuragThePathak:

Fixes #1544 

This PR adds a disclaimer regarding the use of root login in a production cluster on the Brigade Quickstart page.

**If applicable**:

### DIFF
--- a/docs/content/intro/quickstart.md
+++ b/docs/content/intro/quickstart.md
@@ -150,6 +150,8 @@ $ kubectl --namespace brigade port-forward service/brigade-apiserver 8443:443 &>
 
 ### Log into Brigade
 
+> ⚠️ In this section, we'll be logging in as the root user. This option should typically be disabled in a production-grade Brigade deployment. Read more about user authentication [here](https://docs.brigade.sh/topics/administrators/authentication/).
+
 To authenticate to Brigade as the root user, you first need to acquire the
 auto-generated root user password:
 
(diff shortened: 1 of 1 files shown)


krancour (approved):
LGTM after CI runs.

krancour:
@AnuragThePathak that check is added by #1877, which is not merged yet. It was prematurely marked as a required check in our branch protections. The fact that didn't run can be ignored right now.

docs/content/intro/quickstart.md
```diff
@@ -150,6 +150,8 @@ $ kubectl --namespace brigade port-forward service/brigade-apiserver 8443:443 &>
 
 ### Log into Brigade
 
+> ⚠️ In this section, we'll be logging in as the root user. This option should typically be disabled in a production-grade Brigade deployment. Read more about user authentication [here](https://docs.brigade.sh/topics/administrators/authentication/).
```
krancour, line 153: I don't think we should be using an absolute URL here. If the site is deployed to another domain (this happens, for instance, on the Netlify deployment previews) this will cause problems.

AnuragThePathak: Oh I didn't notice that earlier.

krancour:
Thanks for this @AnuragThePathak !

PR #1881, opened 2022-03-21
add generic List type to API server and use throughout

krancour:
Follow up to #1808 

Starts to make use of generics.

This PR is decomposed into the following commits:

* Add generic list type
* Remove non-generic list types
* Transition all API server code from non-generic list types to the generic list type
* Resolving conflicts that arose after #1903 added new units tests

There are no functional changes in this PR.

Additionally, I've proactively commented in-line on the very, very few changes wherein the reason for them might not be immediately evident without explanation.

### DIFF
--- a/v2/apiserver/internal/api/events_test.go
+++ b/v2/apiserver/internal/api/events_test.go
@@ -14,10 +14,6 @@ func TestEventMarshalJSON(t *testing.T) {
 	metaTesting.RequireAPIVersionAndType(t, &Event{}, EventKind)
 }
 
-func TestEventListMarshalJSON(t *testing.T) {
-	metaTesting.RequireAPIVersionAndType(t, &EventList{}, "EventList")
-}
-
 func TestCancelManyEventsResultMarshalJSON(t *testing.T) {
 	metaTesting.RequireAPIVersionAndType(
 		t,
--- a/v2/apiserver/internal/api/events_test.go
+++ b/v2/apiserver/internal/api/events_test.go
@@ -55,39 +55,16 @@ func TestEventsServiceCreate(t *testing.T) {
 		name       string
 		event      Event
 		service    EventsService
-		assertions func(EventList, error)
+		assertions func(meta.List[Event], error)
 	}{
 		{
 			name: "unauthorized",
 			service: &eventsService{
 				authorize: neverAuthorize,
 			},
-			assertions: func(_ EventList, err error) {
+			assertions: func(_ meta.List[Event], err error) {
 				require.Error(t, err)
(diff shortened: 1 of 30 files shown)


v2/apiserver/internal/api/mongodb/projects_store.go
```diff
@@ -205,16 +205,16 @@ function() {
...
+			"$elemMatch": preliminaryMatchCriteria,
 		},
-		findOptions,
-	)
+		"$where": where,
+	}
+	if event.ProjectID != "" {
+		query["id"] = event.ProjectID
+	}
+	cur, err := p.collection.Find(ctx, query, findOptions)
```
krancour, line 217: Proactively addressing the reason for this change.

`ProjectList` (which is gone in favor of `meta.List[Project]`) used to have a `Contains()` function and if an event identified a specific project, what we used to do was find _all_ the subscribed projects and then check the service would call `projects.Contains(event.projectID)` to see if it was among them.

`meta.List` has no `Contains()` functions, but it hardly matters because what we were doing in the first place was really inefficient.

This change right here just adds the project ID to the query criteria if the event had specified one.

v2/apiserver/internal/api/events_test.go
```diff
@@ -59,41 +55,18 @@ func TestEventsServiceCreate(t *testing.T) {
...
-						return ProjectList{}, nil
-					},
-				},
-			},
-			assertions: func(_ EventList, err error) {
-				require.Error(t, err)
-				require.Contains(t, err.Error(), "error retrieving project")
-				require.Contains(t, err.Error(), "projects store error")
-			},
-		},
```
krancour, line 96: This test case goes away after the change described [here](https://github.com/brigadecore/brigade/pull/1881/files#r841099739) because there's no longer a separate step for retrieving a specific project by ID.

v2/apiserver/internal/api/events_test.go
```diff
@@ -102,22 +75,12 @@ func TestEventsServiceCreate(t *testing.T) {
 			service: &eventsService{
 				projectAuthorize: alwaysProjectAuthorize,
 				projectsStore: &mockProjectsStore{
-					GetFn: func(context.Context, string) (Project, error) {
-						return Project{
-							ObjectMeta: meta.ObjectMeta{
-								ID: "blue-book",
-							},
-						}, nil
-					},
```
krancour, line 111: This goes away after the change described [here](https://github.com/brigadecore/brigade/pull/1881/files#r841099739) because there's no longer a separate step for retrieving a specific project by ID.

vdice (approved):

PR #1891, opened 2022-03-29
upgrade artemis to 2.19.1

krancour:
Looks like the URL for Artemis 2.19.0 is no longer valid. It's returning a 404 when we try to build the Artemis image.

Browsing [here](https://dlcdn.apache.org/activemq/activemq-artemis/), you can see 2.19.0 is no longer in the index.

This PR bumps Artemis to 2.19.1 to move us past this.


krancour:
It think it's reasonable to label this PR as a chore and accept @jorgearteiro's review.

PR #1892, opened 2022-03-29
fix nightly cleanup job

### LINKED ISSUE #1831: automation: nightly cleanup job can fail
The nightly cleanup jobs that cleans out our unstable image registry fails if any of the repositories it deletes do not currently exist. This could happen after a slow day.

This isn't a huge deal, but it can be improved.

mariuskimmina:
Signed-off-by: Marius Kimmina <mar.kimmina@gmail.com>

Closes #1831
This PR removes the hardcoded list of repositories from the nightly cleanup and instead fetches the list of existing repos from azure.

I have tested the `nightly-cleanup.sh` on my own azure account but I have not yet setup a local brigade cluster to test calling the script from `brigade.ts`


scripts/nightly-cleanup.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+az login --service-principal --username $AZ_USERNAME --password $AZ_PASSWORD --tenant $AZ_TENANT
+repo_list=$(az acr repository list --name unstablebrigade)
+
+# remove leading and trailing []
+repo_list=${repo_list::-1}
+clean_repo_list=${repo_list:1}
```
krancour, line 8: `repo_list=$(az acr repository list --name unstablebrigade -o table | tail -n +3)`

Would give you a clean list right out of the gate with no further need to massage the results before/during iterating the list.

krancour:
Hey @mariuskimmina this looks great. I left notes about two small optimizations.

mariuskimmina:
I wasn't aware of the `-o table` option, that's of course better, this way I also didn't have to trim any quotation marks.

scripts/nightly-cleanup.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
```
krancour, line 2: Can we add `set -eou pipefail` here? (But _not_ `x`, as that would bleed secrets into the job logs!)

After that, this PR LGTM.

mariuskimmina: Good call, added it

krancour (approved):

PR #1895, opened 2022-03-30
ci: increase timeout on worker build job

krancour:
Frustratingly the multiarch worker image sometimes takes quite some time to build. I believe the arm64 emulation involved in the build process is mostly to blame.

This PR increases the timeout on the job that builds the worker.

In the future, I may consider doing all the dependency resolution only with the native architecture and only going multiarch in the final stage.

For now, this will unblock a 2.4.1 RC.


krancour:
Nevermind... I am going to use the other approach I described.

PR #1900, opened 2022-03-31
Fix setting secrets on a non-existing project tells you're unauthorized

### LINKED ISSUE #1893: bug: setting secrets on a non-existing project tells you're unauthorized
This is only true if you're logged in as a real user and not as root.

The heart of the issue is the same as for #1750 and resolution should be the same as it was for #1884

cc @AnuragThePathak since you are already familiar with the fix, thanks to your great work on #1884, do you think you'd be willing to fix it here as well?

AnuragThePathak:
Fixes #1893 

While setting a secret, instead of directly checking for PROJECT_ADMIN role we'll now be checking if the user is a READER followed by checking if the project exists so that we can avoid getting unauthorized error when the project doesn't exist.

**If applicable**:

### DIFF
--- a/v2/apiserver/internal/api/secrets_test.go
+++ b/v2/apiserver/internal/api/secrets_test.go
@@ -197,7 +197,7 @@ func TestSecretsServiceSet(t *testing.T) {
 		{
 			name: "unauthorized",
 			service: &secretsService{
-				projectAuthorize: neverProjectAuthorize,
+				authorize: neverAuthorize,
 			},
 			assertions: func(err error) {
 				require.Error(t, err)
@@ -207,7 +207,7 @@ func TestSecretsServiceSet(t *testing.T) {
 		{
 			name: "error getting project from store",
 			service: &secretsService{
-				projectAuthorize: alwaysProjectAuthorize,
+				authorize: alwaysAuthorize,
 				projectsStore: &mockProjectsStore{
 					GetFn: func(context.Context, string) (Project, error) {
 						return Project{}, errors.New("something went wrong")
@@ -221,8 +221,25 @@ func TestSecretsServiceSet(t *testing.T) {
 			},
 		},
 		{
+			name: "user is not a project admin",
+			service: &secretsService{
+				authorize:        alwaysAuthorize,
+				projectAuthorize: neverProjectAuthorize,
+				projectsStore: &mockProjectsStore{
+					GetFn: func(context.Context, string) (Project, error) {
+						return Project{}, nil
+					},
+				},
+			},
+			assertions: func(err error) {
+				require.Error(t, err)
+				require.IsType(t, &meta.ErrAuthorization{}, err)
+			},
+		},
+		{
 			name: "error setting secret in store",
 			service: &secretsService{
+				authorize:        alwaysAuthorize,
 				projectAuthorize: alwaysProjectAuthorize,
 				projectsStore: &mockProjectsStore{
 					GetFn: func(context.Context, string) (Project, error) {
@@ -244,6 +261,7 @@ func TestSecretsServiceSet(t *testing.T) {
 		{
 			name: "success",
 			service: &secretsService{
+				authorize:        alwaysAuthorize,
 				projectAuthorize: alwaysProjectAuthorize,
 				projectsStore: &mockProjectsStore{
 					GetFn: func(context.Context, string) (Project, error) {
--- a/v2/apiserver/internal/api/secrets_test.go
+++ b/v2/apiserver/internal/api/secrets_test.go
@@ -195,7 +195,7 @@ func TestSecretsServiceSet(t *testing.T) {
 		assertions func(error)
 	}{
 		{
-			name: "unauthorized",
+			name: "user does not have read permissions",
 			service: &secretsService{
 				authorize: neverAuthorize,
 			},
--- a/v2/apiserver/internal/api/secrets.go
+++ b/v2/apiserver/internal/api/secrets.go
@@ -165,7 +165,7 @@ func (s *secretsService) Set(
 	projectID string,
 	secret Secret,
 ) error {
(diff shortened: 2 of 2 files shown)


AnuragThePathak:
Should we add a unit tests as well?

krancour:
> Should we add unit tests as well?

They haven't run yet, but I'm pretty sure they'll fail if they're not amended.

AnuragThePathak:
Should I create a new file for the tests of secrets.go file? It doesn't exist right now ig.
Moreover, I was not able to able to run unit tests using make test-unit-go (because of network issue or proxy issue) yesterday. But when I ran tests directly in my system all tests passed. So I believe there are no tests for the secrets.

krancour:
> Should I create a new file for the tests of secrets.go file? It doesn't exist right now ig.

Oooh. That's a gap in our testing. Thank you for calling my attention to it.

I'll add those tests today, including ones for the new stuff you've added in this PR and we'll combine our commits together.

AnuragThePathak:
Should I create the PR from a different branch? I remembered that after committing only so created the PR from main only.

krancour:
@AnuragThePathak nope. You did everything right.

krancour:
@AnuragThePathak I opened #1903 to add the missing tests. After that's merged, we can rebase this PR and you can make the incremental changes necessary to the tests to validate the changes you've already made.

Happy to help if you need help rebasing or anything like that.

krancour:
@AnuragThePathak afaik, Kind and our whole `make hack` process should work fine as long as Docker is working correctly.

The only thing I highly suggest (even for folks with fast internet) is to use a local Docker registry when you're hacking away. `make hack-new-kind-cluster` will actually set one up for you and connect the KinD cluster to it, but you do need to set an env var for `make hack` to use it:

```shell
$ export DOCKER_REGISTRY=localhost:5000
```

If you do that, your internet connection will only ever slow you down the first time base images are pulled and 100% of everything else that happens will happen locally.

AnuragThePathak:
> export DOCKER_REGISTRY=localhost:5000

Yeah I did that but still brigade release installation fails with the following statements
> Error: timed out waiting for the condition
   make: *** [Makefile:521: hack-deploy] Error 1

I'm not really sure everything is fine because of that proxy issue.

krancour:
@AnuragThePathak what state are all the pods in?

```
$ kubectl get pods -n brigade
```

AnuragThePathak:
> @AnuragThePathak what state are all the pods in?
> 
> ```
> $ kubectl get pods -n brigade
> ```

Before running `make hack` you mean?

AnuragThePathak:
I can see that the problem is occurring because of not being able to pull local images.

krancour:
@AnuragThePathak how did you start the KinD cluster?

If you just did `kind create cluster`, it will lack the necessary configuration to connect to a local registry.

We provide a script that sets everything up correctly. You can kick it off with `make hack-new-kind-cluster`.

Again... I really need to cook up some kind of primer on all of this, so thank you for your patience.

AnuragThePathak:
I created the kind cluster using `make hack-new-kind-cluster` as mentioned in the [docs](https://docs.brigade.sh/topics/developers/#kind-configuration).

It was working perfectly when I was at my home and working without proxy, not sure if it's some proxy issue.

AnuragThePathak:
I've figured out the issue it was occurring as brigade-dev-registry was not in no proxy.
Now again I got stuck in the github OAuth login as localhost is no proxy ig, hence not able to connect to github servers.

krancour:
@AnuragThePathak I'm not completely understanding your proxy issues. Specifically, I don't understand why individual containers, for instance, would need any special configuration to work with your proxy. Is it not the case that your whole _system_ is configured to use the proxy for all outbound traffic?

Also, you don't need to use OAuth for hacking on Brigade. It's a non-production cluster, so root is enabled. Just log in as root.

AnuragThePathak:
> Is it not the case that your whole system is configured to use the proxy for all outbound traffic?

I think the proxy is not limited to outbound traffic only as we need to add localhost and a few other services to no proxy list.

> Also, you don't need to use OAuth for hacking on Brigade. It's a non-production cluster, so root is enabled. Just log in as root.

But I need to do so because of the following reason
> This is only true if you're logged in as a real user and not as root.

krancour:
> I think the proxy is not limited to outbound traffic only as we need to add localhost and a few other services to no proxy list.

This sounds strange to me, but I can't get bogged down in troubleshooting your proxy.

> But I need to do so because of the following reason
>
>> This is only true if you're logged in as a real user and not as root.

Ah. Right.

This is where the beauty and utility of the unit tests comes in. Unit tests exist precisely so you can validate the functionality of a small piece of code without being forced to run the entire system.

AnuragThePathak:
> This is where the beauty and utility of the unit tests comes in. Unit tests exist precisely so you can validate the functionality of a small piece of code without being forced to run the entire system.

Cool. Btw I have updated the tests to which was required with the fix of this issue.

v2/apiserver/internal/api/secrets_test.go
```diff
@@ -197,7 +197,7 @@ func TestSecretsServiceSet(t *testing.T) {
 		{
 			name: "unauthorized",
```
krancour, line 198: If you could possibly change the name of this test case to "user does not have read permissions" it would be more clear and then everything else looks great!

krancour:
> I have updated the tests to which was required with the fix of this issue.

Awesome! Just took a look.

I left one small comment, but all of this looks fantastic.

Once again, thank you so much for this and other recent contributions!

v2/apiserver/internal/api/secrets_test.go
AnuragThePathak: Yeah agree.

AnuragThePathak:
> Once again, thank you so much for this and other recent contributions!

You're welcome. I personally like this project not only from developer perspective but also from user perspective.

krancour (approved):
LGTM pending CI

PR #1908, opened 2022-04-05
expand the contributor guide

krancour:
Fixes #1694

It is also my intention that new and very short contributor guides for all peripheral components will include only component-specific instruction and will link back to here to discuss our general development approach.

PR #1910, opened 2022-04-05
use tilt for faster feedback loop during development

krancour:
This PR adds everything we need to use Tilt for improved devx, but it does _not_ remove our hack make targets yet. I am holding off on that until #1908 is ready to go.

i.e. The old and documented way of doing things will still work for now.

### DIFF
--- a/.tiltignore
+++ b/.tiltignore
@@ -1,0 +1,2 @@
+charts/brigade/charts
+charts/brigade/tmpcharts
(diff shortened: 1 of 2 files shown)


vdice (approved):
This is awesome!  Tested with a kind cluster created from https://github.com/brigadecore/brigade/pull/1915.  Works great -- and fast.

LGTM

PR #1915, opened 2022-04-08
devx: use ctlptl as alternative to hacky kind script

krancour:
This PR creates a less hacky alternative `make hack-new-kind-cluster` script.

This PR does not remove the old way of doing things. We can revisit after #1908 is amended appropriately and merged.


vdice (approved):
Works great!  Tested in tandem with the tilt PR.  LTGM!

PR #1916, opened 2022-04-08
Fallible jobs pt2

krancour:
Since #1772 we've allowed jobs to be marked as allowed to fail without impacting the overall success or failure of the worker.

Until now, the worker has kept this information to itself.

This is the first of a few related PRs that will get the worker to start sharing this information with the API server when it creates new jobs.

The API server won't do anything directly with this information, however, having this information available makes it possible to share with gateways that monitor event/job status and report that status upstream. For a perfect example, see https://github.com/brigadecore/brigade-github-gateway/issues/93


vdice (approved):

krancour:
@vdice I made a very important amendment to this PR in fefa0e46f3a8842c021aae3e35174947d2ed51be

Not sure if you want to re-review.

PR #1958, opened 2022-05-17
fix project update failure if project doesn't exist and not logged in as root

### LINKED ISSUE #1947: brig update -c fails if project doesn't exist and not logged in as root
With `-c` set, the project should be created if it doesn't already exist, but instead there is an auth error.

This doesn't occur if you're root.

This may be a similar issue to #1893

cc @AnuragThePathak since you have experience fixing this exact sort of issue.

AnuragThePathak:
Fixes #1947 

**What this PR does / why we need it**: It solves the issue of brig update receiving authorization error when project doesn't exist and not logged in as root.

**Special notes for your reviewer**:  I've removed checking of project not found error in the project update function after the update attempt using Project Store because I feel it is no longer relevant now as we already did that earlier part of that function.
I have also modified the unit test just that all test cases pass. However they may be rearranged now to match the change in the update function (I can do that as well if required).
Moreover when -c flag is used we create a project which is requested to be updated but in logs we inform the project to be updated instead of informing it to be created. Not really sure if that's how it's supposed to work.

**If applicable**:

### DIFF
--- a/v2/apiserver/internal/api/projects.go
+++ b/v2/apiserver/internal/api/projects.go
@@ -345,27 +345,34 @@ func (p *projectsService) Update(
 	project Project,
 	opts ProjectUpdateOptions,
 ) error {
+	if err := p.authorize(ctx, RoleReader, ""); err != nil {
+		return err
+	}
+
+	if _, err := p.projectsStore.Get(ctx, project.ID); err != nil {
+		_, isErrNotFound := errors.Cause(err).(*meta.ErrNotFound)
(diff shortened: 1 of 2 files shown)


krancour:
@AnuragThePathak this looks great!

I am noticing one issue -- but it's not new -- _not_ introduced by your change. That is that we aren't verifying the user has permission to create projects before proceeding with project creation.

Especially given that this PR does not introduce that problem, let's treat it as a separate problem. I'll open a new issue and if you don't mind, I'll assign it to you.

/brig run

krancour (approved):

